PR #1, opened 2018-08-26
Updated vertical rails

kruthers:
Updated vertical rails to 1.13 and added an advancement to it


gm4_vertical_rails/data/vertical_rails/functions/momentum.mcfunction
```diff
@@ -0,0 +1,5 @@
+execute if block ~ ~ ~ ladder run tp ~ ~.3 ~
```
SpecialBuilder32: At this point, the function should only be running if the minecart is in a ladder. The teleport should not double-check the block.

gm4_vertical_rails/data/vertical_rails/functions/main.mcfunction
```diff
@@ -0,0 +1,8 @@
+execute as @e[type=minecart] at @s if block ~ ~ ~ ladder run tag @s add gm4_minecart
```
Bluefire-610: `gm4_minecart` is a really generic name, should be `gm4_vertical_minecart`

gm4_vertical_rails/data/vertical_rails/functions/main.mcfunction
```diff
@@ -0,0 +1,8 @@
...
+execute as @e[type=chest_minecart] at @s if block ~ ~ ~ ladder run tag @s add gm4_minecart
```
Bluefire-610: `gm4_minecart` is a really generic name, should be `gm4_vertical_minecart`

gm4_vertical_rails/data/vertical_rails/functions/main.mcfunction
```diff
@@ -0,0 +1,8 @@
...
+execute as @e[type=hopper_minecart] at @s if block ~ ~ ~ ladder run tag @s add gm4_minecart
```
Bluefire-610: `gm4_minecart` is a really generic name, should be `gm4_vertical_minecart`

gm4_vertical_rails/data/vertical_rails/functions/main.mcfunction
```diff
@@ -0,0 +1,8 @@
...
+execute as @e[type=furnace_minecart] at @s if block ~ ~ ~ ladder run tag @s add gm4_minecart
```
Bluefire-610: `gm4_minecart` is a really generic name, should be `gm4_vertical_minecart`

gm4_vertical_rails/data/vertical_rails/functions/main.mcfunction
```diff
@@ -0,0 +1,8 @@
...
+execute as @e[type=spawner_minecart] at @s if block ~ ~ ~ ladder run tag @s add gm4_minecart
```
Bluefire-610: `gm4_minecart` is a really generic name, should be `gm4_vertical_minecart`

gm4_vertical_rails/data/vertical_rails/functions/main.mcfunction
```diff
@@ -0,0 +1,8 @@
...
+execute as @e[type=tnt_minecart] at @s if block ~ ~ ~ ladder run tag @s add gm4_minecart
```
Bluefire-610: `gm4_minecart` is a really generic name, should be `gm4_vertical_minecart`

gm4_vertical_rails/data/vertical_rails/functions/pulse_check.mcfunction
```diff
@@ -0,0 +1,3 @@
+#compares assigned run tick to current tick and calls main
+execute if score gm4_currentTick gm4_clockTick = vertical_rails gm4_clockTick run function vertical_rails:main
+execute as @e[tag=gm4_minecart] at @s if block ~ ~ ~ ladder run function vertical_rails:momentum
```
Bluefire-610: Change to `execute as @e[tag=gm4_vertical_minecart]`

gm4_vertical_rails/data/vertical_rails/functions/momentum.mcfunction
```diff
@@ -0,0 +1,5 @@
+tp ~ ~.3 ~
```
Bluefire-610: Just wondering, is the tp necessary?

gm4_vertical_rails/data/vertical_rails/functions/main.mcfunction
```diff
@@ -0,0 +1,10 @@
+execute as @e[type=minecart] at @s if block ~ ~ ~ ladder run tag @s add GM4_cart
+execute as @e[type=chest_minecart] at @s if block ~ ~ ~ ladder run tag @s add GM4_cart
+execute as @e[type=hopper_minecart] at @s if block ~ ~ ~ ladder run tag @s add GM4_cart
+execute as @e[type=furnace_minecart] at @s if block ~ ~ ~ ladder run tag @s add GM4_cart
+execute as @e[type=spawner_minecart] at @s if block ~ ~ ~ ladder run tag @s add GM4_cart
+execute as @e[type=tnt_minecart] at @s if block ~ ~ ~ ladder run tag @s add GM4_cart
```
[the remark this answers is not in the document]
Bluefire-610: We want to change that for ziprails. Using the same tag will cause unnecessary checks to run if both modules are installed.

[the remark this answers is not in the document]
Bluefire-610: Actually, Zip Rails tags the minecarts regardless of whether they are zipped, it just tags them. This module is supposed to tag them once near a ladder. Therefore, Zip Rails should continue to use the gm4_minecart tag, whereas this module should use it's own tag.

gm4_vertical_rails/data/vertical_rails/functions/pulse_check.mcfunction
SpecialBuilder32: Why cant both ziprails and vertical rails use `gm4_minecart` ? It's just used as a generic marker for all minecarts types amd isn't ever removed.

Bluefire-610: If both do, then ziprails will look for ziprail hooks whilst you're riding a vertical rail - which can't happen. Even worse, vertical rails will always try to apply a force to minecarts or check for ladders around them. It will just cause a huge amount of not needed checks. (see my comments on your original re-naming edit)

SpecialBuilder32: But `gm4_minecart` isn't use in ziprails as a "entity currently riding a zipline" marker, its just a tag that takes the place of `type=minecart` in selectors, which can't be optimized any further. 

Regarding your comment on the original edit: The tag can't reflect carts that are near ladders since the only way to know if a cart is near a ladder is to execute off of it, and the only (good) way to execute off all carts to tell if they're near a ladder is to give them all a common tag. 

If the tag *were* to reflect the minecart being near a ladder, then you would need to have an `if block ~ ~ ~ ladder` in every `tag add gm4_vertical_minecart` statement, which is many more checks than just checking the block before you run the main cart functions. 

The only way for ziprails and vertical rails to work is for them to constantly check the blocks around them, as there's no other way for them to get information about if they need to do something.

Bluefire-610: So do we leave that first check in?

PR #2, opened 2018-08-28
Add better armor stands and sweethearts modules

misode:


SpiderRobotMan:
I'd try to keep each modules a separate PR

misode:
Ah, my bad. I'm still a bit new with pull requests

PR #6, opened 2018-08-29
Add orbis custom terrain module

misode:
I updated the custom terrain module just by looking at what it does, so there may be slight differences in the way things are done. Instead of separating this into different modules, it's now one module. Due to the way it's made, other modules can run functions when a chunk is generated

**Main Features**
* Cooler Caves (affects caves and mineshafts in deserts, snow biomes, oceans and mesa's)
* Dangerous Dungeons (customized for each of the above biomes)
* Mountain Towers (customized for each of the above biomes)
* Dungeons and towers generated with random spawners and loot

**Biome Scanning**
A bunch of block checking makes sure the biomes are detected optimally. When comparing adjacent biomes it fixes the biome type in the case it was incorrectly detected. For example this stops an ocean chunk from generating just because there was a large river.

**Images**


misode:
For the chest population, I assume we want to use custom loot tables? Do they need to be different based on height, biome, structure type?

misode:
Following commit is essential to this module as it ensures structures are seed specific. Previously it was using /spreadplayers to find the surface. Now it uses recursion which also has the benefits of ignoring trees and snow layers.

Another improvement is that dungeons and towers will both be as common. Previously it spawned a tower if the terrain was high and a dungeon if it was low. Dungeons now check for granite and towers check for diorite at y=1. There's still the "issue" that structures always get spawned in the same x and z coordinate inside a chunk, but it's not a major issue and I think it was already the case in the 1.8 version.

misode:
By making structure spawning extendable, it's no longer seed deterministic when a second module adds a structure variation. It is guaranteed to spawn a structure, but the variation may be different between different generations of the same seed. Different configurations of extensions will of cause differences between generations.

misode:
It might be a good idea to squash these commits when you eventually merge

sparks0:
@misode is this up-to-date with the version used to run the GM4D server? If so I'd say that's a good enough test of the system.

SpiderRobotMan:
Make sure you squash this when you merge, so many commits

sparks0:
@misode this now needs adjusting for the new gm4 clock :)

misode:
Updated to the new standard

sparks0:
Module doesn't automatically start on install. Players are missing the tag that tells Orbis to scan around them.

PR #7, opened 2018-08-29
Add heart canisters module

misode:
Crafting hasn't been implemented due to custom crafting not being done yet. If you want to give yourself a the heart canister items, you can do so by running the following command:
```
function heart_canisters:give
```


gm4_heart_canisters/data/gm4/advancements/heart_canisters_max.json
```diff
@@ -0,0 +1,34 @@
...
+  },
+  "parent":"gm4:heart_canisters_tier_2",
+  "criteria":{
+    "have_all_heart_cans":{
+      "trigger":"minecraft:inventory_changed",
+      "conditions": {
+        "items": [
+          {
+            "item": "minecraft:player_head",
+            "count": 5,
```
SpecialBuilder32: Does this trigger if the player has 5 total in their inventory, or 5 in a given stack. If it is the former, it might be slightly confusing for a player who gets the achievement but not the health boost.

gm4_heart_canisters/data/gm4/advancements/heart_canisters_tier_1.json
```diff
@@ -0,0 +1,27 @@
...
+    "title":"Can't Buy Me Love",
+    "description":{
+      "text":"Craft a tier 1 Heart Canister.",
+      "color":"gray"
+    }
+  },
+  "parent":"gm4:root",
+  "criteria":{
+    "craft_tier_1":{
+      "trigger":"minecraft:inventory_changed",
```
SpecialBuilder32: This will also trigger on the player getting a heart can from another player, not just crafting it themselves. Since most vanilla crafting achievements also can be earned in this manner, it should be alright to leave it this way, even though the title is "Craft a heart can"

gm4_heart_canisters/data/heart_canisters/functions/init.mcfunction
```diff
@@ -0,0 +1,19 @@
+#announce module installation
+tellraw @a[gamemode=creative] ["",{"text":"[GM4]: Installing Heart Canisters..."}]
+execute unless entity @p run say GM4: Installing Heart Canisters...
+
+#declare and initialise scoreboards and settings
+scoreboard players set updateHappened gm4_upCheck 1
+scoreboard players set heart_canisters gm4_modules 1
+scoreboard players set heart_canisters gm4_clockTick 0
+
+scoreboard objectives add heart_can_health dummy
```
SpecialBuilder32: Scoreboard objectives should be prefixed with "gm4_"

gm4_heart_canisters/data/heart_canisters/functions/check.mcfunction
```diff
@@ -0,0 +1,11 @@
+scoreboard players set @s[nbt={Inventory:[{Count:1b,tag:{gm4_heart_canister_tier:1b}}]}] heart_can_health 1
+scoreboard players set @s[nbt={Inventory:[{Count:2b,tag:{gm4_heart_canister_tier:1b}}]}] heart_can_health 2
+scoreboard players set @s[nbt={Inventory:[{Count:3b,tag:{gm4_heart_canister_tier:1b}}]}] heart_can_health 3
+scoreboard players set @s[nbt={Inventory:[{Count:4b,tag:{gm4_heart_canister_tier:1b}}]}] heart_can_health 4
+scoreboard players set @s[nbt={Inventory:[{Count:5b,tag:{gm4_heart_canister_tier:1b}}]}] heart_can_health 5
+
+scoreboard players add @s[nbt={Inventory:[{Count:1b,tag:{gm4_heart_canister_tier:2b}}]}] heart_can_health 1
```
SpecialBuilder32: By using add here, a player would be able to use a tier 2 heart can as a regular heart can. The second tier heart cans should only be active if the player also has 5 tier 1's in their inventory.

gm4_heart_canisters/data/heart_canisters/functions/check.mcfunction
```diff
@@ -0,0 +1,11 @@
+scoreboard players set @s[nbt={Inventory:[{Count:1b,tag:{gm4_heart_canister_tier:1b}}]}] heart_can_health 1
```
SpecialBuilder32: This function is to vaguely named. "heart_canister_count" or "check_heart_cans" would be more descriptive

gm4_heart_canisters/data/heart_canisters/functions/update.mcfunction
```diff
@@ -0,0 +1,14 @@
+scoreboard players operation @s heart_can_time -= @s heart_can_health
```
SpecialBuilder32: This function is to vaguely named. "apply_health_boost" or similar would be more descriptive.

gm4_heart_canisters/data/heart_canisters/functions/update.mcfunction
```diff
@@ -0,0 +1,14 @@
...
+effect give @s[scores={heart_can_time=..1336,heart_can_health=2}] health_boost 1000000 1 true
+effect give @s[scores={heart_can_time=..1336,heart_can_health=3}] health_boost 1000000 2 true
+effect give @s[scores={heart_can_time=..1336,heart_can_health=4}] health_boost 1000000 3 true
+effect give @s[scores={heart_can_time=..1336,heart_can_health=5}] health_boost 1000000 4 true
+effect give @s[scores={heart_can_time=..1336,heart_can_health=6}] health_boost 1000000 5 true
+effect give @s[scores={heart_can_time=..1336,heart_can_health=7}] health_boost 1000000 6 true
+effect give @s[scores={heart_can_time=..1336,heart_can_health=8}] health_boost 1000000 7 true
+effect give @s[scores={heart_can_time=..1336,heart_can_health=9}] health_boost 1000000 8 true
+effect give @s[scores={heart_can_time=..1336,heart_can_health=10}] health_boost 1000000 9 true
+scoreboard players operation @s heart_can_time = @s heart_can_health
```
SpecialBuilder32: So I'm partially confused with what you're doing here. I've figured out that it only tries to apply the health boost if your number of heart canisters has changed, but why try and figure that out in the first place? It looks to me that since its not constantly applying the effect, when the duration expires, the effect won't be reapplied to the player until they move their heart-canisters again. And why 1336? It dosen't seem to represent anything. The highest heart_can_time should ever be is 10.

SpecialBuilder32: Also, "heart_can_time" to represent the last "heart_can_health" value is confusing. I was expecting it to somehow be related to durations or timings.

misode: I honestly don't know it either but that's how it was in the 1.12 version

gm4_heart_canisters/data/heart_canisters/functions/apply_health_boost.mcfunction
```diff
@@ -0,0 +1,11 @@
+effect clear @s health_boost
```
Bluefire-610: Do we need to clear the heath boost here? Won't it just be overridden if the health boost already on the player isn't stronger than the one to be applied?

gm4_heart_canisters/data/heart_canisters/functions/main.mcfunction
```diff
@@ -0,0 +1,9 @@
+# check how many heart canisters the player has
+execute as @a[gamemode=!spectator,gamemode=!creative,nbt={Inventory:[{tag:{gm4_heart_canister:1b}}]}] run function heart_canisters:check_heart_canister_count
+
+# give the player the health boost effect
+execute as @a[gamemode=!spectator,gamemode=!creative,scores={gm4_heart_can=1..}] run function heart_canisters:apply_health_boost
```
Bluefire-610: I think it'd be smart to move this function call into the `#heart_canisters:check_heart_canister_count` function. That way it only runs if there are players with canisters around without the need for another check if anyone's using a heart can.

gm4_heart_canisters/data/heart_canisters/functions/main.mcfunction
```diff
@@ -0,0 +1,9 @@
...
+
+# fix broken items
+data merge entity @e[limit=1,type=item,nbt={Item:{id:"minecraft:player_head",Count:1b,tag:{SkullOwner:{Id:"25b1ff93-7926-1cd7-53fe-8e6050792daa"}}}}] {Item:{tag:{gm4_heart_canister:1b,gm4_heart_canister_tier:1b,display:{Lore:["Tier 1"],Name:"{\"text\":\"Heart Canister\",\"italic\":false}"}}}}
+data merge entity @e[limit=1,type=item,nbt={Item:{id:"minecraft:player_head",Count:1b,tag:{SkullOwner:{Id:"25b1ff93-7925-1cd7-53fe-8e6050792daa"}}}}] {gm4_heart_canister:1b,gm4_heart_canister_tier:2b,Item:{tag:{display:{Lore:["Tier 2"],Name:"{\"text\":\"Heart Canister\",\"italic\":false}"}}}}
```
Bluefire-610: These only fix one broken item per clock pulse, is this something we want to "fix", @SpecialBuilder32 ?

gm4_heart_canisters/data/heart_canisters/functions/give.mcfunction
```diff
@@ -0,0 +1,2 @@
+give @s player_head{gm4_heart_canister:1b,gm4_heart_canister_tier:1b,display:{Lore:["Tier 1"],Name:"{\"text\":\"Heart Canister\",\"italic\":false}"},SkullOwner:{Id:"25b1ff93-7926-1cd7-53fe-8e6050792daa",Properties:{textures:[{Value:"eyJ0aW1lc3RhbXAiOjE0MjkzNDc5MDU4NzMsInByb2ZpbGVJZCI6IjYzY2JkZjhkNDg4OTQ3NWY5NDQxMjk3ZTRhM2Q1NjczIiwicHJvZmlsZU5hbWUiOiJWZWxlVCIsImlzUHVibGljIjp0cnVlLCJ0ZXh0dXJlcyI6eyJTS0lOIjp7InVybCI6Imh0dHA6Ly90ZXh0dXJlcy5taW5lY3JhZnQubmV0L3RleHR1cmUvNTNjZjA2ZDViNDdiNzQ2ZDE2ZWU2MzExODdjYjg1N2Q4YzE2YmZjZDU3MGE0MGYyODZiMmMzODg2N2NmZGEifX19"}]}}}
+give @s player_head{gm4_heart_canister:1b,gm4_heart_canister_tier:2b,display:{Lore:["Tier 2"],Name:"{\"text\":\"Heart Canister\",\"italic\":false}"},SkullOwner:{Id:"25b1ff93-7925-1cd7-53fe-8e6050792daa",Properties:{textures:[{Value:"eyJ0aW1lc3RhbXAiOjE0MjkzNDc5OTc3ODcsInByb2ZpbGVJZCI6IjZjZjU0M2E2MGVlOTQzN2NiNjE0YzdiOTRkZTVjNWI3IiwicHJvZmlsZU5hbWUiOiJNcnNNYWtpc3RlaW4iLCJpc1B1YmxpYyI6dHJ1ZSwidGV4dHVyZXMiOnsiU0tJTiI6eyJ1cmwiOiJodHRwOi8vdGV4dHVyZXMubWluZWNyYWZ0Lm5ldC90ZXh0dXJlLzQ1NWQ5YmY4NWFjNTY1YjM1ZTFmYTE5MjQ3Y2E2NTQxZGMyZTMzNGJlYzExNWNiNDQ5ZWZiZThlOWI4MTAyMiJ9fX0="}]}}}
```
Bluefire-610: Function isn't required anymore since custom crafters was added. Don't know how to add files to this, though.

gm4_heart_canisters/data/heart_canisters/functions/main.mcfunction
misode: Oh, yeah, It should be replaced with an `execute as`.

gm4_heart_canisters/data/heart_canisters/functions/apply_health_boost.mcfunction
misode: It needs to be cleared so it can correctly lower the health boost when the player removes a heart canister

gm4_heart_canisters/data/heart_canisters/functions/main.mcfunction
misode: The reason these two are separate is to make it easier in the future to make for example Zauber Armor compatible. All that would have to be done is add a function tag call in between the `check_heart_canister_count` and `apply_health_boost` calls. Zauber Armor would then modify the `gm4_heart_can` scoreboard.

SpecialBuilder32: Note that zauber armor uses attributes to increase the health of the player, not the health boost effect. Heart cans *would* use that too if they didnt have to be worn or held for attributes to take effect.

misode: Ohh okay my bad

PR #9, opened 2018-09-01
Add enchantment extractors module

misode:


gm4_enchantment_extractors/data/enchantment_extractors/functions/destroy.mcfunction
```diff
@@ -0,0 +1,3 @@
+summon item ~ ~1 ~ {Item:{id:enchanting_table,Count:1},Passengers:[{id:item,Item:{id:"hopper",Count:1}},{id:item,Item:{id:"book",Count:7}},{id:item,Item:{id:"writable_book",Count:1}}]}
+data remove entity @e[type=item,distance=..3,limit=1,nbt={Item:{tag:{display:{Name:"{\"text\":\"Enchantment Extractor\"}"}}}}] Item.tag.display
```
Bluefire-610: I'm kinda outta the loop here, what's this line for?

gm4_enchantment_extractors/data/enchantment_extractors/functions/init.mcfunction
```diff
@@ -0,0 +1,18 @@
+#announce module installation
+tellraw @a[gamemode=creative] ["",{"text":"[GM4]: Installing Enchantment Extractors..."}]
+execute unless entity @p run say GM4: Installing Enchantment Extractors...
+
+#declare and initialise scoreboards and settings
+scoreboard players set updateHappened gm4_upCheck 1
+scoreboard players set enchantment_extractors gm4_modules 1
+scoreboard players set enchantment_extractors gm4_clockTick 0
+
+scoreboard objectives add gm4_random dummy
```
Bluefire-610: Like with desire lines, we can't use a generic scoreboard name here. How does gm4_ench_random sound?

gm4_enchantment_extractors/data/enchantment_extractors/functions/create.mcfunction
```diff
@@ -0,0 +1,5 @@
+summon armor_stand ~ ~-1.2 ~ {Invisible:1b,Invulnerable:1b,Marker:1b,NoGravity:1,Tags:["gm4_enchantment_extractor","gm4_no_edit"],ArmorItems:[{},{},{},{id:enchanting_table,Count:1}],DisabledSlots:2039583}
+data merge block ~ ~ ~ {CustomName:"\"Enchantment Extractor\"",Items:[]}
+advancement grant @a[distance=..3] only gm4:enchantment_extractors
```
Bluefire-610: `@a[distance=..3]` should be `@a[distance=..3,gamemode=!spectator]` to exclude spectators. This prevents them getting advancements whilst spectating people which has been an issue with previous versions of multiple modules.

gm4_enchantment_extractors/data/enchantment_extractors/functions/enchantment.mcfunction
```diff
@@ -0,0 +1,94 @@
...
+execute if score extract_chance gm4_random matches ..35 if entity @s[nbt={Item:{tag:{Enchantments:[{lvl:3s,id:"minecraft:lure"}]}}}] run replaceitem block ~ ~ ~ container.1 enchanted_book{StoredEnchantments:[{id:"minecraft:lure",lvl:3}]}
+execute if score extract_chance gm4_random matches ..8 if entity @s[nbt={Item:{tag:{Enchantments:[{lvl:1s,id:"minecraft:curse_of_vanishing"}]}}}] run replaceitem block ~ ~ ~ container.2 enchanted_book{StoredEnchantments:[{id:"minecraft:curse_of_vanishing",lvl:1}]}
+execute if score extract_chance gm4_random matches ..5 if entity @s[nbt={Item:{tag:{Enchantments:[{lvl:1s,id:"minecraft:mending"}]}}}] run replaceitem block ~ ~ ~ container.3 enchanted_book{StoredEnchantments:[{id:"minecraft:mending",lvl:1}]}
+execute if score extract_chance gm4_random matches ..5 if entity @s[nbt={Item:{tag:{Enchantments:[{lvl:1s,id:"minecraft:mending"}]}}}] run advancement grant @a[distance=..4] only gm4:enchantment_extractors_mending
```
Bluefire-610: For consistency `grant @a[distance=..4]` should be `distance=..3` and a `gamemode=!spectator` should be added to avoid giving achievements to them.

gm4_enchantment_extractors/data/enchantment_extractors/functions/destroy.mcfunction
misode: It removes the display tag so it will stack with other hoppers

gm4_enchantment_extractors/data/enchantment_extractors/functions/create.mcfunction
```diff
@@ -0,0 +1,5 @@
+summon armor_stand ~ ~-1.2 ~ {Invisible:1b,Invulnerable:1b,Marker:1b,NoGravity:1,Tags:["gm4_enchantment_extractor","gm4_no_edit"],ArmorItems:[{},{},{},{id:enchanting_table,Count:1}],DisabledSlots:2039583}
```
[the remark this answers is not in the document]
SpecialBuilder32: The name should be the same as the "scoreboard" tag. `gm4_enchantment_extractor`

gm4_enchantment_extractors/data/enchantment_extractors/functions/extract_failed.mcfunction
```diff
@@ -1,6 +1,9 @@
-execute if score extract_chance gm4_ench_random matches ..80 run particle explosion ~ ~1.7 ~ 0.2 0.2 0.2 0.05 25 force @a[distance=..10]
-execute if score extract_chance gm4_ench_random matches ..80 run playsound item.shield.break block @a[distance=..8] ~ ~ ~ 0.6 0.3
+scoreboard players add fail_type gm4_ench_random 27
```
SpecialBuilder32: The issue with using the 27-100 randomization method here is that this whole function only runs when something fails, so failures would always follow the same pattern. 1.8 used the rotation of the item because it was inherently random and isn't controllable by the player except by trial and error. The randomization for the enchantment success can depend on the 27-100 method since 27 gets added constantly, and the score's represents a more fully random number.

gm4_enchantment_extractors/data/enchantment_extractors/functions/extract_failed.mcfunction
```diff
@@ -1,6 +1,9 @@
...
-execute if score extract_chance gm4_ench_random matches 91.. run summon vex ~ ~-1 ~ {CustomName:"\"Living Enchantment\"",CustomNameVisible:0,HandItems:[{id:golden_sword,Count:1},{}],HandDropChances:[0.0f,0.0f],Motion:[0.0,0.25,0.0]}
+execute if score fail_type gm4_ench_random matches ..80 run particle explosion ~ ~1.7 ~ 0.2 0.2 0.2 0.05 25 force @a[distance=..10]
+execute if score fail_type gm4_ench_random matches ..80 run playsound item.shield.break block @a[distance=..8] ~ ~ ~ 0.6 0.3
+
+execute if score fail_type gm4_ench_random matches 81.. run playsound entity.evocation_illager.prepare_attack block @a[distance=..8] ~ ~ ~ .8 1.8
+execute if score fail_type gm4_ench_random matches 81..90 run summon vex ~ ~-1 ~ {CustomName:"\"Congealed Enchantment\"",CustomNameVisible:0,LifeTicks:100,Attributes:[{Name:generic.attackDamage,Base:2}],Health:4.0,Motion:[0.0,0.35,0.0]}
+execute if score fail_type gm4_ench_random matches 91.. run summon vex ~ ~-1 ~ {CustomName:"\"Living Enchantment\"",CustomNameVisible:0,HandItems:[{id:golden_sword,Count:1},{}],HandDropChances:[0.0f,0.0f],Motion:[0.0,0.25,0.0]}
```
SpecialBuilder32: You're sill missing the second sound effect that occurs for the vex. There are two slightly different sound effects,

gm4_enchantment_extractors/data/enchantment_extractors/functions/extract_failed.mcfunction
misode: I'm using a different fake player here. There would be a 80% chance of no effect, and 10% each for both vexes.

gm4_enchantment_extractors/data/enchantment_extractors/functions/extract_failed.mcfunction
misode: Rip...

misode: Ok found it

gm4_enchantment_extractors/data/enchantment_extractors/functions/extract_failed.mcfunction
misode: @SpecialBuilder32 Is this still an issue? It seems that this review was based before I made the "Separate fail type and extract chance randomization" [commit](https://github.com/Gamemode4Dev/GM4_Datapacks/pull/9/commits/ec71742b4f3b988cbab63e94d94374cee645b594)

SpecialBuilder32: I need to take a look at how it behaves in-game. Just from looking at the functions it looks like it might be but I am unsure. My current concern is that the randomizer won't run enough for the value to be sufficiently random.

gm4_enchantment_extractors/data/enchantment_extractors/functions/extract.mcfunction
```diff
@@ -0,0 +1,10 @@
+#@s = item to have enchantments extracted
+#run from enchantment_extractors:process
+
+scoreboard players add extract_chance gm4_ench_random 27
```
SpecialBuilder32: So, because these are in the extract function, which only gets called once per item, it means that the 27 psudo-random doesn't only gets incremented once per item extracted. This leads to a very not random extraction result. If you get multiple identical items with multiple enchantments, this becomes very clear. Extract the first one, and you'll have decent luck. The second, less so, the third, even less ect... until the score wraps around back to 0 and it starts all over again. This is predictable enough that the players could figure out roughly where in the cycle the score is.

misode: This also the case for the general extract chance. I'm not sure why this simple method was used in the previous version. @SpecialBuilder32 Maybe a more complicated LCG should be used?

SpecialBuilder32: The previous version ran the 27-randomiser constantly when there was an item on an enchantment extractor, meaning that when the value it stopped on depended on how long the item was left on the dropper. Since the player has no way of knowing when the module pulses, how many increments of the randomizer had passed was unknowable.

PR #16, opened 2018-09-02
Add particles pack module

misode:
Requires better armor stands, compatible with poses pack


gm4_particles_pack/data/particles_pack/functions/apply_book.mcfunction
```diff
@@ -1,3 +1,4 @@
+# @s = written book ..1 from armor_stand
```
sparks0: Please state where the function is called from (main) on line 2 as comments.

PR #19, opened 2018-09-28
Better armor stands getting handy advancement

sparks0:
The advancement for better armorstands was the one for poses pack instead of the base. Changed from "Pleanty o' Posing" to "Gettin' Handsy". Have tested in game but would appreciate a second check of the changed files.


misode:
Isn't this more a `fix` rather than a `module` (talking about labels)

sparks0:
> Isn't this more a `fix` rather than a `module`? (talking about labels)

How's bug fix?

misode:
The branch can be deleted now btw :)

PR #27, opened 2018-10-03
Despawn Indicator: A small module for showing when items despawn.

magierpichu:
Hi, I had a little idea for a module, so I thought I'd get myself comfy with the new system.
Anyway, this module will create a particle effect depending on how much time there's left before the item despawns. Yellow for 30, red for ten and black for three seconds.

PS. I'm sorry if this is the wrong way to do it. I'm still getting used to github.


SpiderRobotMan:
I think this is a neat idea, but I'm not sure it's enough of a feature to be a whole module.

magierpichu:
Maybe not, but other than items, I don't know any entities that would benefit from having a clear despawn indication. I just don't understand why something like this isn't already in the game, as it is really frustrating to see your items disappear in front of you. Anyway, if you ever find a spot for it in a broader datapack, I would feel honored.

Bluefire-610:
I agree with spider, it seems a bit thin for it's own module and "I have this concept, let's look for things we can apply it to to make it module-sized" never turned out great. 
On the other hand, the idea is neat. maybe we could try to get a flashing behavior like other games have. This is something to keep in mind for sure.

sparks0:
We have a module that just makes bats  explode. I don't think simple is necessarily bad and I think the problem being addressed here is a real one.

magierpichu:
I decided to also make it work with arrows and I changed the particle effects to show yellow particles when there are 15 seconds left and red when there are only 5 seconds left.

alluet:
Clocking standard is still out of date, and I have an interesting idea to make the animation look a bit cooler. Might rewrite this from scratch.

PR #28, opened 2018-10-04
Better jukeboxes - make hoppers interact with jukeboxes

misode:
This is an idea for a new module that makes it possible for hoppers (which aren't locked) to interact with jukeboxes. Needs an unpowered hopper above with a music disc to insert and needs an unpowered hopper below that has space to extract.

Demonstration:
https://gfycat.com/FrankPortlyKentrosaurus

image

Current issues:
* Playing multiple of the same music discs (even when inserted manually) doesn't work. This is not fixable as far as I know but it's not a huge issue.


misode:
> It should behave like a normal jukebox, only one record can play from it at a time.

It already has that behavior. The only difference with normal jukeboxes is that it uses `/stopsound` to clear the music which causes music of the same record in a different jukebox to be stopped.

sparks0:
@misode I'm happy to accept this module once it has been updated to the correct clocking system :)

sparks0:
Let's leave this until 1.14. This is because my programmable discs module would get confused with this because putting a programmable disc in the hopper will cause the data to be lost when it is put in a jukebox using a hopper. In 1.14 I think we can just copy ALL of the disc NBT into the jukebox which will solve that issue.

PR #32, opened 2018-10-09
Disassemblers

federick0090:
Craft a disassembler and throw iron, gold, diamond tools/ armor to get the materials used back
Armor has chances to not return all materials, takes account durability to lower/ raise chances to get items


gm4_disassemblers/data/disassemblers/functions/items/iron_sword.mcfunction
```diff
@@ -0,0 +1,10 @@
+say iron sword
```
sparks0: Left over say testing command needs to be removed

PR #40, opened 2018-10-18
Add advancement

misode:
Add the Plenty O' Posing advancement


misode:
Hmm this is strange... It doesn't include the original commits (which were reverted)

PR #45, opened 2018-10-19
Adding Destructive Arrows

Scommander:
https://streamable.com/fkdp2
^^ This video sums up the module. Arrow can destroy or change blocks when they land.


XiBlueFireiX:
I think this may be a re-upload. I accidentally uploaded **Ore Trees** and **Destructive Arrows** in the same pull request.

Feedback would be nice too :D

sparks0:
Great job on only checking arrows once, that was the first thing I looked at. This definitely wouldn't be one for servers but could be interesting for singleplayer perhaps. I don't think I like the cobble>gravel conversion as they're not usually interchangeable. I have mixed feeling about how this would fit GM4 but would be interested to hear what the other admins think.

SpiderRobotMan:
I think it's an interesting concept, it does seem that it could be quite destructive even if players don't intend to do it. I don't think having it as a module is a bad thing, users can choose to have it if that's what they want, but I don't see it being in use on our survival server at all because of the unintentional greifing effect it would have.

ToffeeMax:
There seems to be alot of cobbled ideas and I remember reading a comment that some seem small, why not release a miscellaneous modules?

Bluefire-610:
I really like this as a concept, especially how it adds cracks to the stonebricks. This might be better for a custom map, though.

sparks0:
I'm gonna say no on this. I like the concept and it's well written but I don't think it's right for GM4.

PR #51, opened 2018-10-22
Adding TNT Landmines Module

Scommander:
https://streamable.com/o18ly
^^^ This video sums it up 

These TNT Landmines work slightly different than the older versions.
To craft a TNT Landmine you need TNT and a stone pressure plate
Throw the TNT Landmine on the ground to create it.
The TNT Landmine will explode if the block in its position is air or an entity who **isn't** sneaking walks over it.


SpiderRobotMan:
This PR has commits that don't belong, please fix that.

Nathan22211:
really. no way to pick them up or destroy them like in previous versions.in the 1.8 version you were able to destroy the armor stand to break them, and in cookieWookie's version for 1.12 you needed shears to break them.

sparks0:
I prefer the behaviours of the TNT Landmines submitted by Kruthers here https://github.com/Gamemode4Dev/GM4_Datapacks/pull/60 so I'm going to close this one. Sorry!

PR #64, opened 2018-10-24
Speed paths advancement

AndanteDevs:


AndanteDevs:
It's mainly because I thought of a good advancement title haha

Issue #59, opened 2018-10-23
Add more block types to Speed Paths


Compsogbrickus:
A helpful and realistic addition to Speed Paths might be to add more block types that give speed to players, such as gravel, concrete dust, cobblestone, mossy cobblestone, sandstone, wooden planks, polished stones, stone bricks, bricks, and concrete. This could also apply to all slabs/stairs for each block, if applicable. Coarse dirt could work too, as it would be a good advantage to Desire Lines. This improvement would require duplication of the current main.mcfunction code.

AndanteDevs:
Will this be added? If so, I could quickly do that.

misode:
Personally I feel it'd take away the unique ability of path blocks to give you speed

AndanteDevs:
I agree with that

Compsogbrickus:
I  am suggesting this as a builder because most paths are not grass. It seems strange to have a main path block that does not really fit most build styles.

Bluefire-610:
> Personally I feel it'd take away the unique ability of path blocks to give you speed

I agree with @misode here. Paths are something special.

sparks0:
If I added it to any other block it would be coarse dirt because of desire lines but I don't think other blocks should have the same feature. Path blocks are specifically for paths but many of your suggestions might just be floors in houses.

Compsogbrickus:
I suggested common materials that are used in real life for floors because they have good building qualities and are also used for outdoor walkways. Most of Ancient Citadel on PSIV was serviced by wood plank, stone, gravel, cobble, and stone brick paths, and I thought it was weird that those materials didn't have the same buff that path blocks did given the reasoning for speed on path blocks.

Bluefire-610:
I'm closing this issue as it is unlikely to be implemented.

PR #65, opened 2018-10-24
More block types with Speed Paths

AndanteDevs:
Added blocks that resemble paths and of course the grass path itself.
I'm creating a pull request now so that, whether or not this goes ahead, it's ready to go.

- _minecraft:grass_path_
- _minecraft:coarse_dirt_
- _minecraft:gravel_

Ref: #59


sparks0:
Thanks for creating this but after a recent discussion about speed paths behaviour, we decided only path blocks should have this feature.

PR #68, opened 2018-10-24
Custom Crafters Expansion - Head Crafting

AndanteDevs:
Now craft more decorative blocks with the head crafting expansion pack!


AndanteDevs:
If somebody could make graphical representations of the dropper interface with these recipes in it would be much appreciated. Even taking a screenshot of the 3x3 grid while just having the items in would be fine.

sparks0:
This module uses the beta clocking standard. It also only has skulls you can create that seem a little randomly chosen to be honest. Not sure this is flesh out (no pun intended) enough.

AndanteDevs:
Menial module. Closing request.

Issue #71, opened 2018-10-24
Breaking a custom crafter drops a dropper named custom crafter


AndanteDevs:
### Expected Behavior
A custom crafter should drop 7 cobblestone, 1 redstone, 1 slightly damaged workbench and a dropper.

### Current Behavior
All that happens, only the dropper is named 'Custom Crafter' due to recent updates' changes to the way you break block entities.

### Possible Solution
Kill the dropped dropper item entity and re-summon a new one on custom crafter destruction.

### Steps to Reproduce
1. Make a custom crafter
2. Break a custom crafter

### Environment
Singleplayer
All other modules installed
Minecraft 1.13.2 (OptiFine E4_pre2)

misode:
Better solution would be to `data remove` the display tag

AndanteDevs:
Changed

Kroppeb:
I think this will just result in inconsistent behaviour. This will be fixed in 1.14. Maybe just leave it as is.

Bluefire-610:
Whilst this is definitely a bug I'd rate it as a minor one that does not break anything. We'll wait for 1.14 to fix the `destroy` argument

SpecialBuilder32:
This is an issue that can be attempted to be fixed by editing a nearby dropper entity when thr armor stand is broken. It's a minor edit that, although not being 100% successful  (if a player picks up the dropper before the clock pulse), I think would be worthwhile to include. The PRd enchant extractors does this.

Also, I don't know what 1.14 features you're looking at, but i dont believe any of them apply to this issue.

PR #72, opened 2018-10-24
Fixes issue #71

AndanteDevs:
Ref #71


Kroppeb:
You can't guarantee the item will get killed. This could duplicate droppers.

PR #75, opened 2018-10-24
Updated ".mcmeta" file to be consistent with other modules

Scommander:
Pack format changed to 0
Description changed to: "A Gamemode 4 Module"


misode:
It would make sense to either merge now or close now

PR #85, opened 2018-11-08
Update CONTRIBUTING.md

sparks0:
Added information on packaging a module for submission (deleting clock-related files)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@ There are several placeholder values in both file names and the files themselves
 - `MODULE_NAME` should be replaced with a display name for the module as this is visible to players such as `Bat Grenades` or `Enderman Support Class`. This should match the name on the website if the module already exists.
 
 ### Folders
-Within the `data` folder there are three folders for the different namespaces this module will use. If the module is an extension, you will need an additional module for the base module.
+Within the `data` folder there are three folders for the different namespaces this module will use. If the module is an extension, you will need an additional module for the base module. Because we merge modules with a copy of the base clock certain folders and files should not be included once the module is submitted but are included in the template so you can test your module.
 #### /minecraft/
 This folder simply contains code that lets Minecraft run the Gamemode 4 clocking system. Nothing in this folder needs to be changed.
 #### /gm4/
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -13,7 +13,7 @@ There are several placeholder values in both file names and the files themselves
 - `MODULE_NAME` should be replaced with a display name for the module as this is visible to players such as `Bat Grenades` or `Enderman Support Class`. This should match the name on the website if the module already exists.
 
 ### Folders
-Within the `data` folder there are three folders for the different namespaces this module will use. If the module is an extension, you will need an additional module for the base module.
+Within the `data` folder there are three folders for the different namespaces this module will use. If the module is an extension, you will need an additional module for the base module. Because we merge modules with a copy of the base clock certain folders and files should not be included once the module is submitted but are included in the template so you can test your module.
```
SpecialBuilder32:
```suggestion
Within the `data` folder there are three folders for the different namespaces this module will use. If the module is an extension, you will need an additional module for the base module. Because we merge modules with a copy of the base clock, certain folders and files should not be included once the module is submitted but are included in the template so you can test your module.
```

CONTRIBUTING.md
```diff
@@ -25,15 +25,21 @@ This folder should be renamed to `MODULE_ID`. All your module code should be wit
...
 - Names for the above two points should be as clear and relevant as possible to avoid confusion or two modules sharing the same name by mistake.
+- Lore on items should have the first word capitalized and no fullstop at the end of the sentence if there is only one sentence. Other punctuation (?!) can be added.
 - ANY FUNCTION that is run as an entity should have a comment on the first line stating the context of @s so it is clear what runs this file e.g. `#@s = cows with speed limit=..1 of red mushrooms`
-- ALL FUNCTIONS except those included with the template file should list the functions they are called from in a comment on the first or second line of the file e.g. `#run from main`. More complicated modules with subdirectories should specify the more specific file path e.g. `#run from zauber_cauldrons:cauldron/create`
+- ALL FUNCTIONS except those included with the template file should list the functions they are called from in a comment on the first or second line of the file e.g. `#Run from main`. More complicated modules with subdirectories should specify the more specific file path e.g. `#Run from zauber_cauldrons:cauldron/create`
```
SpecialBuilder32:
```suggestion
- ALL FUNCTIONS except those included with the template file should list the functions they are called from in a comment on the first or second line of the file e.g. `#run from main`. More complicated modules with subdirectories should specify the more specific file path e.g. `#run from zauber_cauldrons:cauldron/create`
```

PR #87, opened 2018-11-08
Replace armor stand with armour stand

misode:
This is necessary to match the site's name (https://gm4.co/modules/better-armour-stands)


sparks0:
Sorry but due to the clock changes I just made there were conflicts and Kropp was already working on a more recent version of this when I spotted it. I used that one instead :/

PR #88, opened 2018-11-08
Updating

sparks0:
Removed the gm4/tags/init_check and pulse_check files from the clock and put them back in with the module itself. This will improve the ease of testing (where the module will run in git-packaged format if a different module is in charge of running the clock.

Updated better_armour_stands to the site format.


misode:
@sparks0 Could you squash and rebase pull requests this large next time?

PR #89, opened 2018-11-08
Renamed `gm4_better_armor_stands` to `gm4_better_armour_stands`

Kroppeb:


sparks0:
Spelling of "Armour" is Americanized in the files themselves still and should be changed.

PR #97, opened 2018-11-10
Swap license from AGPL to GPL

SpiderRobotMan:
After some discussion it's been decided that AGPL enforces a restriction on server owners that we don't think is right. This change to normal GPL will strip that out but still keep the main things we wanted from a license.

I would like all past contributors to approve this before we merge it.


SpiderRobotMan:
@Kroppeb @kruthers @sparks0 @misode @Bluefire-610 @SpecialBuilder32 @AndanteDevs Please approve this if you are ok with this change.

PR #102, opened 2018-11-15
Removed a bunch of duplicate checks

Kroppeb:
This removes a bunch of repeated score checks


sparks0:
Special and I discussed this and we're not sure it actually creates a performance improvement. It adds more functions after all. However it DOES make things a little harder to read and would make setting up crafting recipes for other modules more confusing for module submitters so I'm going to close this.

If you can do some tests to benchmark and prove that this makes a significant improvement on performance please let us know and we'll re-review it.

PR #105, opened 2018-11-17
Addition of Liquid Tanks

SpecialBuilder32:
This is an incomplete version of the Liquid Tanks Base. I created a pull request to allow people to look for bugs or inefficiencies in what I've done so far, to share my current progress, and to document what I have so far.

This version still uses the outdated clocking standard, and includes all the clocking files that it needs to run. Those files will be removed at a later point.

To test this module (and I ask that you please do), download the files from the Pull Request and install **only** the liquid tanks datapack into your testing world.

The Liquid Tanks base comes packaged with Standard Liquids pack, and all items and recipes have not changed.


gm4_liquid_tanks/data/liquid_tanks/functions/level_report_ray.mcfunction
```diff
@@ -0,0 +1,9 @@
+#@s = all players at @s
+#run from main
+
+execute unless entity @s[tag=gm4_lt_found_tank] anchored eyes positioned ^ ^ ^.5 if block ~ ~ ~ hopper align xyz positioned ~0.5 ~ ~0.5 if entity @e[type=armor_stand,tag=gm4_liquid_tank,distance=..0.5] run function #liquid_tanks:level_report_check
```
Kroppeb: I don't see how `unless entity @s[tag=gm4_lt_found_tank]` can ever do something here

gm4_liquid_tanks/data/liquid_tanks/functions/liquid_value_update.mcfunction
```diff
@@ -0,0 +1,23 @@
...
+scoreboard players set @s gm4_lt_buffer 0
+scoreboard players operation @s gm4_lt_value < @s gm4_lt_max
+
+#if value of 0, set to empty
+execute if score @s gm4_lt_value matches ..0 at @s run function liquid_tanks:empty
+
+#calculating teleport distance
+scoreboard players operation last_liquid_height gm4_lt_disp_val = @s gm4_lt_disp_val
+scoreboard players operation liquid_height_calc gm4_lt_disp_val = @s gm4_lt_value
+scoreboard players set 10 gm4_lt_disp_val 10
```
Kroppeb: `scoreboard players set 10 gm4_lt_disp_val 10`
Can this be done in init? It feels weird to set a constant every 16 ticks for every liquid tank

gm4_liquid_tanks/data/liquid_tanks/functions/liquid_value_update.mcfunction
```diff
@@ -0,0 +1,23 @@
...
+scoreboard players operation liquid_height_calc gm4_lt_disp_val *= 10 gm4_lt_disp_val
+scoreboard players operation liquid_height_calc gm4_lt_disp_val /= @s gm4_lt_max
+scoreboard players operation @s gm4_lt_disp_val = liquid_height_calc gm4_lt_disp_val
+scoreboard players operation liquid_height_calc gm4_lt_disp_val -= last_liquid_height gm4_lt_disp_val
+
+#teleport proper number of times.
+execute if score liquid_height_calc gm4_lt_disp_val matches 1.. at @s run function liquid_tanks:teleport_display_up
+execute if score liquid_height_calc gm4_lt_disp_val matches ..-1 at @s run function liquid_tanks:teleport_display_down
```
Kroppeb: I suggest moving the score into the other armorstand. This way `teleport_display_up` and `teleport_display_down` don't have to constantly do an `execute as`

gm4_liquid_tanks/data/liquid_tanks/functions/level_report_ray.mcfunction
BluePsychoRanger: If the player is not looking at a liquid tank and thus does not have the tag "gm4_lt_found_tank" then the execute will stop processing the rest of the command, saving on lag.

Kroppeb: Yes, but the only places this tag is set is in functions called from inside this function and it gets removed at the end of this function. Therefor the tag will never be set when it checks for this on line 4

Kroppeb:
Whenever people create a liquid tank they might be standing on top to throw in the glass block. This will convert the tank into an xp tank and might confuse/annoy some people.

gm4_liquid_tanks/data/liquid_tanks/functions/level_report_ray.mcfunction
BluePsychoRanger: Oh, I see what you mean.

SpecialBuilder32: You are right in that the first step of the ray will never have that tag set. The rest can though.

gm4_liquid_tanks/data/liquid_tanks/functions/liquid_value_update.mcfunction
Kroppeb: I meant to move the score at the end of the calculation. It’s just temporary.

SpecialBuilder32: Oh I see. In that case, I don't think it would gain any benefit. Copying the score would require at least one selector resolution, and then changing the executor would be another. Currently there is just one in the actual teleport command. It is very rare that the armor stand need to teleport more than one step up or down, so on most every occasion, the selector would only be resolved once.

Nathan22211:
i tested it the tanks and they work fine. i would suggest that any liquids that do 1/3 /30 it lists it as n/90 bowls in the instance of soups. and n/90 bottles in the instance of water bottles and potions

Nathan22211:
wait what about suspicious stew!?

PR #111, opened 2018-11-21
New Module: Sharpening Tools

BluePsychoRanger:
Adds a new item to repair durability to weapons and tools. 

Documentation:
https://docs.google.com/document/d/1p_hpJGLntYgh3ylXDnmCkFAx4mKsrd-IU6ihZt5EE94/edit?usp=sharing


sparks0:
I felt this was unbalanced until I noticed that in order to keep enchantments on a tool, an enhanced, more expensive sharperner is needed. Kinda cool because it allows early-game use of the module as well as late-game.

My one thought of how this could be exploited is that it provides a way to remove unwanted enchantments from a tool (be that a bad enchanting table session or a found item with a curse of vanishing or binding).

BluePsychoRanger:
There's already a mechanic in the game that allows for removing enchantments from an item: simply combining 2 damaged items. But I see how it could be exploited. The sharpening tool doesn't affect non-damaged items, but it could be modified to require a certain amount of damage. I'm for this change, but I feel like it would confuse some people, as the module would appear to be "broken" when it doesn't repair a damaged tool.

BluePsychoRanger:
Perhaps if the sharpening tool sharpens an item with a curse, it removes the enchantments, but the curse causes damage to the tool rather than repair it? The damage could be equal to what it would've repair, or maybe half of what it would've repair.

sparks0:
I'm torn about this module. I find nothing really wrong with it, but I'm not sure people would use it when there are already methods of repairing tools with anvils as well as mending enchants.

sparks0:
After discussing with staff we have decided that because of anvils and mending this module wouldn't really be used by survival players except maybe a little at the very start of the game. Sorry. I really liked the module idea but I'm gonna have to decline the PR <3

Issue #113, opened 2018-11-24
Allow Netherwart to be planted by Boots of Ostara


coolsa:
The Boots of Ostara should be able to replant nether wart, another type of full block crop. Similar to the other crops the Boots may replant, nether wart is a full block crop and harvesting requires the plant to be broken and replanted.

coolsa:
Now with the pull request #114 merged and the features requested implemented, I am closing this suggestion.

Issue #109, opened 2018-11-19
Allow Amorstands to use Boots of Ostara


Compsogbrickus:
Adding grass-placing functionality to armorstands wearing Boots of Ostara would be a helpful and realistic change, as well as allowing for a variety of non-afk farms (sheep, spawner minecarts, etc.).

Bluefire-610:
Coolsa's code for this was merged :)

PR #114, opened 2018-11-24
[Boots of Ostara] Added Armor Stand grass planting, and Player Nether Wart Planting

coolsa:
double commit, but each commit adds totally unrelated features. Let me know if this is ok. I tested, it didn't seem to break anything in a single player world. Small changes, but at least one was not requested by me.

Closes suggestions #109 and #113


sparks0:
Nether wart idea is cool. Can I ask why only grass planting with armour stands and why specifically grass? What's the intended use?

coolsa:
The suggestion of armour stands planting grass was mostly suggested for wool farms. With an armour stand with the boots on in the same block as a sheep, you wouldn't need to have a second grass block, compacting things a bit. It could make for some more decorative and compact sheep farms. 
#109 also mentioned some interesting interactions with spawner minecarts, but I was unable to test this.

PR #116, opened 2018-11-25
New Module: Beacon Cores

Chopper2112:
Adds the ability to modify a beacon so that it stops Phantoms spawning in the local area. Craft a Phantom Core in a Custom Crafter with 8 phantom membranes and an elytra in the center, then throw that on top of a beacon.

Documentation: https://docs.google.com/document/d/1ez7jdyId4ZeEPNPVulfX46FpZvB1K-Q1cJWZdK8llLw/edit?usp=sharing


Chopper2112:
I'll make the box check change since I doubt any of the other admins will be against that, as well as the slowly damaging phantoms & custom crafting fix.

I'm not sure about the damaging all hostile mobs idea. This would be a cool and useful addition, but it does seem pretty OP, especially if the crafting recipe becomes cheaper. I'll wait on more feedback before making this change, but it's a cool idea.

sparks0:
In the end this module inspired Phantom Scarecrows and I've given you a credit link on that module's page on the website but we decided not to accept this module. Thanks for your work and sorry we couldn't accept it!

PR #119, opened 2018-12-02
Fix #100 - Heart canister bug

### LINKED ISSUE #100: Heart Canisters causing damage to player upon health_boost reapply
### Expected Behavior

Heart Canisters should give a player extra health when in their inventory.
### Current Behavior

When the player is above default health (20 health points), their extra hearts are constantly lost, as if they were being damaged by some mob. This is due to a side-effect of applying the health_boost status effect that reset's the players current health.
See Bug : https://bugs.mojang.com/browse/MC-25088
### Possible Solution

Until the aforementioned bug is resolved, we would need some annoying trickery to give them the right health boost. 
A potential solution would be to give the player an infinite health_boost based on their number of heart canisters, and clear it if that number changes.
Unfortunately, it seems like we will once again have the annoying issue of whenever your heart-canister number changes, you have to regenerate all your health.
### Steps to Reproduce

(This pertains specifically to the issue's existence in heart_canisters, not the bug as a whole)
1. Obtain at least one heart canister
2. Heal yourself above 20 hearts
3. Watch as your newfound health repeatedly falls back to nothing.

### Environment

The underlying bug affects 1.13.2 vanilla.

misode:


SpiderRobotMan:
this is a test

PR #125, opened 2018-12-18
Add Disassemblers

federick0090:
Disassemble tools using Disassemblers. The more durability the item has, the more items it can give back


sparks0:
I feel like I already tested and approved this. What happened to the previous pull request?

That said this looks good to me.

Nathan22211:
i've tested this before and it worked fine but i only tested it with diamond armor and in 1.13.2.

PR #134, opened 2018-12-26
Make Spawner Minecarts Invulnerable

Bluefire-610:
The latest iteration of Spawner Minecarts allowed them to take hits and only die due to mass damage like explosions, however, this prevents them from going through nether portals. This PR removes that damage system and makes the carts invulnerable.


kyrkis:
Will they be able to go through portals with this fix?

Bluefire-610:
> Will they be able to go through portals with this fix?

Yes

ToffeeMax:
This is one hot and steamy addition :steam:

PR #137, opened 2018-12-29
Create Squids by Dropping a lightning rod on Dried Kelp Block

ToffeeMax:
Gamemode 4 Public Server V has an issue where due to the player count, spawning of squids in oceans is limited, which led to wanting a way to create them that was not just a spawner. Building on the idea that shocking a purpur block give a shulker, the most logical thing to create a squid that was sea related was the  dried kelp block.

Notes:
- Checks for Dried Kelp Block below, like purpur
- Squid is spawned with a few seconds of resistance so it does not die from suffocation or fire caused by lightning rods

### DIFF
--- a/gm4_lightning_rods/data/lightning_rods/functions/create_squid.mcfunction
+++ b/gm4_lightning_rods/data/lightning_rods/functions/create_squid.mcfunction
@@ -1,3 +1,3 @@
 setblock ~ ~-1 ~ air
-summon squid ~ ~-1 ~ {ActiveEffects:[{Id:10b,Amplifier:3b,Duration:100,ShowParticles:1b}]}
+summon squid ~ ~-1 ~ {ActiveEffects:[{Id:10b,Amplifier:3b,Duration:100,ShowParticles:0b}]}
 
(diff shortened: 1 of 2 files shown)


gm4_lightning_rods/data/lightning_rods/functions/create_squid.mcfunction
```diff
@@ -0,0 +1,3 @@
+setblock ~ ~-1 ~ air
+summon squid ~ ~-1 ~ {ActiveEffects:[{Id:10b,Amplifier:3b,Duration:100,ShowParticles:1b}]}
```
misode: Don't you want `ShowParticles:0b`?

ToffeeMax: Yep!

alluet:
I personally think 5 seconds is way too long. Fire from lightning affects converted mobs in vanilla, and can be completely avoided with the right setup, so the resistance should really only be for the lightning damage. It takes a while for squid to start suffocating on land as well.

sparks0:
Squids can already be generated using Sunken Treasure. I don't want to add this feature just to solve a public server issue.

PR #138, opened 2019-01-01
Added Invisible and Visible Tags

ToffeeMax:
I believe this was an old feature - but if not:

- Type "invisible" in Book to make armour stand invisible
- Type "visibile"in book to make armour stand visible
- Was created to not have to use particles pack to make invisible and have a particle


iGalaxyYT:
I've been wanting this for so long 👀

sparks0:
Issue with this is the "invisible armorstands glow when holding a blank book" is a feature of particles pack not better armour stands so it would be possible to completely lose your stands. The glowing feature would need to be moved to the base. Otherwise I'd be ok with this.

ToffeeMax:
@sparks0  - I moved the functionality from particles pack to better armor_stands as per your request :)

ToffeeMax:
@sparks - Apologies, should have actually thought instead of bringing over old functionality. Has been tested and works. Only gives to armor_stands with nbt tag "Invisible:1b" and works with any written book

ToffeeMax:
@misode - It made sense. Because it is annoying having to empty the book to then retype

AndanteDevs:
I know 1.14 GM4 is not supported yet, but you can't have an empty book in the new update after you've written in it before. It simply won't let you. If players get used to this, it won't be an issue, as such, but it's better to set the format now instead of confusing players when things don't work later.

PR #144, opened 2019-01-07
Update spawner minecarts to new standard and add pack.mcmeta information

misode:
The spawner minecarts datapack was merged with the master branch, but still uses the old module standard. This pull request fixes that and also adds the required information in `pack.mcmeta` for it to show up on the website.


Bluefire-610:
I didn't know it was merged into master. The version in master is most likely a very old version of it and should not be used. @SpecialBuilder32 did you merge this?

Bluefire-610:
Update on this: Spawner Minecarts wasn't intentionally merged into master. We'll have to compare which of both versions is the most recent one where max fuel is at 2000 and delete the other version (should be the one in master)

Bluefire-610:
The version in master is invalid and should not be in there, it will be deleted once spawner minecarts is released. I'm closing this pr.

PR #150, opened 2019-02-12
Chairs

Dennis-0:
Throw a saddle in a stair block to make it a chair


gm4_chairs/data/chairs/functions/kill.mcfunction
```diff
@@ -0,0 +1,7 @@
+# @s = @e[tag=gm4_chairs] and NOT inside stair blocks
+#execute at @s align xyz run tp @e[nbt={RootVehicle:{Entity:{id:"minecraft:pig"}}},dx=0,dz=0,dy=2] ~ ~.4 ~
```
Kroppeb: Why would you want to tp the player and why wouldn't you use `@p` or something. Also why is this in a comment

gm4_chairs/data/chairs/functions/kill.mcfunction
```diff
@@ -0,0 +1,7 @@
+# @s = @e[tag=gm4_chairs] and NOT inside stair blocks
+#execute at @s align xyz run tp @e[nbt={RootVehicle:{Entity:{id:"minecraft:pig"}}},dx=0,dz=0,dy=2] ~ ~.4 ~
+tp @s ~ ~.5 ~
```
Kroppeb: Why is the pig tp-ed?

gm4_chairs/data/chairs/functions/kill.mcfunction
```diff
@@ -0,0 +1,7 @@
+# @s = @e[tag=gm4_chairs] and NOT inside stair blocks
+#execute at @s align xyz run tp @e[nbt={RootVehicle:{Entity:{id:"minecraft:pig"}}},dx=0,dz=0,dy=2] ~ ~.4 ~
+tp @s ~ ~.5 ~
+execute at @s run summon minecraft:area_effect_cloud ~ ~ ~ {Duration:100,Tags:["gm4_chairs_temp"]}
```
Kroppeb: I don't see why a marker needs to be summoned. `at @s` is `positioned ~ ~.5 ~` atm

gm4_chairs/data/chairs/functions/kill.mcfunction
```diff
@@ -0,0 +1,7 @@
+# @s = @e[tag=gm4_chairs] and NOT inside stair blocks
+#execute at @s align xyz run tp @e[nbt={RootVehicle:{Entity:{id:"minecraft:pig"}}},dx=0,dz=0,dy=2] ~ ~.4 ~
+tp @s ~ ~.5 ~
+execute at @s run summon minecraft:area_effect_cloud ~ ~ ~ {Duration:100,Tags:["gm4_chairs_temp"]}
+kill @s
```
Kroppeb: by merging `Health:0b`, the entity won't drop porkchop, you have to summon the saddle back though but killing porkchop can kill the wrong one and stuff like that.

gm4_chairs/data/chairs/functions/main.mcfunction
```diff
@@ -0,0 +1,8 @@
+#saddle item on a stair block -> spawn pig
+execute as @e[type=minecraft:item,nbt={Item:{id:"minecraft:saddle"}}] at @s if block ~ ~ ~ #minecraft:stairs[half=bottom] positioned ~ ~-1 ~ align xyz unless entity @e[tag=chairs,dx=0,dz=0,dy=0] run function chairs:saddle
+
+#kill pig/saddle if stair block broken 
+execute as @e[tag=gm4_chairs] at @s unless block ~ ~1 ~ #minecraft:stairs run function chairs:kill
+
+#OPTIONAL: position correction, offset caused by holding a carrot on a stick
+#execute as @e[tag=gm4_chairs] at @s align xyz if entity @p[dx=0,dz=0,dy=5] run tp @s ~.31 ~.61 ~.5
```
Kroppeb: This command doesn't take into account the different rotations of the pig

gm4_chairs/data/chairs/functions/main.mcfunction
```diff
@@ -0,0 +1,8 @@
+#saddle item on a stair block -> spawn pig
+execute as @e[type=minecraft:item,nbt={Item:{id:"minecraft:saddle"}}] at @s if block ~ ~ ~ #minecraft:stairs[half=bottom] positioned ~ ~-1 ~ align xyz unless entity @e[tag=chairs,dx=0,dz=0,dy=0] run function chairs:saddle
```
Kroppeb: All functions in `chairs:saddle` use `at @s`. This can be done here though.

gm4_chairs/data/chairs/functions/kill.mcfunction
Dennis-0: The pig is slightly inside the block under the stairs, so if the pig is killed the saddle item would glitch inside the block. it's much smoother if the pig is actually killed half a block up

Kroppeb: * As said in my other comments: I think killing the pig without drops and then summoning the saddle is better. This would reduce this file to 2 commands
* Wouldn't the block the pig is in be air, as it's no longer in a stair?

gm4_chairs/data/chairs/functions/kill.mcfunction
Dennis-0: That command would get the player off the saddle, so if you break the stairs you get right where you were before (and again, about half a block up to prevent you going into the block underneath).
@e was because I was trying something with entities as well and commented it out if I wanted to do something with it later, should've left it out for this upload

gm4_chairs/data/chairs/functions/kill.mcfunction
Dennis-0: Changed it to the 2 commands as you said, downside is that it now just appears instantly instead of going in a random direction as if it popped off

sparks0:
Cute little module that could add extra coolness to small builds.
I'm willing to be overruled about this but here's my thought:
If some stairs can be sat on in the game and not others and there is no visual difference, doesn't that seem very inconsistent to players? Can there be some indicator that you can sit on this particular stair? Could be a carpet or visible saddle to mark it differently or even a tooltip when the chair is looked on saying "right-click to sit" or something.

Dennis-0:
> Cute little module that could add extra coolness to small builds.
> I'm willing to be overruled about this but here's my thought:
> If some stairs can be sat on in the game and not others and there is no visual difference, doesn't that seem very inconsistent to players? Can there be some indicator that you can sit on this particular stair? Could be a carpet or visible saddle to mark it differently or even a tooltip when the chair is looked on saying "right-click to sit" or something.

There's a visible saddle when a stair is a chair

PR #152, opened 2019-03-23
Adds a new type of custom crafter.

TheRedKeks:
Adds a custom crafter that acts like a vanilla crafting table, but you can add custom recipes with items that have a  nbt-tag as input and output, for example you can make it so that someone has to lay a broken elyrta and a written book with a spell in the crafting grid and he/she gets a half repaired elytra back.
To get the custom crafter you have to throw a diamond block on top of a crafting table.


sparks0:
Cool module! I'm worried that it deletes items placed in filler slots so not quite suitable for Gamemode 4 but thanks for the submission, keep playing with ideas, you clearly have some good ones. :)

PR #154, opened 2019-04-04
Change parent of advancements from custom crafter required module

BluePsychoRanger:
Some advancements should be a child of the custom crafters advancement because they required things to be crafted or converted from a custom crafter into another block.


Bluefire-610:
This edit history is a mess but yeah, that change should be done! Thanks for the commit!

PR #165, opened 2019-04-07
SCUBA Gear - Enhanced Helmet

BluePsychoRanger:
Adds an enhanced SCUBA Helmet that is the same as the normal helmet in terms of conduit power, but is enchanted with aqua affinity and respiration III. In the current version, the player must have both the helmet and the tank to gain the conduit effect. I think this update would make players use the helmet more, because, unlike the normal SCUBA Helmet, the enhanced helmet is not completely useless without the tank.


Bluefire-610:
We're sorry but we don't think this enhanced helmet is neccessary.

PR #167, opened 2019-04-10
Added cement mixer

sparks0:
This is a liquid tanks expansion which allows water tanks to turn single concrete powder blocks into concrete at the cost of 1 third of a water bucket per block.

The aim of this module is to allow players to create an automatic way to convert powder to concrete but to make them work for it first by creating a redstone contraption that feeds the concrete powder in steadily (and optionally tops up the water too)

### DIFF
--- a/gm4_cement_mixers/data/cement_mixers/functions/item_fill.mcfunction
+++ b/gm4_cement_mixers/data/cement_mixers/functions/item_fill.mcfunction
@@ -1,19 +1,4 @@
 #@s = liquid tank with item in first slot
 #run from liquid_tanks:item_process
 
-execute if score @s[tag=gm4_lt_water] gm4_lt_value matches 1.. if block ~ ~ ~ hopper{CustomName:"{\"text\":\"Water Tank\"}",Items:[{Slot:0b,id:"minecraft:black_concrete_powder"}]} run function cement_mixers:item_fill/black_concrete
-execute if score @s[tag=gm4_lt_water] gm4_lt_value matches 1.. if block ~ ~ ~ hopper{CustomName:"{\"text\":\"Water Tank\"}",Items:[{Slot:0b,id:"minecraft:blue_concrete_powder"}]} run function cement_mixers:item_fill/blue_concrete
-execute if score @s[tag=gm4_lt_water] gm4_lt_value matches 1.. if block ~ ~ ~ hopper{CustomName:"{\"text\":\"Water Tank\"}",Items:[{Slot:0b,id:"minecraft:brown_concrete_powder"}]} run function cement_mixers:item_fill/brown_concrete
-execute if score @s[tag=gm4_lt_water] gm4_lt_value matches 1.. if block ~ ~ ~ hopper{CustomName:"{\"text\":\"Water Tank\"}",Items:[{Slot:0b,id:"minecraft:cyan_concrete_powder"}]} run function cement_mixers:item_fill/cyan_concrete
-execute if score @s[tag=gm4_lt_water] gm4_lt_value matches 1.. if block ~ ~ ~ hopper{CustomName:"{\"text\":\"Water Tank\"}",Items:[{Slot:0b,id:"minecraft:gray_concrete_powder"}]} run function cement_mixers:item_fill/gray_concrete
-execute if score @s[tag=gm4_lt_water] gm4_lt_value matches 1.. if block ~ ~ ~ hopper{CustomName:"{\"text\":\"Water Tank\"}",Items:[{Slot:0b,id:"minecraft:green_concrete_powder"}]} run function cement_mixers:item_fill/green_concrete
-execute if score @s[tag=gm4_lt_water] gm4_lt_value matches 1.. if block ~ ~ ~ hopper{CustomName:"{\"text\":\"Water Tank\"}",Items:[{Slot:0b,id:"minecraft:light_blue_concrete_powder"}]} run function cement_mixers:item_fill/light_blue_concrete
-execute if score @s[tag=gm4_lt_water] gm4_lt_value matches 1.. if block ~ ~ ~ hopper{CustomName:"{\"text\":\"Water Tank\"}",Items:[{Slot:0b,id:"minecraft:light_gray_concrete_powder"}]} run function cement_mixers:item_fill/light_gray_concrete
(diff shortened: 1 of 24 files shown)


ToffeeMax:
I like this. I believe It was suggest in Labs discord for some other methods, but I am all up for expansions.

sparks0:
> I like this. I believe It was suggest in Labs discord for some other methods, but I am all up for expansions.

I think it was discussed in labs in a variety of ways and we considered an LT expansion. I like this method because unlike just throwing the items in water to change them, this requires a bit of setup to start doing in bulk. It's been sitting in needs-developer for a while on labz now so I thought I'd do it :)

gm4_cement_mixers/data/cement_mixers/functions/item_fill.mcfunction
```diff
@@ -0,0 +1,19 @@
+#@s = liquid tank with item in first slot
+#run from liquid_tanks:item_process
+
+execute if score @s[tag=gm4_lt_water] gm4_lt_value matches 1.. if block ~ ~ ~ hopper{CustomName:"{\"text\":\"Water Tank\"}",Items:[{Slot:0b,id:"minecraft:black_concrete_powder"}]} run function cement_mixers:item_fill/black_concrete
```
Kroppeb: * `if score @s[tag=gm4_lt_water] gm4_lt_value matches 1..`
    * This should be done once and then call a second function with all the recipes.
* `CustomName:"{\"text\":\"Water Tank\"}"`
    * Is this really needed. I'm guessing this is the way it's done in LT itself but I don't like it.

gm4_cement_mixers/data/cement_mixers/functions/item_fill/black_concrete.mcfunction
```diff
@@ -0,0 +1,3 @@
+replaceitem block ~ ~ ~ container.0 black_concrete 1
+scoreboard players remove @s gm4_lt_buffer 1
+tag @s add gm4_lt_fill
```
Kroppeb: `tag @s add gm4_lt_fill`

Concrete can't be drained. There is no need for this tag.
I even think we should add a separate hook for filling(/draining) items where their counterpart can't be drained(/filled) again. The current system would need a module tick where the tank has no items for it to be able to make concrete again

gm4_cement_mixers/data/cement_mixers/functions/item_fill.mcfunction
sparks0: First point is a good one. Second one is something Special did in order to stop people breaking and hopper and replacing it with a differently named one really quickly (a bug someone found)

gm4_cement_mixers/data/cement_mixers/functions/item_fill/black_concrete.mcfunction
SpecialBuilder32: The `gm4_lt_fill` tag informs the tank that the last action it performed was filling an item with liquid, preventing the next tick from attempting to drain that same item. Even if concrete cant be drained, the tag lowers the number of checks the tank performs; it dosen't go through the list of drainable items at all. This partially increases performance, even though it isn't needed to prevent a loop.
The current system doesn't need a tick with an empty hopper to make concrete; The tank can do the same operation (filling or draining) in consecutive module pulses.

Kroppeb: Mmh, I thought the tag was only removed if there is no item is slot 0 and this module places the Concrete there.

SpecialBuilder32: The tag Is only removed if the first slot is empty. That does not prevent the module from placing another concrete (filling operation) though, only from attempting to drain the item (as if it were a bucket) the next clock pulse around.

gm4_cement_mixers/data/cement_mixers/functions/item_fill.mcfunction
Kroppeb: That someone is me.
The bug was that there were no tags, and only the name is used. Currently both name and tag have to match.

I kinda told special about this after the bug fix. Again not a bug but could be changed. https://github.com/Gamemode4Dev/GM4_Datapacks/pull/105#r31746669

SpecialBuilder32: Since the module has to examine block nbt for the item content anyways, it dosen't hurt to also check the name. This also helps to double-check that something funky hasn't gone on with the tank and that the user can always rely on the name of the tank to accurately reflect what they can get from it.

gm4_cement_mixers/data/cement_mixers/functions/item_fill/black_concrete.mcfunction
Kroppeb: That's the thing. If you want to automate this you have to fill the concrete, wait a module tick, and only then you can fill it with water again.

sparks0: He's right. Just tested it and swapping an empty just-drained bucket with powder causes the tank to do nothing.

sparks0: Actually just removed the tag from all of the fill functions and the tank still doesn't respond to the powder until the tank detects the slot is empty.

Kroppeb: > Actually just removed the tag from all of the fill functions and the tank still doesn't respond to the powder until the tank detects the slot is empty.

Yes, this is because the standard liquids adds the tag. That's why I made this comment.
> I even think we should add a separate hook for filling(/draining) items where their counterpart can't be drained(/filled) again. The current system would need a module tick where the tank has no items for it to be able to make concrete again

The other way around (water after concrete) should work with the tag removed.

BluePsychoRanger:
I've tested this and it works as it should. With liquid tanks out now, when is the ETA of this module release?

Bluefire-610:
> I've tested this and it works as it should. With liquid tanks out now, when is the ETA of this module release?

NOW!

PR #170, opened 2019-04-12
[1.14] Update Text Strings

AndanteDevs:
Update to new `'` string open/close format


AndanteDevs:
Pushed to wrong branch, oops

PR #172, opened 2019-04-12
Move root advancement to base

AndanteDevs:
Moves `gm4:root` to the base pack, as it is the same in all modules.

Sorry about the massive amount of commits, I'm doing this painstakingly on mobile.


AndanteDevs:
Test

PR #175, opened 2019-04-18
Cement mixer & LT merge

Kroppeb:
Add's the `Liquid Tanks` module to the `Cement Mixer` branch making it easier to test the module.


sparks0:
This just means that changes to LT will cause the LT in this PR to be outdated. It's not hard to download two PRs to test.

Kroppeb:
I don't know how you guys test but I just git pull a branch and test on there. Downloading both would be either a mess on my end or a merge on a local branch anyway. If I then were to make a modification to the code I would have to revert the merge, do the modification and redo the merge to see if I din't break anything.

PR #178, opened 2019-04-19
Fix #96 - Trapped signs clearing text

### LINKED ISSUE #96: Trapped sign loses its text on servers or after relogging
When editing a trapped sign, the sign doesn't actually save it's data. It only appears visually in single player.

misode:
This pull request fixes #96 by only modifying the sign data after the user has finished editing the sign. This has as consequence that you can't have an empty trapped sign. A sign that gets broken before the user finishes editing the sign will still drop its items.


misode:
Both recipes support multi-crafting. The secret trapped sign recipe is shapeless.
image image

sparks0:
I just tested this in 1.13 on the stream server. The recipes to create signs work but nothing is being applied to the signs and there is no entity in the sign. It doesn't appear to be placing it as "trapped"

PR #179, opened 2019-04-19
Remove Diamond Recipes

ToffeeMax:
Due to the current nature (or intended) of villagers in 1.13 selling diamond gear and tools, and how that late game play means with little time and investment, a diamond farm is possible using them and dissembling their wares.
This PR:

- Removes all diamond recipes for tools and armour
- Removes the relative advancement for disassembling a diamond chest-plate
- Change pack.mcmeta to reflect that diamonds are no longer correct.

(I am not sure if I am set up this PR correctly, as it contains 3 other commits of me updating my base branch, If I have done this wrong-  let me know)


sparks0:
Since we were considering adding this back in future, how about we leave the files and simply comment-out the lines checking for those recipes so adding them back saves time?

ToffeeMax:
That does make sense. I will do that

sparks0:
10 days without an update and this is duplicated by https://github.com/Gamemode4Dev/GM4_Datapacks/pull/183 so I'm going to close it.

sparks0:
Blue and I decided that for 1.13, the diamond rates should be heavily reduced but still exist since it's strange to only have diamond tools not be disassemblable.

PR #180, opened 2019-04-19
Fix #149 - Cave air and void air not handled like air

### LINKED ISSUE #149: Cave Air not handled like air by most modules.
Most modules don't treat Cave Air the same as air. As a result modules with air checks don't work as intended in caverns, natural ponds, ravines, dungeons…
### Expected Behavior
The Module should treat Cave Air just like normal air.

### Current Behavior
Cave Air is treated like any non air block.

### Possible Solution
Create a block tag #air, maybe even include it in the base.

### Affected Modules (extend this list in the comments)
Shroomites (spores see cave air as a solid block)
Phantom Scarecrows (rockets explode in cave air)

misode:


ToffeeMax:
Cool. Using a tag to fix it is a good trick :)

Kroppeb:
Why not use a single tag?

misode:
> 
> 
> Why not use a single tag?

Do you mean `#gm4:air`? Imo it doesn't really matter and because some modules (like phantom scarecrows and shroomites) already have their block tags that just had to be expanded. Keeping all module related block tags under their own namespace seems like a good idea.

Kroppeb:
(But it uses more memory :( )

PR #181, opened 2019-04-21
Update Orbis to 1.14

misode:
Updated the Orbis module and expansions using a loot table to detect the biome. The current datapacks work, but it needs some more changes regarding structure spawning and chest/spawner population.

*This PR is not yet ready for merging*


sparks0:
This PR still a draft?

PR #192, opened 2019-05-03
[Submission] Damaging Stonecutters

AndanteDevs:
Damage players and mobs when on a stonecutter

*Original idea: @BluePsychoRanger*


gm4_damaging_stonecutters/data/damaging_stonecutters/functions/damage.mcfunction
```diff
@@ -0,0 +1,11 @@
...
+#@s - @e[type=!#damaging_stonecutters:no_target]
+#called by damaging_stonecutters:check_damage
+
+#damage entity
+effect give @s minecraft:absorption 1 0 true
+effect give @s[type=!#damaging_stonecutters:undead] minecraft:instant_damage 1 0 true
+effect give @s[type=#damaging_stonecutters:undead] minecraft:instant_health 1 0 true
+
+#audiovisuals
+function damaging_stonecutters:audiovisuals
```
misode: Instead of using a different function, put the commands in here directly.

AndanteDevs:
*And yes, I do realise this is running two block checks at every alive (logically, not technically) entity in the world, every tick*

gm4_damaging_stonecutters/data/damaging_stonecutters/functions/init.mcfunction
```diff
@@ -0,0 +1,17 @@
+#announce module installation
+tellraw @a[gamemode=creative] [{"translate":"%1$s","with":["[GM4]: ",{"translate":"text.gm4.prefix"}]},{"translate":"%1$s","with":["Installing Damaging Stonecutters...",{"translate":"text.gm4.modules.update.installing","with":["Damaging Stonecutters"]}]}]
+execute unless entity @p run say GM4: Installing Damaging Stonecutters...
```
misode: Use `@a[limit=1]`
```suggestion
execute unless entity @a[limit=1] run say GM4: Installing Damaging Stonecutters...
```

gm4_damaging_stonecutters/data/damaging_stonecutters/functions/init.mcfunction
```diff
@@ -0,0 +1,17 @@
...
+#declare and initialise scoreboards and settings
+scoreboard players set update_happened gm4_up_check 1
+scoreboard players set damaging_stonecutters gm4_modules 1
+scoreboard players set damaging_stonecutters gm4_clock_tick 0
+
+scoreboard objectives add gm4_ds_dmg_tick dummy
+
+#announce success
+tellraw @a[gamemode=creative] [{"translate":"%1$s","with":["[GM4]: ",{"translate":"text.gm4.prefix"}]},{"translate":"%1$s","with":["Damaging Stonecutters Installed!",{"translate":"text.gm4.modules.update.installed","with":["Damaging Stonecutters"]}]}]
+execute unless entity @p run say GM4: Damaging Stonecutters Installed!
```
misode: Use `@a[limit=1]`
```suggestion
execute unless entity @a[limit=1] run say GM4: Damaging Stonecutters Installed!
```

gm4_damaging_stonecutters/data/damaging_stonecutters/tags/entity_types/no_target.json
```diff
@@ -0,0 +1,41 @@
+{
```
misode: I suggest using the tag `immune`, following `#minecraft:wither_immune`

gm4_damaging_stonecutters/data/damaging_stonecutters/functions/pulse_check.mcfunction
```diff
@@ -0,0 +1,4 @@
+#compares assigned run tick to current tick and calls main
+#execute if score current_tick gm4_clock_tick = damaging_stonecutters gm4_clock_tick run function damaging_stonecutters:main
+
+execute as @e[type=!#damaging_stonecutters:no_target] at @s run function damaging_stonecutters:check_block
```
misode: You're only damaging entities every 10 ticks anyways, so a slower clock should be used (perhaps every 8 ticks). This will not only make the module more efficient, but will also simplify the code a lot while getting rid of the scoreboard.

gm4_damaging_stonecutters/data/damaging_stonecutters/functions/damage.mcfunction
AndanteDevs: That's a remnant from when the audiovisuals were used in more than one function

gm4_damaging_stonecutters/data/damaging_stonecutters/functions/pulse_check.mcfunction
AndanteDevs: I used a separate clock so that it damaged you the second you went onto the stonecutter - it felt better without a delay

Bluefire-610:
Vexes and fishing rod bobbers should also be immune.
Checking @e every gametick is a bit much, there must be a better way of doing this.

AndanteDevs:
The `enter_block` advancement criteria, but that's just for players 😕

AndanteDevs:
> Vexes [...] should also be immune

Why vexes?

Bluefire-610:
> > Vexes [...] should also be immune
> 
> Why vexes?

Vexes don't get damaged by any blocks they fly through.

AndanteDevs:
Ah, makes sense

misode:
Mobs could be damaged on a slower clock and player could use the `enter_block` advancements if you really want that.

sparks0:
Yeah I think this needs to run on a much slower clock because of the broad entity checks. If you don't think it looks right on a slower clock then maybe this isn't a module that should be added to GM4 at this time.

gm4_damaging_stonecutters/data/damaging_stonecutters/functions/damage.mcfunction
```diff
@@ -0,0 +1,14 @@
+#damages entity
+#@s - @e[type=!#damaging_stonecutters:stonecutter_immune]
+#called by damaging_stonecutters:check_damage
+
+#damage entity
+effect give @s minecraft:absorption 1 0 true
+effect give @s[type=!#damaging_stonecutters:undead] minecraft:instant_damage 1 0 true
```
SpecialBuilder32: You could use the wither effect to avoid needing to check if the mob is undead.

gm4_damaging_stonecutters/data/damaging_stonecutters/functions/damage.mcfunction
```diff
@@ -0,0 +1,14 @@
...
+effect give @s[type=#damaging_stonecutters:undead] minecraft:instant_health 1 0 true
+
+#audiovisuals
+particle minecraft:block minecraft:stonecutter ~ ~ ~ .25 0 .25 0 10 normal @a
+particle minecraft:block minecraft:redstone_block ~ ~ ~ .25 0 .25 0 10 normal @a
```
SpecialBuilder32: I'm not to sure about the "blood" particles used here. No other source of damage in the game uses such an effect.

gm4_damaging_stonecutters/data/damaging_stonecutters/functions/damage.mcfunction
AndanteDevs: We *did* use the wither effect at first, however we found too many downsides and just copied the code from heart canisters

gm4_damaging_stonecutters/data/damaging_stonecutters/functions/damage.mcfunction
AndanteDevs: I agree with you here, someone else suggested it (I can't remember who). Should I remove the stonecutter particle too?

PR #195, opened 2019-05-04
[Better Fire] Add new features

AndanteDevs:
Notable edits:
- ~~Crossbow piercing arrows have the `enchanted_hit` particle~~
- Flame arrows set campfires alight
- Added a sound for when flame arrows set a block to fire

### DIFF
--- a/gm4_better_fire/data/better_fire/functions/checks.mcfunction
+++ b/gm4_better_fire/data/better_fire/functions/checks.mcfunction
@@ -1,10 +1,9 @@
 #check block at flame arrow
-#@s - @e[nbt=!{Fire:-1s},type=arrow]
+#@s - @e[type=minecraft:arrow,nbt=!{Fire:-1s}]
 #called by better_fire:main
 
 #set fire block if air
-execute if data entity @s inBlockState if block ~ ~ ~ #better_fire:air run playsound minecraft:block.fire.extinguish block @a[distance=..20] ~ ~ ~ 100 2
-setblock ~ ~ ~ fire keep
+execute if data entity @s inBlockState if block ~ ~ ~ #better_fire:air run function better_fire:set_fire
 
 #campfire lighting
 execute if block ~ ~ ~ minecraft:campfire[waterlogged=false,lit=false] run function better_fire:set_campfire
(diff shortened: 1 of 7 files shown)


gm4_better_fire/data/better_fire/functions/checks.mcfunction
```diff
@@ -0,0 +1,10 @@
+#check block at flame arrow
+#@s - @e[nbt=!{Fire:-1s},type=arrow]
+#called by better_fire:main
+
+#set fire block if air
+execute if data entity @s inBlockState if block ~ ~ ~ #better_fire:air run playsound minecraft:block.fire.extinguish block @a[distance=..20] ~ ~ ~ 100 2
```
misode: You're using this check to play the sound, but instead of only doing that, put the setblock commands from below also under this check, using a function.
But because this is then the only command in this function, you should rather move this command downwards into `main`.

gm4_better_fire/data/better_fire/functions/main.mcfunction
```diff
@@ -1,6 +1,6 @@
 #fire arrows place fire
-execute at @e[nbt=!{Fire:-1s},type=arrow] run setblock ~ ~ ~ fire keep
+execute as @e[nbt=!{Fire:-1s},type=minecraft:arrow] at @s run function better_fire:checks
```
misode: Put type before nbt for efficiency and move the check from `better_fire:checks` into this command. Probably rename the checks function
```suggestion
execute as @e[type=minecraft:arrow,nbt=!{Fire:-1s},nbt={inBlockState:{}}] at @s if block ~ ~ ~ #better_fire:air run function better_fire:in_block
```

gm4_better_fire/data/better_fire/functions/pulse_check.mcfunction
```diff
@@ -1,2 +1,5 @@
 #compares assigned run tick to current tick and calls main
 execute if score current_tick gm4_clock_tick = better_fire gm4_clock_tick run function better_fire:main
+
+#arrow piercing enchanted hit particle
+execute at @e[type=minecraft:arrow,nbt=!{PierceLevel:0b}] run particle minecraft:enchanted_hit ~ ~ ~ 0 0 0 0 1 normal @a
```
misode: Is this really necessary? This is doing nbt serializations for every arrow, every tick, simply for some particles.

AndanteDevs: How about tag an entity once and run the particle from that tag? If I put the tag first in the selector, it won't check for NBT.

misode:
This has become a feature in 1.14.1-pre2 making this PR obsolete.

Bluefire-610:
Closing this PR since the feature was added by mojang in 1.14.1

PR #196, opened 2019-05-04
Saddle item motion

Dennis-0:
When breaking the chair, the saddle item no longer spawns without a motion (a fox is spawned holding the item and is killed)


sparks0:
Dennis can you push this to the same PR as your other chairs change? Otherwise whichever one we merge first will conflict with this one.

Kroppeb:
Would they conflict though?

PR #198, opened 2019-05-04
Update Wormhole Bottles for 1.14

alluet:
A change by Mojang in 1.14 resulted in Wormhole Bottles no longer
working if their destination lay in unloaded chunks. This change has
been accommodated for by replacing the `teleport_to_dimension` function
with `set_position` and `set_dimension`, which use the marker area
effect cloud as `@s` instead of the player.

Computational loot tables are now used to check the player's mainhand
and offhand slots for Wormhole Bottles, this avoids the cost of
serializing every player's NBT twice every 16 ticks (which is extremely
expensive).

Due to 1.14 adding the ability to copy NBT between entities, player NBT
is only serialized a single time per Wormhole Bottle in their hands. A
temporary item entity is used to minimize the cost of copying the
stored coordinates into the scoreboard. Once Gamemode 4 provides a
forceloaded chunk, this optimization can be made even better by using a
Jukebox rather than a temporary item entity.

Due to 1.14 adding advanced NBT paths, it is now possible to efficiently
retrieve the coordinates of the offhand Wormhole Bottle. Therefore, this
commit has enabled the ability for the player to consume Wormhole
Bottles held in their offhand and get teleported to the intended
destination.

Proper offhand support has also fixed an exploit which allowed the
player to consume a Wormhole Bottle in their offhand, but in fact get
teleported to the location of a Wormhole Bottle in their mainhand (which
they did not actually consume). Now, they will be correctly teleported
to the location stored in their offhand Wormhole Bottle.


misode:
> A change by Mojang in 1.14 resulted in Wormhole Bottles no longer
working if their destination lay in unloaded chunks

@alluet Is this change still needed in 1.14.1 pre-2?

alluet:
I'll need to test. I doubt they changed anything relating to it. It's necessary for various other reasons too, I'll elaborate when I have more time later.

alluet:
The "bug" that caused me to reimplement the Wormhole Bottles' teleportation code has still not been fixed. Therefore it is highly unlikely that the current 1.13 module could work.

Also, my new implementation is more optimized than the previous one, as it minimizes non-`@s` selectors except when absolutely required, so I don't see a reason to reference it.

gm4_zauber_cauldrons/data/zauber_cauldrons/functions/cauldron/wormhole_targeting/choose_dimension.mcfunction
```diff
@@ -1,19 +1,35 @@
...
+#copy the correct scores into fake players wormhole_x, wormhole_y, wormhole_z, wormhole_d
+execute if score offhand_warp gm4_zc_data matches 0 run function zauber_cauldrons:cauldron/wormhole_targeting/use_mainhand
+execute if score offhand_warp gm4_zc_data matches 1 run function zauber_cauldrons:cauldron/wormhole_targeting/use_offhand
+
+#affect player with resistance
+execute unless data entity @s ActiveEffects[{Id:11b,Ambient:0b}] run effect give @s resistance 1 12 true
+
 #particles and sound for depart
-particle minecraft:portal ~ ~.6 ~ .25 .25 .25 0 100
-playsound minecraft:entity.enderman.teleport player @a[distance=1..8] ~ ~ ~ 1 .3
```
Bluefire-610: > No need to remove this minecraft:. It's in there in case mojang ever decides that we have to put it there, it does no harm having it.

gm4_zauber_cauldrons/data/zauber_cauldrons/functions/cauldron/wormhole_targeting/arrival_animation.mcfunction
```diff
@@ -1,11 +1,11 @@
...
 #particles and sound
-particle minecraft:portal ~ ~.6 ~ .25 .25 .25 0 100
-playsound minecraft:entity.enderman.teleport player @a[distance=..8] ~ ~ ~ 1 .3
+particle portal ~ ~.6 ~ .25 .25 .25 0 100
+playsound entity.enderman.teleport player @a[distance=..8] ~ ~ ~ 1 .3
 
 #explosion if no cauldron in target block
 execute align xyz unless entity @e[type=area_effect_cloud,tag=gm4_zauber_cauldron,dx=0,dy=0,dz=0] run summon tnt
-execute align xyz unless entity @e[type=area_effect_cloud,tag=gm4_zauber_cauldron,dx=0,dy=0,dz=0] run playsound minecraft:entity.generic.explode block @s
+execute align xyz unless entity @e[type=area_effect_cloud,tag=gm4_zauber_cauldron,dx=0,dy=0,dz=0] run playsound entity.generic.explode block @s
```
Bluefire-610: Something seems to be off with this arrival animation: 

1. Valid Cauldron at dest. , unloaded:
- **Current Behavior:** Enderman teleport sound, explode sound (no TNT).
- **Expected Behavior:** Enderman arrival particle and teleport sound, no explode sound.

2. No Cauldron at dest. , unloaded: 
- **Current Behavior:** Enderman teleport sound, explode sound, but no TNT is summoned.
- **Expected Behavior:** Enderman teleport sound, explode sound, primed TNT entity.


This might be correlated to the missing arrival particle unless one warps to a cauldron within particle render distance - as in these checks might run at the wrong `at @s` or  simply the wrong time.

alluet:
Mhm, I did notice the weird behavior with the cauldrons, but as I was working in 1.14, there were some very strange bugs. I'll probably reorganize the code when I have some time in an attempt to fix the bugs.

Bluefire-610:
I'm testing in 1.14.2, just for the record

alluet:
I figured you were--I was just pointing out that I was making this for 1.14 and some weird Minecraft bugs made it impossible to fix the odd arrival behavior. I'm certain it can be fixed now.

alluet:
I believe I have fixed the issues with this data pack:

 - The the arrival animation is now in `set_dimension`, rather than a dedicated `arrival_animation` function. This fixes the issues with TNT not spawning for unloaded cauldrons.

 - The teleport particles now display as expected; this was only possible by making them `force`, but they are limited to `@a[distance=..32]` to simulate the `normal` particle setting as much as possible.

 - The explosion `playsound` was removed to avoid duplicate sounds, mainly because keeping it would involve an extra `@a` (due to moving the arrival animation to `set_dimension`, which is run by an area effect cloud). My testing suggested there was no point in playing the extra explosion sound; the TNT itself made an explosion sound.

 - I also forcibly load the chunk using `forceload add` and `forceload remove` (only if the chunk is not already forceloaded). This is not necessary to fix the issues the previous iteration had in vanilla, but I included them in my commit so I can pull the changes to my Paper testing server to confirm if they are needed for Paper.

Bluefire-610:
Tested this in game, but found a weird bug.

1. Get wormhole to an invalid location that's in the dimension you're in (this does not occur when switching dimensions!) but that's **not** chunkloaded.

2. Consume the wormhole

3. Expected: An explosion should occur.
    Actual: No explosion, just a safe teleport.

PR #199, opened 2019-05-04
Update Trapped Signs to 1.14

### LINKED ISSUE #96: Trapped sign loses its text on servers or after relogging

misode:
* Added all sign variants
* Added support for translations and custom models
* Fixes #96 (the same was as #178 did)
* Uses an advancement instead of an nbt check and statistic to detect sign placement.


Bluefire-610:
Just tested this in game and it seems to work!

A couple of things I noticed:
-When you don't write any text on the sign it doesn't become trapped - as in it won't trigger redstone - you still get the drops upon destruction so it's not gamebreaking but still weird.
-When you destroy the sign during a redstone pulse the redstone dust stays lit until it's updated by another block

misode:
> When you don't write any text on the sign it doesn't become trapped - as in it won't trigger redstone - you still get the drops upon destruction so it's not gamebreaking but still weird.

This is a side effect of the method used to detect when you're done editing the sign. I found no other way to detect that.

> When you destroy the sign during a redstone pulse the redstone dust stays lit until it's updated by another block

I should be able to fix that really easily

PR #204, opened 2019-05-06
Update base: Check for updates whenever the player is targetable

misode:
The base will wait to check updates and out-of-date messages when the server starts until it can find a player.

The only side-effect that this will cause in case there are modules with different bases installed is that `gm4:update` will be called twice on each reload.


misode:
> Why did you write your new lines using `@a[limit=1]` (which makes sense) but CHANGE the existing `@a[limit=1]` to `@p`?

Oh, I think that must be a mistake while resolving a previous conflict. I will change that back.

PR #210, opened 2019-05-09
Add Better Jukeboxes

misode:
Original module for 1.13 was shelved: #28 

The code has been updated to copy the whole nbt data of the music disc to not interfere with another module. Additionally, extracting and inserting of music discs is now separated by 8 ticks to avoid bugs when using /playsound and /stopsound in the same tick.


sparks0:
Tested in game. Works great. custom nbt on discs is preserved correctly! Small issue is the jukebox can only pull a disc from the first slot of the top hopper.

Some files should be more carefully commented indicating the @s and parent function.

Happy to merge.

sparks0:
small bug if the disc is inserted with a hopper but is removed by right-click, the sound continues to play. Not a big issue though.

Bloo-dev:
There's nothing wrong with this PR but it feels weird that this does not work for all Jukeboxes, so here's one proposed change:
- When a player places a hopper, raycast to the hopper then look at the block below.
- If there is a jukebox below, place the marker AEC in it.

misode:
@Bloo-dev Doing a raycast for every hopper placed seems like it could cause lag very quickly. It would also not catch it if you place the jukebox afterwards.

sparks0:
> @Bloo-dev Doing a raycast for every hopper placed seems like it could cause lag very quickly. It would also not catch it if you place the jukebox afterwards.

a) raycasting can be triggered using an advancement, no? Which means no clocking needed. As for placing jukebox second, also raycasting to placed jukeboxes to look for hoppers above would be a solution.

misode:
The raycasting itself can cause lag. Creating lag spikes every time someone places a hopper. 

I don't know of any GM4 modules that do raytracing and marking blocks behind your back. Usually there's always a recipe or creation process. This module changes the way jukeboxes behave and that should be clear to the player.

If throwing a redstone block on a jukebox it too random, it could also be implemented as a custom crafter recipe. Just like blast furnaces, which replace the dropper with a hopper.

SandSander:
I don't think it will be updated anymore?
 is also old version?

Bloo-dev:
I'm closing this as it was made in an outdated version of Minecraft.

misode:
I'm incorporating this into the Boxes module.

PR #211, opened 2019-05-10
[Submission] Midnight Menaces 1.13

BluePsychoRanger:
This module adds 6 expansion packs to mysterious midnights. The expansion packs add a variety of mob events to mysterious midnights, including Slime Fests, an expansion pack present during GM4D, but not added as one of the default 6 expansions packaged with the Mysterious Midnights module.
Documentation: https://docs.google.com/document/d/1fYJ9w3ZnwnhxDvCigHM7bKOOiPQBJ2CBMc8foVuR8RU/edit?usp=sharing


Bluefire-610:
Did you thoroughly test the illusioners? Last time I messed with them (1.12) they were really glutchy and not really a nice enemy from a survival standpoint.
During their illusion attack - when they split up - none of the copies actually had a hitbox, insteaf you had to hit the geometric centre of all copies (which might be inside blocks).

BluePsychoRanger:
> During their illusion attack - when they split up - none of the copies actually had a hitbox, insteaf you had to hit the geometric centre of all copies (which might be inside blocks).

During the illusion attack, all four of the copies are illusions. The real illusioner is in the middle, yes, but never inside the blocks. It actually turns invisible and can be indicated by the particles that are given off when a mob has the invisibility effect. The illusions can be inside blocks, as they have no hitbox, but the real mob is invisible hence why you can't see the hitbox with F3+B

AndanteDevs:
Illusioners actually run their illusion based on whether or not they are invisible (invisible, illusion; visible, normal)

BluePsychoRanger:
The changes to #220 have been made

gm4_midnight_menaces/data/sandy_husks/loot_tables/husk.json
```diff
@@ -0,0 +1,53 @@
...
+        {
+          "type": "item",
+          "name": "minecraft:rotten_flesh",
+          "weight": 1,
+          "functions": [
+            {
+              "function": "set_count",
+              "count": {
+                "min": 0,
+                "max": 2,
```
BluePsychoRanger:
```suggestion
                "max": 2
```

SandSander:

**This is the new version:**
https://github.com/Gamemode4Dev/GM4_Datapacks/pull/396

PR #217, opened 2019-05-10
Made `Mending Tanks`

Kroppeb:
Putting an item with the `Mending` enchantment repairs into a XP tank will repair it.
Will use `#damage / 2 (rounded up)` XP to do so with al limit of 16 XP / module tick

### DIFF
--- a/gm4_mending_tanks/data/gm4/tags/functions/init_check.json
+++ b/gm4_mending_tanks/data/gm4/tags/functions/init_check.json
@@ -1,6 +1,6 @@
 {
     "values":[
-        "liquid_tanks:init_check"
+        "mending_tanks:init_check"
     ]
 
 }
--- a/gm4_mending_tanks/data/gm4/tags/functions/init_check.json
+++ b/gm4_mending_tanks/data/gm4/tags/functions/init_check.json
@@ -1,0 +1,6 @@
+{
+    "values":[
+        "liquid_tanks:init_check"
+    ]
+
+}
--- a/gm4_mending_tanks/data/mending_tanks/functions/init.mcfunction
+++ b/gm4_mending_tanks/data/mending_tanks/functions/init.mcfunction
@@ -1,0 +1,21 @@
+#announce module installation
+tellraw @a[gamemode=creative] ["",{"text":"[GM4]: Installing Mending Tanks..."}]
+execute unless entity @p run say GM4: Installing Mending Tanks...
+
+#declare and initialise scoreboards and settings
+scoreboard players set update_happened gm4_upCheck 1
+scoreboard players set mending_tanks gm4_modules 1
+scoreboard players set mending_tanks gm4_clock_tick 0
+
+# make sure this exists
+scoreboard objectives add gm4_lt_util dummy
+scoreboard players set #mt_repair_per_xp gm4_lt_util 2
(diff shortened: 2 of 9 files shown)


gm4_mending_tanks/data/gm4/tags/functions/init_check.json
```diff
@@ -0,0 +1,6 @@
+{
+    "values":[
+        "liquid_tanks:init_check"
+    ]
```
misode: Why does this datapack contain function tags that call liquid tanks? These are already part of the liquid tanks base module. Same with `pulse_check.json`.

gm4_mending_tanks/data/mending_tanks/functions/item_fill.mcfunction
```diff
@@ -0,0 +1,6 @@
+#@s = liquid tank with item in first slot
+#run from liquid_tanks:item_process
+
+
+#experience tank
+execute if score @s[tag=gm4_lt_experience] gm4_lt_value matches 1.. if block ~ ~ ~ hopper{CustomName:"{\"text\":\"Experience Tank\"}",Items:[{Slot:0b,tag:{Enchantments:[{id:"minecraft:mending"}]}}]} run function mending_tanks:check_item
```
misode: This is called from liquid tanks so it shouldn't need to check for the name of the block. That will cause problems when an update changes it (eg. using translation keys).

gm4_mending_tanks/data/mending_tanks/functions/init.mcfunction
```diff
@@ -0,0 +1,21 @@
+#announce module installation
+tellraw @a[gamemode=creative] ["",{"text":"[GM4]: Installing Mending Tanks..."}]
```
misode: Use translation keys for these tellraws. See the [template pack](https://github.com/Gamemode4Dev/GM4_Datapacks/blob/master/gm4_template_pack/data/template/functions/init.mcfunction) as an example. Also an issue on line 16

gm4_mending_tanks/data/mending_tanks/functions/init.mcfunction
```diff
@@ -0,0 +1,21 @@
...
+execute unless entity @p run say GM4: Installing Mending Tanks...
```
misode: Use `@a[limit=1]`. Also an issue on line 17.

gm4_mending_tanks/data/gm4/tags/functions/init_check.json
Kroppeb: Should be `mending_tanks:...` I'll fix those

gm4_mending_tanks/data/mending_tanks/functions/item_fill.mcfunction
Kroppeb: Tell that to special. If he removes those checks, I will do so too.

gm4_mending_tanks/data/mending_tanks/functions/init.mcfunction
Kroppeb: I just copied liquid tanks I'll fix this.

sparks0:
Not gonna be accepting new modules for 1.13. Please submit things for the 1.14 branch. You can always make a PR for a 1.13 version afterwards if you want back support

Issue #215, opened 2019-05-10
[LT:Standard Liquids] Util functions use slow selectors to reselect the liquid tank


Kroppeb:
None of the selectors in the util functions use a limited range when reselecting the liquid tank. 
This means that every entity in the whole world will be checked.

### Possible Solution
2 solutions

#### Distance Check
Adding a distance means that the selector only goes through the armorstands in the chunks that are covered in the distance check

#### Fake player
The selectors are almost all used to set the buffer value. If instead the buffer would be stored on a fake player, no selector is needed.

SpecialBuilder32:
Would adding a distance check to the selector not increase the required processing operations and thus be less efficient?

Kroppeb:
> Would adding a distance check to the selector not increase the required processing operations and thus be less efficient?

No, atm it will go through all entities in the world. This means that checking the type is actually harming efficiency as the only entity with this tag will be an `armor_stand` anyway.
Adding a distance or cuboid area, the game will instead iterate over the chunks in that area and ask them their list of the entities of that type (this is stored)

See https://minecraftcommands.github.io/commanders-handbook/selector-argument-order

alluet:
Special, assuming a distance check works on Marker armor stands (and last I knew, it did not), it would indeed be an optimization to make use of one. Despite taking more operations for each entity in the entity list, the entity list itself is much smaller. This is because adding a distance check means that only the chunks covered by the distance are used to generate the entity list, rather than the whole save.

Kroppeb:
> Special, assuming a distance check works on Marker armor stands (and last I knew, it did not)

running `/say @e[distance=..5]` selects the liquid_tank

PR #221, opened 2019-05-12
Add Liquid Minecarts

sparks0:
Liquid Minecarts first draft. I threw this together in a few days so I'm sure there are errors or places where performance can be improved.

Video: https://www.youtube.com/watch?v=qQctcD3WvWw

Liquid Minecarts are used to move liquids between tanks. This module needs feedback and testing. The goal in its design is to support any new liquids automatically. Instead of hard-coded values, liquid minecarts copy the scores, tags and textures from the tank they load from.

Minecarts can hold 40% of the capacity of the tank from which they got their liquid from (usually 120 bottles).
Minecarts are created by placing down a hopper minecart and entering the same recipe as a liquid tank.
Minecarts will fill if there is a liquid tank pointed at them and they are empty or the liquid in the tank matches their own.
Minecarts will empty their contents into an empty tank or a tank with matching liquid if they are on top of the tank.
Minecarts will empty completely if they drive over a powered activator rail. This is so automated systems can assure tanks have no residual contents.
Minecarts have no GUI for survival players and have no functionality outside of being able to interface with tanks. Think of them as pipes.

A note on why liquid minecarts are command_block_minecart entities. Originally I used regular minecarts but this caused the passenger tank armour stand to be ejected when going over an activator rail. Minecarts themselves don't support skulls as a custom display block. I didn't want the minecarts to have a GUI because they have no GUI functionality and seeing a hopper interface on a mobile liquid tank will cause players to expect it to be able to bottle and unbottle. Furthermore a GUI will cause the tank to suck items out of hoppers or place them in, potentially messing up auto-filler systems if it picks something up while travelling down a track. Furnace minecarts drive weird as we all know so command block minecart it is! There's an added advantage that the "drain" feature is entered as a command in the minecart itself meaning there's no 20hz clock needed to check for activator rails. As far as I can tell, in survival it acts like a normal minecart. You can't view the command GUI and breaking it doesn't drop a command block.


gm4_liquid_minecarts/data/liquid_minecarts/functions/compare_load_liquid.mcfunction
```diff
@@ -0,0 +1,8 @@
+#@s = liquid minecart containing a liquid with a liquid tank containing liquid pointed at it
+#run from try_to_load
+
+summon area_effect_cloud ~ ~ ~ {Tags:["gm4_liquid_minecart_compare"],Duration:1}
+data modify entity @e[type=area_effect_cloud,limit=1,sort=nearest,tag=gm4_liquid_minecart_compare] CustomName set from block ~ ~ ~ CustomName
```
SpecialBuilder32: I worry that this won't be a safe way to compare what type of liquid the tank when things are moved to translation keys, though I'm not certain. Thoughts?

gm4_liquid_minecarts/data/liquid_minecarts/functions/init_liquid.mcfunction
```diff
@@ -0,0 +1,19 @@
+#@s = empty liquid minecart next to a liquid tank pointed at it containing liquid
+#run from liquid_minecarts:try_to_unload
+
+data modify entity @s CustomName set from block ~ ~ ~ CustomName
+data modify entity @e[type=armor_stand,tag=gm4_liquid_minecart_display,limit=1,sort=nearest] HandItems[0] set from entity @e[type=armor_stand,tag=gm4_liquid_tank_display,limit=1,sort=nearest] ArmorItems[3]
```
SpecialBuilder32: What is the point of having the liquid texture copied to the hand item if you cannot see it? If its not intended to be seen, maybe moving it to the chest slot would avoid forcing the player to render the skull.

gm4_liquid_minecarts/data/liquid_minecarts/functions/init_liquid.mcfunction
```diff
@@ -0,0 +1,19 @@
+#@s = empty liquid minecart next to a liquid tank pointed at it containing liquid
+#run from liquid_minecarts:try_to_unload
+
+data modify entity @s CustomName set from block ~ ~ ~ CustomName
+data modify entity @e[type=armor_stand,tag=gm4_liquid_minecart_display,limit=1,sort=nearest] HandItems[0] set from entity @e[type=armor_stand,tag=gm4_liquid_tank_display,limit=1,sort=nearest] ArmorItems[3]
+data modify entity @s Tags set from entity @e[type=armor_stand,tag=gm4_liquid_tank,limit=1,sort=nearest] Tags
+tag @s remove gm4_lt_idle
```
SpecialBuilder32:
```suggestion
```
This tag no longer exists in the Liquid Tanks base. Since functions run per tank, tanks that would be "idle" due to no items inside them simply don't run the processing functions

gm4_liquid_minecarts/data/liquid_minecarts/functions/init_tank.mcfunction
```diff
@@ -0,0 +1,22 @@
+#@s = liquid minecart
+#run from try_to_unload
+
+data modify block ~ ~ ~ CustomName set from entity @s CustomName
+summon armor_stand ~ ~-.45 ~ {CustomName:"\"gm4_liquid_tank_display\"",Tags:["gm4_no_edit","gm4_liquid_tank_display","gm4_lm_needs_texture"],NoGravity:1,Marker:1,Invisible:1,Invulnerable:1,Small:1,DisabledSlots:2039552,Fire:20000}
+data modify entity @e[type=armor_stand,tag=gm4_lm_needs_texture,limit=1,sort=nearest] ArmorItems[3] set from entity @e[type=armor_stand,tag=gm4_liquid_minecart_display,limit=1,sort=nearest] HandItems[0]
+tag @e[type=armor_stand] remove gm4_lm_needs_texture
```
SpecialBuilder32: Why is there a tag for wether the tag needs to be applied if it's removed right after? If its not needed, it would save on an `@e`

gm4_liquid_minecarts/data/liquid_minecarts/functions/load_liquid.mcfunction
```diff
@@ -0,0 +1,17 @@
...
+#get remaining space in minecart
+scoreboard players operation @s gm4_lm_data = @s gm4_lt_max
+scoreboard players operation @s gm4_lm_data -= @s gm4_lt_value
+#lower lm_data if tank has less liquid than the remaining space
+scoreboard players operation @s gm4_lm_data < @e[type=armor_stand,limit=1,sort=nearest,tag=gm4_liquid_tank] gm4_lt_value
+
+#add liquid to minecart
+scoreboard players operation @s gm4_lt_value += @s gm4_lm_data
+#take liquid from tank
+scoreboard players operation @e[type=armor_stand,limit=1,sort=nearest,tag=gm4_liquid_tank] gm4_lt_buffer -= @s gm4_lm_data
```
SpecialBuilder32: Since this module isn't controlled by the liquid tanks base, it should force a liquid update on the liquid tank it interacts with. The update applies the buffer and updates display information. 
`liquid_tanks:liquid_value_update`

gm4_liquid_minecarts/data/liquid_minecarts/functions/minecart_to_tank_init.mcfunction
```diff
@@ -0,0 +1,8 @@
+#@s = liquid tank that a minecart is putting a new liquid into
+#run from init_tank
+
+tag @s add gm4_liquid_tank
+tag @s add gm4_lt_idle
```
SpecialBuilder32:
```suggestion
```

gm4_liquid_minecarts/data/liquid_minecarts/functions/pulse_check.mcfunction
```diff
@@ -0,0 +1,3 @@
+#compares assigned run tick to current tick and calls main
+function #liquid_minecarts:tick
```
SpecialBuilder32: Is this tag used anymore? I couldn't find any use of it but I might be blind

gm4_liquid_minecarts/data/liquid_minecarts/functions/init_liquid.mcfunction
```diff
@@ -0,0 +1,19 @@
...
+data modify entity @e[type=armor_stand,tag=gm4_liquid_minecart_display,limit=1,sort=nearest] HandItems[0] set from entity @e[type=armor_stand,tag=gm4_liquid_tank_display,limit=1,sort=nearest] ArmorItems[3]
+data modify entity @s Tags set from entity @e[type=armor_stand,tag=gm4_liquid_tank,limit=1,sort=nearest] Tags
+tag @s remove gm4_lt_idle
+tag @s remove gm4_lt_drain
+tag @s remove gm4_lt_fill
+tag @s remove gm4_liquid_tank
+tag @s add gm4_liquid_minecart
+
+#set tank max based on 40% of hopper max
+scoreboard players operation @s gm4_lt_max = @e[type=armor_stand,tag=gm4_liquid_tank,limit=1,sort=nearest] gm4_lt_max
```
SpecialBuilder32: @Kroppeb you said in #215 that adding distance arguments to selectors helps reduce the number of chunks that the game scans through when resolving a selector. Would they also be beneficial here?

gm4_liquid_minecarts/data/liquid_minecarts/functions/unload_liquid.mcfunction
```diff
@@ -0,0 +1,19 @@
...
+
+#remove liquid from minecart
+scoreboard players operation @s gm4_lt_value -= @s gm4_lm_data
+#give liquid to tank
+scoreboard players operation @e[type=armor_stand,limit=1,sort=nearest,tag=gm4_liquid_tank] gm4_lt_buffer += @s gm4_lm_data
+#update display
+function liquid_minecarts:liquid_value_update
+playsound minecraft:block.brewing_stand.brew block @a ~ ~ ~ 1.0 1.5
+#clear liquid tags
+execute if score @s gm4_lt_value matches 0 run data merge entity @s {Tags:["gm4_liquid_minecart","gm4_liquid_minecart_empty"]}
```
SpecialBuilder32: Should the name of the minecart also be cleared here, or is it okay to overwrite it the next time the cart receives liquid?

gm4_liquid_minecarts/data/liquid_minecarts/functions/init_liquid.mcfunction
Kroppeb: Yes it would.

gm4_liquid_minecarts/data/liquid_minecarts/functions/compare_load_liquid.mcfunction
sparks0: I don't know, we'd have to test I guess. Tehnically the tank should copy the translation string and compare that, right?

There isn't really any other way to make this module work with ANY liquid without modifying liquid tanks to support this one.

gm4_liquid_minecarts/data/liquid_minecarts/functions/init_liquid.mcfunction
sparks0: I guess there was no reason except that I could see it worked. I can move it to the chest slot!

gm4_liquid_minecarts/data/liquid_minecarts/functions/init_liquid.mcfunction
sparks0: When you say "no longer", is that a change you made since I created this PR? Cos tanks definitely had this tag while I was testing.

gm4_liquid_minecarts/data/liquid_minecarts/functions/load_liquid.mcfunction
sparks0: Ah, I was trying to work out how to do that and found that setting the buffer caused the tank to update. Does running the function simply cause the update to happen immediately instead of when LT gets around to it?

gm4_liquid_minecarts/data/liquid_minecarts/functions/unload_liquid.mcfunction
sparks0: I guess now would make sense. The only thing it'll affect though is the name of the cart if you break it.

gm4_liquid_minecarts/data/liquid_minecarts/functions/compare_load_liquid.mcfunction
Bluefire-610: This brings up a more general question:
Should we scrap concepts / introduce extra checks just so an optional language pack works?

sparks0: If GM4 was chinese and I wanted to use it and some of the modules had english translations... I'd rather have chinese liquid Minecarts I can work out how to use than no liquid minecarts at all.

Kroppeb: Oh, wait. This os a glaring bug. In 1.14 we were planning to remove any tank name checks from Liquid Tanks as a user can quickly swap them with a different one, which now just locks some of the functionality. 

The check has to be done on the tags so maybe when you copy the data, you should store the tags list into an nbt tag.

Bluefire-610: This seems to be an issue according to @Kroppeb. I don't understand how liquid tanks works but is it fixable? Should we try to fix it?

gm4_liquid_minecarts/data/liquid_minecarts/functions/load_check.mcfunction
```diff
@@ -0,0 +1,12 @@
+#@s = liquid minecart
+#run from main
+
+#load
+execute if block ~1 ~ ~ hopper[facing=west] positioned ~1 ~ ~ run function liquid_minecarts:try_to_load
+execute if block ~-1 ~ ~ hopper[facing=east] positioned ~-1 ~ ~ run function liquid_minecarts:try_to_load
+execute if block ~ ~ ~1 hopper[facing=north] positioned ~-1 ~ ~ run function liquid_minecarts:try_to_load
```
Kroppeb:
```suggestion
execute if block ~ ~ ~1 hopper[facing=north] positioned ~ ~ ~1 run function liquid_minecarts:try_to_load
```

gm4_liquid_minecarts/data/liquid_minecarts/functions/load_check.mcfunction
```diff
@@ -0,0 +1,12 @@
+#@s = liquid minecart
+#run from main
+
+#load
+execute if block ~1 ~ ~ hopper[facing=west] positioned ~1 ~ ~ run function liquid_minecarts:try_to_load
+execute if block ~-1 ~ ~ hopper[facing=east] positioned ~-1 ~ ~ run function liquid_minecarts:try_to_load
+execute if block ~ ~ ~1 hopper[facing=north] positioned ~-1 ~ ~ run function liquid_minecarts:try_to_load
+execute if block ~ ~ ~-1 hopper[facing=south] positioned ~-1 ~ ~ run function liquid_minecarts:try_to_load
```
Kroppeb:
```suggestion
execute if block ~ ~ ~-1 hopper[facing=south] positioned ~ ~ ~-1 run function liquid_minecarts:try_to_load
```

gm4_liquid_minecarts/data/liquid_minecarts/functions/compare_load_liquid.mcfunction
Kroppeb: I was able to confirm this bug actually works.
Here are the steps:
## How to convert liquid types:
* Have a liquid tank of the desired type `desired` and one with the liquid you want to use `current`
* Move liquid from `desired` to a liquid minecart `LM` but make sure there is still some liquid left in `desired`
* Break `desired` (and replace with a different hopper if you want to keep the liquid)
* Break `current` and quickly replace with the hopper you got from breaking `desired`
* Move `LM` below `current`

gm4_liquid_minecarts/data/liquid_minecarts/functions/init_liquid.mcfunction
SpecialBuilder32: I just looked into it. Turns out tanks were given this tag when created, but the tag was never used anywhere in the module. I've removed that tag from the 1.14 module now.

gm4_liquid_minecarts/data/liquid_minecarts/functions/load_liquid.mcfunction
SpecialBuilder32: Yes. On its own, the base will run an update if the buffer is not 0. The buffer is where changes in liquid go before being applied and is used for the "smart" interface.

gm4_liquid_minecarts/data/liquid_minecarts/functions/compare_load_liquid.mcfunction
SpecialBuilder32: Does breaking a hopper with json- CustomName really transfer its name to the item properly? Or is this only applicable to tanks without translation keys?

Kroppeb: Before executing the bug, I fetched this branch and your liquid tanks branch and merged them and tested in that state. I did forcefully make the ml take the liquid out due to a bug that now *should* be fixed. Even if that caused it and breaking json tanks doesn’t work, one can still reset the name before transferring the liquid to and from tanks

sparks0: So.... this is fixed because the bug you say *should* be fixed?

Kroppeb: > So.... this is fixed because the bug you say _should_ be fixed?

No, bug I'm referencing was the issue that minecarts weren't filled from south and north IIRC

Kroppeb: I do have a solution in mind, will make a pull for that later maybe.

PR #229, opened 2019-05-19
Fix bug where a mould could consume 2 ores at once

misode:
This happened a few times while testing and adding in this `limit=1` check fixes it.


Bluefire-610:
Manually merged this into a new metallurgy build on the test server, thanks for the commit!

PR #230, opened 2019-05-19
Make ores in animation unstackable

misode:
These small changes make sure that the item is unstackable in the casting animation.

The current implementation makes ores stack when there is a lag spike or when the clock is misaligned. This new implementation is required if you want to reliably drop ores every clock cycle.


Bluefire-610:
Manually merged this into a new metallurgy build on the test server, thanks for the commit!

PR #232, opened 2019-05-21
Fix issues with metallurgy casting

misode:
This pull request fixes a few issues that were found while trying to make an automated caster.

* Obsidian casts (Shamirs!) could jump and be absorbed by a mould
* Two ore items could jump on the same mould in one clock pulse
* Only one obsidian item globally was being checked to create a mould
* A mould could be formed with the items not directly on the sand, but for example on a slab

I tested these changes thoroughly to make sure that no items could be deleted or duplicated.

### DIFF
--- a/gm4_metallurgy/data/metallurgy/functions/casting/check_mould_creation.mcfunction
+++ b/gm4_metallurgy/data/metallurgy/functions/casting/check_mould_creation.mcfunction
@@ -1,0 +1,3 @@
+# @s = obsidian block with count of 1 on top of sand
+# run from main
+execute align xyz unless entity @e[type=vex,tag=gm4_sand_ring,dx=0,dy=-1,dz=0] if entity @e[type=item,dx=0,dy=0,dz=0,nbt={Item:{id:"minecraft:clay_ball",Count:1b},OnGround:1b}] run function metallurgy:casting/create_mould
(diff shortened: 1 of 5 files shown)


misode:
@Bluefire-610 Can you take a look at this please?

gm4_metallurgy/data/metallurgy/functions/casting/check_mould_creation.mcfunction
```diff
@@ -0,0 +1,3 @@
+# @s = obsidian block with count of 1 on top of sand
+# run from main
+execute align xyz unless entity @e[type=vex,tag=gm4_sand_ring,dx=0,dy=-1,dz=0] if entity @e[type=item,dx=0,dy=0,dz=0,nbt={Item:{id:"minecraft:clay_ball",Count:1b},OnGround:1b}] run function metallurgy:casting/create_mould
```
Bluefire-610: I don't see why this is it's own function and not appended to the execute in `metallurgy/functions/main.mcfunction` but that's fine with me.

misode: execute makes it list of entities before calling the first function, so if this execute was chained in `main`, it could remove two obsidian items for one clay.

Bluefire-610: Alright!

PR #241, opened 2019-06-12
Use namespaced %1$s

misode:
Use a namespaced `%1$s` to prevent conflicts with other resource packs using this language trick.

This number is calculated using the following formula, which gives `3427655` as result.
```python
ord("G") + 256*ord("M") + 256*256*ord("4")
```

This requires changes in the GM4_Resources repo: https://github.com/Gamemode4Dev/GM4_Resources/pull/10


Kroppeb:
So the value spells `4mg`?

alluet:
The bytes of the number (a 31 bit unsigned integer) are `['G', 'M', '4', 0]` on most modern architectures (i.e. little endian architectures). I don't really want to argue about which endianness to use, but note that if you do it in the reverse order, the size of the number would be much bigger unless you add padding.

Kroppeb:
Oh, I never thought about the fact that the number had to be a valid int32

misode:
Yep this is ready to be merged, some PR's waiting to be merged will now need to be modified.

PR #242, opened 2019-06-18
Add Holographic Tags

BluePsychoRanger:
Recently, SethBling tweeted an idea of creating floating text using name tags and item frames. https://twitter.com/SethBling/status/1140675391012528128/video/1
I've basically replicated the features he showcased, but modified it to be more lag friendly (running on the GM4 16 gt clock cycle), and added particles, sounds, and an advancement, plus a feature to color the message if the item frame is placed on wool before placing the name tag.
Documentation: 
https://docs.google.com/document/d/1M_tTZc2tkksFDvv4pWa3e8cbaI_vWBKLbXHPzRWpngw/edit?usp=sharing


AndanteDevs:
> its not possible to color the name using json formatting in this context is it?

The `CustomName` tag is a string with JSON inside of it, not a pure JSON text component - you could use the append function in `data modify`, but for that reason you can't

gm4_holographic_tags/data/holographic_tags/functions/init.mcfunction
```diff
@@ -0,0 +1,49 @@
+#announce module installation
+tellraw @a[gamemode=creative] [{"translate":"%1$s","with":["[GM4]: ",{"translate":"text.gm4.prefix"}]},{"translate":"%1$s","with":["Installing Holographic Tags...",{"translate":"text.gm4.modules.update.installing","with":[{"translate":"module.gm4.holographic_tags"}]}]}]
+execute unless entity @a[limit=1] run say GM4: Installing Holographic Tags...
+
+#declare and initialise scoreboards and settings
+scoreboard players set update_happened gm4_up_check 1
+scoreboard players set holographic_tags gm4_modules 1
+scoreboard players set holographic_tags gm4_clock_tick 0
+
+team add gm4_holo_white
```
AndanteDevs: Why is this `gm4_holo_white` and not `gm4_ht_white`, like the rest?

gm4_holographic_tags/data/holographic_tags/functions/main.mcfunction
```diff
@@ -0,0 +1,3 @@
+execute as @e[type=area_effect_cloud,tag=gm4_hologram] at @s positioned ^ ^ ^-0.3 positioned ~ ~.3 ~ if entity @e[type=item_frame,nbt=!{Item:{}},distance=..0.1,limit=1] run function holographic_tags:destroy_hologram
+execute as @e[type=item_frame,tag=!gm4_hologram] if data entity @s {Item:{id:"minecraft:name_tag",tag:{display:{}}}} at @s positioned ^ ^ ^0.3 positioned ~ ~-.3 ~ unless entity @e[type=area_effect_cloud,tag=gm4_hologram,distance=..0.1] run function holographic_tags:create_hologram
+tag @e[type=item_frame,nbt=!{Item:{id:"minecraft:name_tag"}}] remove gm4_hologram
```
misode: Check for the `gm4_hologram` tag before the nbt to exclude most item frames from the nbt check
```suggestion
tag @e[type=item_frame,tag=gm4_hologram,nbt=!{Item:{id:"minecraft:name_tag"}}] remove gm4_hologram
```

gm4_holographic_tags/data/holographic_tags/functions/create_hologram.mcfunction
```diff
@@ -0,0 +1,12 @@
+# @s = item frame with renamed name tag
+# run from main
+
+summon area_effect_cloud ~ ~ ~ {CustomNameVisible:1,Particle:"block air",Age:-2147483647,Duration:2147483647,Tags:["gm4_hologram"]}
+data modify entity @e[type=area_effect_cloud,tag=gm4_hologram,limit=1,sort=nearest,distance=..0.1] CustomName set from entity @s Item.tag.display.Name
+data modify entity @e[type=area_effect_cloud,tag=gm4_hologram,limit=1,sort=nearest,distance=..0.1] Rotation set from entity @s Rotation
```
misode: Why are you copying rotation onto the area effect cloud? Afaict you aren't storing that back onto the item when destroying the hologram so I don't see the point.

gm4_holographic_tags/pack.mcmeta
```diff
@@ -0,0 +1,22 @@
+{
+  "pack":{
+    "pack_format":0,
+    "description":"A Gamemode 4 Module"
+  },
+  "module_name":"Holographic Tags",
+  "module_id":"holographic_tags",
+  "site_description":"Setup Floating Messages with a simple name tag!",
```
sparks0: should be "Set Up"

AndanteDevs: The capitalisation seems a bit off as well, might I add

misode:
Oh, and replace all `%1$s` with `%1$s%3427655$s`

gm4_holographic_tags/data/holographic_tags/functions/create_hologram.mcfunction
BluePsychoRanger: It's used to detect which item frame to check when putting the name tag back into an item frame. Unlike Sethbling's there can be multiple holograms in a single block, and removing them requires the item frame to be placed on the same block face that the original item frame was placed on to properly be detected.

misode: Ah, thanks for the clarification 👌

gm4_holographic_tags/data/holographic_tags/functions/init.mcfunction
BluePsychoRanger: It was a typo; gm4_holo was the original prefix, but some names were too long, like light blue, so I changed it to ht, but I guess I forgot to change the white one. I didn't notice cuz white is the default, so it didn't change anything while I was testing.

BluePsychoRanger:
Okay, there's a slight bug (or we could just call it a feature ¯\_(ツ)_/¯) with using AECs. They can activate wooden, iron, and golden pressure plates, and will keep them activated, rendering the pressure plate useless.

BluePsychoRanger:
> Okay that's weird. Definitely seems like an MC issue, though in my testing the AEC couldn't press the plate down, but could prevent it from unpressing. Happens with clouds from lingering potions too.

Yeah, that's what I was experiencing too. Since there's not really any way around it (without using armor stands), let's just call it a hidden feature :P

PR #243, opened 2019-06-22
Liquid tanks removal of fill and drain

Kroppeb:
This builds on #237

Comes with following features:
* Removes the need of adding the fill and drain tag to the tank. 
* Allows a drain operation immediately after a full operation unless that would undo the first. (and vica versa)

### DIFF
--- a/gm4_liquid_tanks/data/liquid_tanks/functions/init.mcfunction
+++ b/gm4_liquid_tanks/data/liquid_tanks/functions/init.mcfunction
@@ -7,12 +7,15 @@ scoreboard players set update_happened gm4_upCheck 1
 scoreboard players set liquid_tanks gm4_modules 1
 scoreboard players set liquid_tanks gm4_clock_tick 0
 
+# add scoreboards
 scoreboard objectives add gm4_lt_value dummy
 scoreboard objectives add gm4_lt_max dummy
-scoreboard objectives add gm4_lt_buffer dummy
-#scoreboard objectives add gm4_lt_disp_max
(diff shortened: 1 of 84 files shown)


gm4_liquid_tanks/data/liquid_tanks/functions/init.mcfunction
```diff
@@ -4,16 +4,19 @@ execute unless entity @p run say GM4: Installing Liquid Tanks...
 
 #declare and initialise scoreboards and settings
 scoreboard players set update_happened gm4_upCheck 1
-scoreboard players set liquid_tanks gm4_modules 1
+scoreboard players set liquid_tanks gm4_modules 2
```
misode: Don't manually increment this score

gm4_liquid_tanks/data/liquid_tanks/functions/init_check.mcfunction
```diff
@@ -1,3 +1,3 @@
 #unless the module is already initialized
-execute unless score liquid_tanks gm4_modules matches 1.. run function liquid_tanks:init
+execute unless score liquid_tanks gm4_modules matches 2.. run function liquid_tanks:init
```
misode: Same here

gm4_liquid_tanks/data/liquid_tanks/functions/init.mcfunction
Kroppeb: Oh, I did not know the site does this. This is an issue in #237.

PR #244, opened 2019-06-25
Copy all spawner data to minecart rather than a hardcoded list

misode:
This allows you to convert any spawner into a spawner minecart. And it makes sure to copy all the necessary tags to the entity.


Bluefire-610:
This removes the performance boost that spawner minecarts get vs normal spawners.

Besides that, I think that a whitelist-like system is a better approach, this is to be discussed, though.

misode:
> This removes the performance boost that spawner minecarts get vs normal spawners.

Oh I forgot about that, I'll recheck the additional tags.

In the 1.13 version spawner minecarts had this bug where the tag wasn't copied that made Orbis spawners not drop XP. This change would avoid any problems like this in the future

Bluefire-610:
Looking good. Like I said, I'm not sure whether we should support _all_ spawners by default, I'll talk with Special and Sparks about this

sparks0:
I think we need a whitelist for this module. I don't want Orbis spawners or other custom spawners used by people's town systems or custom bosses to be moved.

PR #245, opened 2019-06-30
Add Potion Infusers

BluePsychoRanger:
Allows Potion Tanks to feed into a potion infuser that will infuse certain blocks surrounding it and convert them into different blocks, mobs, or special infused crops. Original idea by Tesset.

Documentation:
https://docs.google.com/document/d/1hVApTMaLTWsnSjN4IG7uvC5hcOt4lbLBpjDDlO_K5no/edit?usp=sharing


sparks0:
This is actually a really cool idea! I love most of these conversion options and they sound really cool!

I worry that installing this module would make it difficult to store your potions in your base without inadvertantly applying the effects so while I appreciate it would require quite a rework, I feel this is better off either as another block that takes potions from an attached tank or a tank upgrade. Idk whether we should go back to "items in other slots set tank modes" like we did in 1.8 because we kinda got rid of all that but that might also be an option.

sparks0:
Sames on the second one.

BluePsychoRanger:
Alright, it's been changed to a specific block. The doc is updated too with details on crafting and such

misode:
I really like this module! Although there are definitely concerns that things could be exploited. Especially with the mob recipes. Here are some of my concerns.

This modules makes sponges renewable. I'm not against this since I feel this will result in some interesting farms, but just making sure that everyone is ok with this.

ghasts drop quartz, which can be used to respawn a ghast. Make sure that you don't get more out of it than with what you started.

Poppies turning into wither roses seems like a really cheap and safe way to get them.

A slime with size 2 is spawned, which is larger than a full block. I think it would make more sense to spawn it with size 1. It would also eliminate the chance of dropping more than 9 slime balls.

Endermite drops chorus plant. This block has a lot of quirks when placed and it has no use. I don't see why it should be a drop. Alternatively it could drop a chorus fruit?

Why use a different bat loot table? The thing I'm most skeptical about is that the same custom item is now referenced in two modules.

Infused food felt really weird at first. Are they basically stackable potions?

BluePsychoRanger:
> Poppies turning into wither roses seems like a really cheap and safe way to get them.
Yeah, that one is probably too overpowered, especially since poppies are extremely cheap

> ghasts drop quartz, which can be used to respawn a ghast. Make sure that you don't get more out of it than with what you started.
> A slime with size 2 is spawned, which is larger than a full block. I think it would make more sense to spawn it with size 1. It would also eliminate the chance of dropping more than 9 slime balls.
Actually, the idea of these two is that players are able to gain a net profit of slime/quartz at the expense of the potion. I believe it could introduce some interesting farms.

> Endermite drops chorus plant. This block has a lot of quirks when placed and it has no use. I don't see why it should be a drop. Alternatively it could drop a chorus fruit?
Yeah, I think I'll change it to chorus fruit 

> Why use a different bat loot table? The thing I'm most skeptical about is that the same custom item is now referenced in two modules.
I removed the gunpowder from the bat loot table. I didn't think it made sense for leaves to be used to farm gunpowder...

> Infused food felt really weird at first. Are they basically stackable potions?
Yes, the idea is that you infuse a strong potion into a crop to gain a weaker version, but it's stackable

sparks0:
I feel like we should trail this on the server.

SandSander:

New Version: https://github.com/Gamemode4Dev/GM4_Datapacks/pull/370

PR #246, opened 2019-07-01
Update Orbis to 1.14

misode:
- Use a loot table to detect the biome
- Allow multiple biome features per chunk
- Don't fill ocean caves with water
- Rebalance and buff the loot in structures
- Improve debug messages
- Simplify expansions
- Improve clock system


Bluefire-610:
Is this the version that's currently doing the work? @misode

misode:
> Is this the version that's currently doing the work? 

Yes

PR #248, opened 2019-07-05
Tinkering Compressors

Sheepe:
A module developed in Gamemode 4 Labs, created by myself and the idea created by Tesset.

**Description:** Allows for automated tinkering of items as an extension of block compressors.

**Dependencies:** 
- Metallurgy
- Block Compressors (and hence Custom Crafters)

**How to use:**

Start by creating a block compressor. Throw a Tinker shamir on top of the block compressor, and use a piston to "smoosh" the shamir into the compressor. Setup shown below: 
image
Upon creation, it will look like this:
image

You can then place anywhere from 2-64 items into the tinkering compressor to set it to **tinkering** mode. It will then perform the same function as tinker would on a tool, by compressing anything that is thrown onto it as tinker would. It will only tinker it *once*, and must be picked up by a player or hopper before any tinkering compressor can do anything to it. Placing 1 item into the tinkering compressor will set it into **detinkering** mode, and the reverse operation will be completed.

Upon breaking the compressor, you will recieve all of the items that it took to create the tinkering compressor, (everything that you get from a block compressor, plus the shamir).


sparks0:
Love this idea! Does a tinkering compressor still function as a compressor or does it now only tinker?

Sheepe:
It only tinkers and untinkers

Bluefire-610:
This is some neat cross-module interaction! Just tested this in game and it works as intended.

Bluefire-610:
Reminder @sparks0 @SpecialBuilder32 - did we check the code for this one?

PR #249, opened 2019-07-05
Noisy Blocks

MightyBurger:
A small datapack developed for the Gamemode 4 module suite, works with the latest game version (1.14.4).

**Description:** Interactive blocks, such as crafting tables, play a noise when interacted with by a player. 

Some blocks in the game, such as chests and barrels, already have interaction noises. This module adds interaction noises for fourteen additional blocks:

noisy_blocks_thumb
- Crafting table
- Furnace
- Beacon
- Loom
- Brewing Stand
- Smoker
- Blast Furnace
- Cartographer's Table
- Stonecutter
- Dropper
- Dispenser
- Hopper
- Lectern (with book)
- Campfire (placing food)

There are a couple important points I want to mention.
- This module works based on scoreboard objectives for interacting with blocks. I found no such scoreboard objective for the grindstone or the enchanting table. The Fletching and Smithing tables, being unfinished and unusable blocks, also do not have corresponding scoreboard objectives and may need to be added in once they can be interacted with.
- This module runs multiple commands every tick (20hz), which may have a small impact on server performance. This problem could be solved by calling `main.mcfunction` one time the moment a player opens any GUI - however, I was unable to find a way to detect whether a player has a GUI open. Even so, the module is small enough any impact on performance should be minimal. 
- This is my first time using GitHub, so I probably messed something up. If so, please let me know.


MightyBurger:
I've moved the commands over and changed the lectern noise. The lectern feels much more natural now, thank you. Would anyone have suggestions for changing other sounds?

BluePsychoRanger:
I wonder if it's better for performance to just do `playsound @a[scores={gm4_nbk_crafting=1..}]` rather than `execute as @a[] at @s`. But IDK, maybe @misode knows?

misode:
> I wonder if it's better for performance to just do `playsound @a[scores={gm4_nbk_crafting=1..}]` rather than `execute as @a[] at @s`

This would indeed be more efficient, but then the sound doesn't play at the player's location. @BluePsychoRanger

BluePsychoRanger:
> This would indeed be more efficient, but then the sound doesn't play at the player's location. @BluePsychoRanger

Oh right, what about using only `execute at @a[]` instead of `execute as @a[] at @s`? Because it doesn't seem that there needs to be a reason to execute as the player if the sound is just being played at the location of the player

misode:
Currently the sound is only played for the player interacting with the block, so `as` is required here.

BluePsychoRanger:
Oh right

sparks0:
@MightyBurger I like this module idea and am happy to approve it once the suggested changes by Misode/BPR are applied!

MightyBurger:
I've applied BPR's suggestion that should hopefully improve performance, with the side effect that nearby players will be able to hear the block interaction noises as well.

gm4_noisy_blocks/data/noisy_blocks/functions/pulse_check.mcfunction
```diff
@@ -0,0 +1,29 @@
+#This function runs every tick. It plays a sound for anyone with an interaction score of 1 and immediately resets their score.
+execute at @a[scores={gm4_nbk_crafting=1..}] run playsound minecraft:entity.shulker.close block @a[distance=..8] ~ ~ ~ 0.5 1.4
+scoreboard players reset @a[scores={gm4_nbk_crafting=1..}] gm4_nbk_crafting
+execute at @a[scores={gm4_nbk_furnace=1..}] run playsound minecraft:block.stone.place block @a[distance=..8] ~ ~ ~ 0.5 0.6
+scoreboard players reset @a[scores={gm4_nbk_furnace=1..}] gm4_nbk_furnace
+execute at @a[scores={gm4_nbk_beacon=1..}] run playsound minecraft:block.stone.place block @a[distance=..8] ~ ~ ~ 0.5 0.5
+scoreboard players reset @a[scores={gm4_nbk_beacon=1..}] gm4_nbk_beacon
```
sparks0: Do you think for all of these it would be more performant to have a function per sound and reset the player's score in the function so we're halving the player tests?

sparks0:
It's very subtle but I like it. The one that feels the most needed is placing items on the campfire.

gm4_noisy_blocks/data/noisy_blocks/functions/pulse_check.mcfunction
```diff
@@ -0,0 +1,29 @@
+#This function runs every tick. It plays a sound for anyone with an interaction score of 1 and immediately resets their score.
+execute at @a[scores={gm4_nbk_crafting=1..}] run playsound minecraft:entity.shulker.close block @a[distance=..8] ~ ~ ~ 0.5 1.4
+scoreboard players reset @a[scores={gm4_nbk_crafting=1..}] gm4_nbk_crafting
```
sparks0: Do you think for all of these it would be more performant to have a function per sound and reset the player's score in the function so we're halving the player tests?

sparks0:
So the module is definitely cool. I just tested it in game and it works fine but Bloo just pointed out that we don't need to run everything at 20hz from pulse_check. You can use advancements to trigger the sounds without needing a clock so we think moving this to advancement triggers is the way to go.

I'm going to close this PR, you can comment to ask it to be re-opened when you have a version that uses advancement triggers to push :)

MightyBurger:
I've looked through the list of advancement triggers and cannot find one that would be useful here. I looked for one that triggers when a player interacts with a block, one that triggers when a scoreboard updates, and one that triggers when a statistic updates, but none of those seem to exist. Nothing on the list seems to be useful for this application. I haven't played with advancements very much, so this is unfamiliar territory for me and I could be missing something. Any ideas?

MightyBurger:
Sparks, I've updated the datapack and pushed it to Github, but it does not seem to be updating on this pull request. I'm not sure whether it's because the PR is closed or because I'm a confused and incompetent Github user :) I would appreciate if you looked at the changes and consider re-opening the PR. Thank you! 

Here is a description of the changes I've made.

> I've made some changes. Bloo's suggestion seems perfect and I would love to implement it, but I cannot find an advancement trigger that I could use for this application. 
> Instead, I've halfed the player tests done each tick by creating many individual functions per block and running a function when a player interacts with a particular block. 
> 1.15 added scoreboard objectives for anvil and grindstone interactions, so I was able to add sounds for those. That, however, may mean that this datapack will no longer work with versions 1.14.4 and below.

misode:
You need to create a new PR. You can use this link: https://github.com/Gamemode4Dev/GM4_Datapacks/compare/master...MightyBurger:noisy-blocks

PR #252, opened 2019-07-10
Add Magical Musical Jukeboxes

BluePsychoRanger:
Allows a piston to push a musical shamir cast into a jukebox, which will allow it to give effects to nearby entities while a music disc is playing.

https://docs.google.com/document/d/1xSYyCTEXQxpdIphvgPgySOy9pHIxiPXxhkFbgf8XHvw/edit?usp=sharing


sparks0:
I don't think I like this module... No offence to the coding. Where does the potion effect come from? Minecraft doesn't really have an association between music and potions. It should probably also be a radius instead of a square to match the range at which the music can be heard. I think combining this with the automatic jukeboxes would just allow people to create cheap, free beacons at no cost after setup.

Bloo-dev:
This seems very op and most of these effects do not correlate to musical. We think it'd be better if it only gave the effects musical provides, so:
- strength
- regeneration
- speed
- slowness

How about a tiered effect? Having Musical Tier 2 in a 8 block radius and Musical Tier 1 in a 9 to 16 blocks donut?

PR #257, opened 2019-07-11
Add Private Lockers

BluePsychoRanger:
Adds a way to lock containers

https://docs.google.com/document/d/1pXnAD8dYIIxYY-hCF7Tg7PkbvxOY7tQiWZby277Kzmw/edit?usp=sharing


Bluefire-610:
I like the idea of this.
Back in 1.8 I tried to create somthing very similar based on a scoreboard ID system, but it had too many flaws.
Some of those were:
- it is still possible to break the container
- making it unbreakable is a bad idea

Personally, I don't mind if these locked containers can still be broken. It's more about feeling secure rather than being secure.

One small thing I'd prefer is having the player craft the lock onto the container. Throwing stuff onto things just feels outdated to me but Sparks, and possible Special, would disagree.

BluePsychoRanger:
> One small thing I'd prefer is having the player craft the lock onto the container. Throwing stuff onto things just feels outdated to me but Sparks, and possible Special, would disagree.

Do you mean something similar to the piston pushing the item into the chest or right clicking the chest with the lock? Both are possible but the latter requires a player score check every tick.

Bluefire-610:
> > One small thing I'd prefer is having the player craft the lock onto the container. Throwing stuff onto things just feels outdated to me but Sparks, and possible Special, would disagree.
> 
> Do you mean something similar to the piston pushing the item into the chest or right clicking the chest with the lock? Both are possible but the latter requires a player score check every tick.

Neither, I'd suggest renaming the lock item in an anvil and then crafting that lock with the container of choice in a custom crafter

BluePsychoRanger:
Oh right, so I mentioned this on discord when I showed it off, but you can't go from name component to Lock component because Name uses JSON while Lock only uses text strings. This means that a lock named "ABCDE" would turn into a locked container with the lock "{"text":"ABCDE"}"

Bluefire-610:
Oh I see. Well that's inconvenient.

BluePsychoRanger:
Yeah, I know. My initial idea was to make a simple module where you just name a lock and lock the chest. Then I had to get into book and quill cuz that's the only way I can think of to have a text component that's not JSON

sparks0:
What does the system do if the text in the book is over the character limit?

sparks0:
> Oh right, so I mentioned this on discord when I showed it off, but you can't go from name component to Lock component because Name uses JSON while Lock only uses text strings. This means that a lock named "ABCDE" would turn into a locked container with the lock "{"text":"ABCDE"}"

Does this still apply in 1.15 with the new storage command?

misode:
> Does this still apply in 1.15 with the new storage command?

Yes it still does. We would need string manipulation to convert `{"text":"ABCDE"}` into `ABCDE`

SandSander:
**This module idea is quite nice!**
but i think it is still in old version?

SandSander:

Bloo-dev:
I'm closing this as it was developed for an outdated version of Minecraft

PR #260, opened 2019-07-13
[Submission] Impulse Cores

AndanteDevs:
When wearing enchanted iron armour, and holding an 'Impulse Core' in your offhand slot, there's some kind of power running through the armour that gives the player very high effects of speed, jump boost and strength. To keep this balanced, every 'use' of this effect will remove charges of the core ~~(for example *every 10 blocks ran or every jump*) and the player will have a high level of weakness~~. It needs to be recharged after a couple uses by placing it on top of a beacon and throwing a nether star on top of it.

Custom Crafter Recipe:
recipe

Note that this module is in early beta and probably has balancing/code issues.

*Idea by @Dennis-0*


sparks0:
This feels mechanically very *similar* to the existing crystals from Zauber Cauldrons. I'm gonna vote no on this module's addition to GM4. We'll see what other admins think.

Bluefire-610:
I'm with sparks, it feels very similar to zauber's crystals

sparks0:
Based on the module's current functionality I'm closing this PR. Sorry!

PR #262, opened 2019-07-21
Added some data to the witch villager

Kroppeb:
Things this commit adds
+ The villager can now be traded with
+ The villager's trades are locked
+ The villager is now a cleric, this means to refill the trades one has to supply a brewing stand for him to use

I am looking into adding something like a zombie cured reputation to the villager. Any thoughts?


Kroppeb:
Also currently the `maxUses` defaults to `4`. Should we change this? Make this more random?

PR #263, opened 2019-07-21
Adds a zombiecured-like reputation and delete thrown gapple

Kroppeb:
This PR adds:
+ Gossip which will result in the thrower of the gapple to have the same discount as curing a zombie
    - This does mean that the curer only has to pay 1 emerald for each potion
+ Deletes the gapple on spigot now (I think)

Questions:
* What to do about the massive discount?
    * Reduce the price influence reputation has on the prices. This would also affect negative reputation
    * Only have the temporary spreadable `minor_positive`
    * Increase default prices
* MaxUses for all the trades are set to the default 4, should this be higher/random?


Bluefire-610:
I think the discount is fine, it's not like this is the quickest way to get potions anyways.

Kroppeb:
> Looking good. But won't the `Count:0` issue also affect golden carrots dùring Pigmen->Pig conversion?

They all use the `consume_item` function, so it should be patched for all of them.

Bluefire-610:
They'd need the `gm4_gapple` tag, though - right?

Kroppeb:
```mcfunction
execute as @e[type=item,limit=1,nbt={Item:{id:"minecraft:golden_carrot"}},distance=..3,sort=nearest] run function mob_conversion:consume_item
```

No, I just did the removing of the tag there cause it was easier that way and shouldn't harm performance that much.

Kroppeb:
If you add `?w=1` to the end of the link it hides whitespace changes
https://github.com/Gamemode4Dev/GM4_Datapacks/pull/263/files?w=1

PR #236, opened 2019-06-06
Update Metallurgy to 1.14

misode:
Initial changes to make metallurgy work in 1.14. This does not use any new 1.14 features to improve efficiency yet.

* use `#minecraft:signs` in conduction block tag
* all lore changed to use text components
* tnt exploding within 16 ticks of each other now works correctly
* use `Age:0s,PickupDelay:10s` instead of `Age:1s,PickupDelay:9s` (I'm not sure why)
* uncomment the repair function call

PR #267, opened 2019-08-02
Bats now no longer explode when they are defused

suppergerrie2:
Added a check for the defused tag before the explode function is called so now bats do not explode when they are defused. ( Needs #236 to be able to defuse bats )


gm4_bat_grenades/data/bat_grenades/functions/main.mcfunction
```diff
@@ -1,2 +1,2 @@
 execute at @e[type=bat] run playsound minecraft:entity.bat.ambient hostile @a[gamemode=!creative,gamemode=!spectator,distance=..7]
-execute as @e[type=bat] at @s if entity @a[gamemode=!creative,gamemode=!spectator,distance=..3] unless entity @s[nbt={NoAI:1b}] run function bat_grenades:explode
+execute as @e[type=bat] at @s if entity @a[gamemode=!creative,gamemode=!spectator,distance=..3] unless entity @s[nbt={NoAI:1b}] unless entity @s[tag=gm4_defused_bat] run function bat_grenades:explode
```
misode: You can combine the NoAI check and the defused check in one selector. Put the tag check in front as optimization.

suppergerrie2: I'll put the tag in front, when I combined the check it had to be both NoAI and defused (so it would be NoAI && defused) instead of either NoAI or defused (NoAI || defused). But I may have messed up in my testing,

suppergerrie2:
I moved the tag check in front of the NoAI check @misode

gm4_bat_grenades/data/bat_grenades/functions/main.mcfunction
misode: Oh yeah you're right

BluePsychoRanger:
Would it not be more performant to do `/execute as @e[type=bat,tag=!gm4_defused_bat]` instead of `unless entity @s[tag=gm4_defused_bat]`

PR #276, opened 2019-08-11
Update Cement Mixers to 1.14

AndanteDevs:


gm4_cement_mixers/data/gm4/advancements/root.json
```diff
@@ -0,0 +1,20 @@
+{
+  "display":{
+    "icon":{
+      "item":"command_block",
+      "nbt":"{display:{Name:\"\\\"GM4 Logo\\\"\"}}"
```
misode: This is an outdated root advancement

PR #279, opened 2019-08-13
Add string recipes

Nik3141:
1 wool -> 3 string in the custom crafter, per request. I couldn't get the minecraft:wool tag to work, so they're in separate commands.


Nathan22211:
isn't 1 wool 4 string though? i feel like it should be 4 not 3

AndanteDevs:
We had a discussion about this in GM4 Labs and decided to make it 3 as a kind of cost

sparks0:
Have tested in game. It only works with white wool. Is this an acceptable behavior? On one hand people might have other wool they want to make into string, on the other it would make you lose the dye...

sparks0:
> All of these commands should be moved into a function, like stair recipes:
> `execute if score @s gm4_slot_count matches 1 if score @s gm4_stack_size matches ..21 run function standard_crafting:string_recipes`
> 
> Inside that function all the recipes should be checked with:
> `execute if block ~ ~ ~ dropper{Items:......`

Oh yeahhh good catch.

PR #280, opened 2019-08-13
Roped arrows

Nik3141:
Please note that although the branch is still named roped_arrows, everything inside is called crossbow_cables.

The aim of the crossbow_cables module is to provide an efficient method for creating ziprail track. To use the module, create a spooled crossbow with a crossbow and a fishing rod like so:  
image  
(thanks to adriaan1313 for the picture)
Then, place at least one piece of string in the offhand and fire the crossbow in one of the cardinal directions. A line of string will be placed in a line along that axis, and the string will be removed from the player's inventory, up to a maximum of 40 pieces (which is the maximum length of a tripwire).

Edit: I have renamed all of the namespaces to efficient_track_laying, as well as the advancements and MODULE_ID and MODULE NAME placeholders, because I think it's a name that better reflects what the module does.

Edit 2: I have renamed everything to crossbow_cables because that was the highest voted name in the Labs discord poll.


sparks0:
Please provide a description detailing the aims of the module, steps to use the module and any crafting recipes that are used.

adriaan1313:
The usage of a spooled crossbow consumes string in gamemode creative and crafting seems to be broken, but that could just be me.

image

Nik3141:
It wasn't supposed to work in creative, but I'll fix it so it does when I get home, as well as write up a description. That's really weird that the crafting doesn't work, as it definitely should work. It didn't work in my first commit but only in my second, and I wasn't sure that the second went through or not, so that may be the problem. I'll look into it when I get home.

Nik3141:
adriaan1313 The crafting works fine for me, so I think you may have the original version, which did not properly implement any crafting. Make sure you have the latest commit I just sent.

adriaan1313:
I see, I cloned your datapack from the master branch which is not the roped_arrows branch.
I'm sorry for bothering you.

Nik3141:
No no problem you were right the crafting was broken in creative.

gm4_crossbow_cables/data/crossbow_cables/functions/recipe_check.mcfunction
```diff
@@ -0,0 +1,13 @@
+#spooled crossbow
+
+#neither have been used
+execute if score @s gm4_slot_count matches 2 if score @s gm4_stack_size matches ..1 if block ~ ~ ~ dropper{Items:[{Slot:0b,id:"minecraft:crossbow"},{Slot:4b,id:"minecraft:fishing_rod"}]} unless data block ~ ~ ~ Items[0].tag.Damage unless data block ~ ~ ~ Items[4].tag.Damage run data merge block ~ ~ ~ {Items:[{Slot:8,id:"minecraft:crossbow",Count:1b,tag:{gm4_spooled_crossbow:1b,gm4_custom_crafters:{multiplier:1},display:{Lore:["{\"text\":\"Use with string in the offhand\",\"color\":\"dark_aqua\",\"italic\":\"false\"}","{\"text\":\"to create lines of tripwire.\",\"color\":\"dark_aqua\",\"italic\":\"false\"}"],Name:"{\"text\":\"Spooled Crossbow\",\"italic\":\"false\"}"}}}]}
```
sparks0: These recipe checks should check to make sure the crossbow being used in the recipe is not already a spooled crossbow. I can currently do this.

sparks0:
I'm an idiot. I missed the note about holding the string in the offhand. It works fine now and consumes the correct amount of string!

My other comments are worth discussing though maybe.

gm4_crossbow_cables/data/crossbow_cables/functions/shoot_arrow.mcfunction
```diff
@@ -8,5 +8,5 @@ advancement revoke @s only crossbow_cables:shot_crossbow
 advancement grant @s only gm4:almost_a_grappling_hook
 
 #gm4_rot_y -20..20 means looking relatively near to horizon
```
misode: You forgot to change the comment as well

sparks0:
@Nik3141 Any news on this? The module has had changes requested since August. Our policy is to close PRs if the submitter doesn't respond to change requests for a long time so I'm closing this. If you make your changes you can message us to re-open this PR or you can make a new one.

PR #281, opened 2019-08-14
[Submission] Solid Ice

adriaan1313:
Adds item that allows for the creation of normal (clear) ice that does not melt from light.
([https://docs.google.com/document/d/1eePZIKxIjshL3UoS-P9LZAwNqbPGrRf3vjtMecxVU5Y/edit?usp=sharing](https://docs.google.com/document/d/1eePZIKxIjshL3UoS-P9LZAwNqbPGrRf3vjtMecxVU5Y/edit?usp=sharing))

(sorry for constantly checking the thing btw)


Kroppeb:
This can be made into a very quick ice farm.

AndanteDevs:
> This can be made into a very quick ice farm

I don't see how? It gets set to ice if it finds water, and destroys if else - unless I've overlooked something

Kroppeb:
> > This can be made into a very quick ice farm
> 
> I don't see how? It gets set to ice if it finds water, and destroys if else - unless I've overlooked something

No, it replaces the ice every tick but only checks once every 16 ticks to break the thing. It wont be that hard to make something that is gonna replace the water fast enough.

Even if they both happen each tick, I'm pretty sure you can 0-tick the ice out and get new water into the spot in the same tick.

Bluefire-610:
What would a use case for this be? I'm concerned about lag with that 20Hz AEC lookup.

Kroppeb:
If a player decides to place down the skull, it becomes useless once broken.

adriaan1313:
> What would a use case for this be? I'm concerned about lag with that 20Hz AEC lookup.

The use of this is purely decoration, so you probably won't accept it, but why not try.

Bluefire-610:
> > What would a use case for this be? I'm concerned about lag with that 20Hz AEC lookup.
> 
> The use of this is purely decoration, so you probably won't accept it, but why not try.

Deco modules are neat and we have plenty of them! I'm just concerned by one AEC per permanent ice.

adriaan1313:
> If a player decides to place down the skull, it becomes useless once broken.

I'll look into this and how heart canisters does it.

adriaan1313:
> > > What would a use case for this be? I'm concerned about lag with that 20Hz AEC lookup.
> > 
> > 
> > The use of this is purely decoration, so you probably won't accept it, but why not try.
> 
> Deco modules are neat and we have plenty of them! I'm just concerned by one AEC per permanent ice.

True, I'm just not smart enough to like do something smart to reduce the amount of AEC's apart from like making it a radius, so you probably won't accept it but why not try

adriaan1313:
> If a player decides to place down the skull, it becomes useless once broken.

fixed that in commit [9628558](https://github.com/Gamemode4Dev/GM4_Datapacks/pull/281/commits/9628558b83354676c06df17fd42c7ec9e0f50863)... (not if picked up to fast but, Heart Canisters doesn't do that either)

adriaan1313:
> > > > What would a use case for this be? I'm concerned about lag with that 20Hz AEC lookup.
> > > 
> > > 
> > > The use of this is purely decoration, so you probably won't accept it, but why not try.
> > 
> > 
> > Deco modules are neat and we have plenty of them! I'm just concerned by one AEC per permanent ice.
> 
> True, I'm just not smart enough to like do something smart to reduce the amount of AEC's apart from like making it a radius, so you probably won't accept it but why not try

in addition to this comment: 
This (square) radius thing could be done with either something like tiers or some other way of configuring the `gm4_solid_ice` tag or something like this.

sparks0:
I can't see this really working with the tools we have for managing blocks. I feel like players using this will want to do things like build ice palaces which will require a LOT of these entities which will cause them lag and as people above have said, it could potentially be abused to make hyper speed ice farms.

PR #291, opened 2019-08-29
Update destroy_liquid_tank.json

BluePsychoRanger:
Add glass to the loot table. Currently, the glass isn't dropped when breaking a liquid tank.


Kroppeb:
> I mean, sure. But I think this might have been intentional. Summoning @SpecialBuilder32

If so I would like it to drop when using a silktouch picaxe, but checking that isn’t that easy.

PR #292, opened 2019-08-29
Add translation namespaces everywhere

AndanteDevs:
Translation namespaces were added recently; however, quite a few instances were missed. This pull fixes this.


misode:
Thanks for spotting these! 👍

PR #296, opened 2019-09-04
Small patch for vertical minecarts to make use of 1.14 features

kruthers:
Added a new entity_type tag called minecart in vertical minecarts to replace the old lines responsible for taging every minecart


gm4_vertical_rails/data/vertical_rails/tags/entity_types/minecart.json
```diff
@@ -0,0 +1,10 @@
+{
+    "values":[
+        "minecraft:minecart",
+        "minecraft:chest_minecart",
+        "minecraft:hopper_minecart",
+        "minecraft:furnace_minecart",
+        "minecraft:spawner_minecart",
+        "minecraft:tnt_minecart"
```
sparks0: This list should now include command_block_minecart since liquid minecarts are that type :)

kruthers: I though i added that one, i have done it now

PR #301, opened 2019-09-09
Fixed disabled slots and added support for sloped rail loading (liquid Minecarts)

kruthers:
I fixed the disabled slots NBT (updated values from bloo's patch and moved it from the minecart NBT to armour stand NBT in the create function.

The load check commands were modified to support them being aligned to allow for support for sloped rails when the cart is being loaded from the North, South, East or West directions.

### DIFF
--- a/gm4_liquid_minecarts/data/liquid_minecarts/functions/init_tank.mcfunction
+++ b/gm4_liquid_minecarts/data/liquid_minecarts/functions/init_tank.mcfunction
@@ -2,7 +2,7 @@
 #run from try_to_unload
 
 data modify block ~ ~ ~ CustomName set from entity @s CustomName
-summon armor_stand ~ ~-.45 ~ {CustomName:"\"gm4_liquid_tank_display\"",Tags:["gm4_no_edit","gm4_liquid_tank_display","gm4_lm_needs_texture"],NoGravity:1,Marker:1,Invisible:1,Invulnerable:1,Small:1,DisabledSlots:2039583,Fire:20000}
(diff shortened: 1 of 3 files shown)


gm4_liquid_minecarts/data/liquid_minecarts/functions/init_tank.mcfunction
```diff
@@ -2,7 +2,7 @@
 #run from try_to_unload
 
 data modify block ~ ~ ~ CustomName set from entity @s CustomName
-summon armor_stand ~ ~-.45 ~ {CustomName:"\"gm4_liquid_tank_display\"",Tags:["gm4_no_edit","gm4_liquid_tank_display","gm4_lm_needs_texture"],NoGravity:1,Marker:1,Invisible:1,Invulnerable:1,Small:1,DisabledSlots:2039583,Fire:20000}
+summon armor_stand ~ ~-.45 ~ {CustomName:"\"gm4_liquid_tank_display\"",Tags:["gm4_no_edit","gm4_liquid_tank_display","gm4_lm_needs_texture"],NoGravity:1,Marker:1,Invisible:1,Invulnerable:1,Small:1,DisabledSlots:4144959,Fire:20000}
```
Bluefire-610: Why doesn't the value `2039583` work here? I was pretty sure that `2039583` locks all slot interactions.

kruthers: I just used that value as that is the one i normal used, after testing with the other one that looks to work too, i can revert it to that one if you want?

Bluefire-610: It would be more consistent to revert, sry :/

Kroppeb: The wiki doesn't say, but the value @kruthers uses it might lock the offhand.

PR #307, opened 2019-10-07
Fix Malachite Lump skull repairing CMD

AndanteDevs:
Malachite look like Enderpuff ye


adriaan1313:
would it not be more efficient to check if the name is  `"Player Head"` before checking the nbt?

misode:
Checking the name of a dropped item depends on the language of the server, so is not reliable.

AndanteDevs:
It's an inconsistency fix, yes. I decided to change the repairing function because then people can fix old heads

PR #315, opened 2019-10-28
Crossbow Cartridges

Dennis-0:
Hold certain items in your offhand to shoot them with a crossbow.

- Bucket of fish (any type): Arrow is replaced by fish and bucket of fish is replaced by a water bucket (if survival mode) -> gives advancement
- Potions (splash/lingering, any type): Arrow is replaced by potion
-Torches: Places torch where the arrow hits a block
-Fire charge: shoots blaze fireball (places 1 fire block on aimed block)

Optional:
- Crops (seeds, carrots, potatoes): When the arrow lands on top of a dirt type block, farmland is placed here with this crop
- Anvils (decided to keep this one in the datapack, but commented out because the anvil appears late and it seems a little odd)

This could possibly be merged with the roped arrows datapack later.

The advancement is still open to changes

(Sorry for making another PR, there was a folder in it that wasn't supposed to be in it and the comments wtih requested changes were on the wrong files.)


Dennis-0:
Ignore the commits, I messed up some file uploads and tried to fix it.

Dennis-0:
Update since last PR: 
- Crops are now optional, just like anvils
- Fixed some item clearing/giving inconsistencies
- Torches and crops now take away the item instantly, which gets dropped on the ground if placing the torch/crop fails. 
- Tidied the files a bit with a line in each file where it is run from and more folders

Kroppeb:
Why did you make a new pr?

Dennis-0:
> Why did you make a new pr?

There was another folder "better_crossbows" in it which should've been deleted, and all comments were in this folder anyway, so thought it would be fine by making a new PR without this other folder. There were a lot of changes as well, with filenames and folders in the datapack so making a new PR seemed easier so I didnt have to manually delete them (in the browser)

sparks0:
I think in the stream we said that we didn't want crop planting if I remember correctly?

Dennis-0:
> I think in the stream we said that we didn't want crop planting if I remember correctly?

That was indeed said, I think the crops are commented out but they can be removed completely if that is prefered

Bloo-dev:
We do not want to ship modules with commented out features or unused files, so we'll merge this as soon as those are removed from this pr ❤️

Dennis-0:
> We do not want to ship modules with commented out features or unused files, so we'll merge this as soon as those are removed from this pr ❤️

Understandable, removed those features now.

sparks0:
Shooting an entity with a torch arrow loses you the torch and the arrow (and damages the entity)

Dennis-0:
> Shooting an entity with a torch arrow loses you the torch and the arrow (and damages the entity)

That's indeed a drawback that can happen now that it takes away the torch the moment the arrow is shot. I don't see an easy fix for this at the moment

PR #320, opened 2019-11-04
Chestpack

suppergerrie2:
When a shulker box is thrown on a chest the chest will be destroyed and the shulker box will get filled with the items from the chest.
This only works if the chest does not contain another shulker box and if the shulker box thrown onto it is empty.
It works for normal chests and trapped chests and all colors of shulker boxes.

If someone has a better name, please send it!


gm4_chestpack/pack.mcmeta
```diff
@@ -0,0 +1,24 @@
...
+  "site_description":"Shulker boxes can now take the items in a chest when dropped on one",
+  "site_categories":[],
+  "video_link":"YOUTUBE_VIDEO_URL",
+  "wiki_link":"WIKI_URL",
+  "required_modules":[
+  ],
+  "recommended_modules":[
+  ],
+  "credits":{
+    "comment":"All credit arrays are optional and more can be created as needed. They will be displayed on the site page under their headings.",
```
AndanteDevs:
```suggestion
```

gm4_chestpack/pack.mcmeta
```diff
@@ -0,0 +1,24 @@
...
+    "Creator":[
+      ["suppergerrie2","https://www.suppergerrie2.com"]
+    ],
+    "Updated By":[
+    ]
```
AndanteDevs: Can be omitted

gm4_chestpack/data/gm4/advancements/MODULE_ID.json
```diff
@@ -0,0 +1,29 @@
...
+      "color": "gray"
+    }
+  },
+  "parent": "gm4:root",
+  "criteria": {
+    "impossible": {
+      "trigger": "minecraft:impossible"
+    }
+  }
+}
```
adriaan1313: this should be gone or renamed/changed

sparks0:
Interesting! I think there should be some sort of fail sound/particle if there's an issue so players get feedback that there's a problem instead of thinking the module is broken? You could tag the shulker box item with a fail tag so it doesn't constantly check the chest it is on for an improved condition.

suppergerrie2:
Oh yeah good idea, will look into that.
I have been wanting to use it about twice a week depending on how much I play and how many random chests I find.

suppergerrie2:
Added a particle and a sound when it fails, I'm no designer tho and the sound and particle can probably be improved.

Bloo-dev:
Thanks for submitting this PR. Sadly, we don't see it as being useful enough to have it in our collection.
The potential use cases are really just:
a) cleaning up early game stuff that you left in your base for too long
b) raiding dungeons

a) really doesn't need a dedicated solution and b) sounds good, but since you can't use partially filled shulker boxes the problem of having to sort an inventory is just delayed.

PR #322, opened 2019-11-09
add better hoppers

adriaan1313:
This module adds the ability for hoppers to pick up arrows, spectral arrows and tridents.
If they are shot/thrown by players(non creative) and on top of a hopper with space, they go in.

That's about it I guess
(I did not add tipped arrows because I would have had to hardcode every type in)

note: idk if the hopper minecart pickup "radius" thing is correct.
also, it involves loads of nbt checking so meh
note: commits 06ce080..., 8f11cf3..., 3cef8ce..., d8ce397... were reverting a thing I accidentally did and me not understanding github desktop, sorry


sparks0:
Interesting concept but what need is it filling? How would players use this in game? Hoppers cannot currently pick up any entity. If this was added they would have to be able to catch all tipped arrows, flying anvils, gravel, sands and concrete powders and I'm not sure why?

This is a no from me but thanks for the submission :)

Bloo-dev:
I like this concept and I feel like this should be in vanilla but Arrows are very easy to get and this doesn't seem to justify the lag this module will cause

PR #324, opened 2019-11-14
Updated vertial_rails:minecarts to be gm4:minecarts

kruthers:
Originally vertical's used had the minecrats entity tag inside its namespace, this has been moved to the gm4 namespace to make it the same as zip rails and the upcoming trains module.

The module has been fully tested and works


misode:
Ziprails uses `gm4:minecarts` while this PR changes it to `gm4:minecart` (contradicting its title)

The plural form should be used.

kruthers:
> Ziprails uses `gm4:minecarts` while this PR changes it to `gm4:minecart` (contradicting its title)
> 
> The plural form should be used.

@misode thanks for spotting that fixed it now

PR #329, opened 2019-12-11
Rename module namespaces

misode:
Adds `gm4_` to all module namespaces.
Commits from `dev/load` are included because this branches was started before #312 was squashed.


sparks0:
I think with a change this size there's not really much you can do in a review. I'll merge it and then I suppose we have to test them all including the obscure features of modules to check everything still works.

PR #346, opened 2019-12-14
Ver/1.15

sparks0:
Hid untested/broken 1.15 datapacks from the website until they can be tested.


misode:
@sparks0 This will also stop them from being tested from github. Is this what we want?

PR #358, opened 2019-12-30
fix ability to get free protection 1 book (honey beetroot)

adriaan1313:
You could get a "free" protection 1 book by throwing a honey beetroot on an enchantment extractor (the beetroot still works), this was caused by the enchantment level being 1 instead of 0.


adriaan1313:
> I'm pretty sure you can also do `Enchantments:[{{}]`, and this is what most modules use.
yes `Enchantments:[{}]` also work, will change

PR #247, opened 2019-07-02
TNT landmines

kruthers:
#### Since Last Pull request
- added skull textures
- updated lore (to 1.14)
- Updated changes to standard (for 1.14)
- optimised a few bits of code
- added a system to fix placed skulls
- added mycelium particles to invisible mines
- Raised the height of the creeper to damage the player
- Removed the silk touch requirement from shears.

PR #60, opened 2018-10-23
Tnt landmines

kruthers:
Added TNT Landmines

PR #362, opened 2020-01-11
TNT landmines for 1.15

kruthers:
4th pull request for this pack (5 if you include the one someone else made).
I have taken the old branch and deleted it so i had to make a new pr.
(old pull requests: 1.14: #298 | 1.14.4 with lots of info: #247 | og: #60)

### Added in by version of TNT landmines
**New Mines!**
Disguise your mine as:
- Grass blocks
- Dirt block
- Stone
- Stone bricks
- Crafting tables
- Nether rack
- and Invisible (crafted with diamond)

**Crafting recipes!**
For the default it is:
image
Then for custom ones it is: (replace barriers with the Disguise block/item)
image

**Upgrades!**
There is only one upgrade that is used to stop landmines grieving your terrain. Its crafting recipe:
image
The potion are normal Instant Damage II but with Zauber cauldrons installed it uses a Potion of Harming IV

**Disarming!**
Landmines can now be disarmed by dropping shears on them

**Other info**
- Invisible landmines give off mycelium particle, they give off a lot more when you hold shears near them
- the recipe was the one uses in hermit craft gm4
- they still blow up in air
- the current version has been fully tested by me but will want someone else to quick go though it

### Changes in the 1.15 update
- Updated to the new base
- nothing else realy


SandSander:
This would actually be placed on the website but not ingame?

SandSander:
I think you guys said few times it's not going to be for the public server? But maybe on the website for players there own worlds?

Bloo-dev:
This repo is not for our public server nor do we develop modules for our public server, it is merely a side project. @SandSander

SandSander:
Ok thanks for the info!
was not sure what would happen to this module.
@Bloo-dev

sparks0:
- Disarmed landmines should have a tag that stops them from attempting to re-bury. Otherwise they keep bouncing on the shears.
- Can we switch disarming to a right-click with shears? Should be able to do this by enabling the hand slot of the stand, switching from marker to invisible and checking the held item.

gm4_tnt_landmines/data/gm4_tnt_landmines/functions/checks/blowup.mcfunction
```diff
@@ -0,0 +1,4 @@
+#ran at all landmines every tick
+#@s = armed landmine
+execute positioned ~-1 ~ ~-1 if entity @a[dy=3,dx=1.1,dz=1.1,gamemode=!creative,gamemode=!spectator] run function gm4_tnt_landmines:blowup/type_check
```
misode, line 3:
```suggestion
execute positioned ~-1 ~ ~-1 if entity @a[dy=3,dx=1.1,dz=1.1,gamemode=!creative,gamemode=!spectator] at @s run function gm4_tnt_landmines:blowup/type_check
```

PR #363, opened 2020-01-11
Optimize Custom Crafting for 1.15

misode:
- Changes the `load` version of custom crafters to 2
- Copy the `Items` tag from the dropper to storage to optimize the recipe checks

## DANGER
This update will break every currently existing module that depends on custom crafters


misode:
Closing this since the optimization turned out not to be significant enough to warrant a breaking change.

Bloo-dev:
We could push these changes for the 1.16 update of our modules.

misode:
Even for 1.16 I'm not sure if this is worth it. I did some benchmarks back then and I really had to look for a performance increase. This would require updating all existing modules that depend on custom crafters.

An alternative I thought about was updating CC in a way where old modules still work, which could be considered for 1.16.

PR #369, opened 2020-02-11
Templates

misode:
This is still a draft - figuring out what the structure should be


sparks0:
Oh yeah! I was just thinking about this today. Will be nice to add recipe and shamir support to the template generator once this is figured out!

PR #370, opened 2020-02-19
Add Potion Infusers 1.15

BluePsychoRanger:
Allows Potion Tanks to feed into a potion infuser that will infuse certain blocks surrounding it and convert them into different blocks, mobs, or special infused crops. Original idea by Tesset.

Documentation:
https://docs.google.com/document/d/18QuRT6GHeDOAsC3nX_Dl5anw4UVDE5YT1XENZLPKZFM/edit?usp=sharing


SandSander:
I like this module, 
**maybe my favorite!** 
Works fine! Did a test.

SandSander:
Found bug: about whenever you fill the liquid tank, the liquid never run out so you can infinite keep for example: (netherrack to magma blocks forever)

Bloo-dev:
@SandSander You should not approve these changes in that case

SandSander:
I only found the bug after the review, 
wasn't sure it was a bug either.

sparks0:
We've had a long discussion about this module. I know it's been sitting here for a while. We sort of liked it but felt other things didn't fit and after taking a long look we realized that we feel block conversion is a little random and players might not use it, but we love the infused crops idea! It would be great to have a call and discussion about it with you but we're thinking that scaling the module down to just infuse foods would make this module fit much more neatly with GM4 and be something we'd be excited to add!

BluePsychoRanger:
Alright sounds good, I’ll be away for the rest of this week, but we can discuss it next week

BluePsychoRanger:
With the addition of a force loaded chunk the infused crops could be extended to all(? Up for discussion) potion types and all crops. Technically a force loaded chunk isn’t necessary, but it’d reduce the number of function files by having a command stored on the item nbt then having that command run by a command block.

BluePsychoRanger:
@sparks0 I'm closing this, as the upcoming Jeb combat update added potion that are stackable up to 16.

PR #377, opened 2020-03-23
1.15 Liquid Tanks, Liquid Minecarts, and Liquid Expansion Packs

### LINKED ISSUE #213: [LT:Standard Liquids] XP wasted when tank is almost full, and an XP orb passes over the tank
### Expected Behavior
The XP orb(s) passing over don't get collected or the amount of XP that can't be picked up is left in the orb.

### Current Behavior
The XP that doen't fit into the tank is wasted

### Possible Solution
Keep track of how much space there is in a tank and if an the orb has more xp in this, don't pick it up.
```python
lt_space = lt_max - lt_value # maybe also `- lt_buffer`  to be safe? Is needed if calculated/orb
# lt_util is already storing xp-value
if (lt_space >= lt_util):
    lt_space -= lt_util # only if this variable is set once for each tank, not if `lt_space` is only calculated when an orb passes
    lt_buffer += lt_util
    kill orb
```

### LINKED ISSUE #216: [LT] Liquids Tanks can use more liquid then they have
Any liquid modification is written to the buffer first, and only updated after all interactions are handled

### Expected Behavior
Using a liquid (for example using milk to clear effects) and emptying the tank with a bucket shouldn't work if there is only 1 bucket of liquid left.

### Current Behavior
It does work

### Possible Solution
Instead of using a `buffer` scoreboard, any modifications should be execute on `value` immediately and a tag `gm4_lt_value_changed` can be added instead.

### Steps to Reproduce

1.
2.
3.
etc.

### LINKED ISSUE #284: Liquid Minecarts liquid conversion bug
Note: Liquids should check tags not names of tanks they are being imported into.

SpecialBuilder32:
Updates all Liquid-Tank related datapacks to 1.15.

Implements and Closes #237 methods to enable more stable liquid interactions
Closes #216 

Improves Experience Orb handling. If a tank becomes full mid-tick, xp-orbs above will no longer be deleted. This does not check the value of the orbs however because the potential minor losses were deemed negligible compared to the additional computing required. Closes #213 

Adds liquid type to a string stored on the Tank's head to enable Liquid Minecarts to retrieve liquid type information more readily. 

Reworks much of Liquid Minecarts to have more stable tank initialization and liquid type storage by storing data on the Minecart's ArmorItems. This should fix #284 , though additional testing is required.

Publishes Zauber Liquids to the Site and adds Zauber Cauldrons to it's dependancies.


gm4_liquid_minecarts/data/gm4_liquid_minecarts/functions/init_tank.mcfunction
```diff
@@ -1,22 +1,29 @@
 #@s = liquid minecart
 #run from try_to_unload
 
-data modify block ~ ~ ~ CustomName set from entity @s CustomName
+#summon display stand
 summon armor_stand ~ ~-.45 ~ {CustomName:"\"gm4_liquid_tank_display\"",Tags:["gm4_no_edit","gm4_liquid_tank_display","gm4_lm_needs_texture"],NoGravity:1,Marker:1,Invisible:1,Invulnerable:1,Small:1,DisabledSlots:2039583,Fire:20000}
-data modify entity @e[type=armor_stand,tag=gm4_lm_needs_texture,limit=1,sort=nearest] ArmorItems[3] set from entity @e[type=armor_stand,tag=gm4_liquid_minecart_display,limit=1,sort=nearest] HandItems[0]
+data modify entity @e[type=armor_stand,tag=gm4_lm_needs_texture,limit=1,sort=nearest] ArmorItems[3] set from entity @s Passengers[0].ArmorItems[0].tag.gm4_liquid_minecarts.tank.texture
+data modify entity @e[type=armor_stand,tag=gm4_liquid_tank,limit=1,sort=nearest] ArmorItems[3].tag.gm4_liquid_tanks.liquid_tag set from entity @s Passengers[0].ArmorItems[0].tag.gm4_liquid_minecarts.liquid_tag
```
misode, line 7: There is a possibility it could modify the `liquid_tag` of another liquid tank nearby. Especially since the armor stand is summoned `~ ~-.45 ~` it would not be difficult to exploit this in survival. Adding the `tag=gm4_lm_needs_texture` check to this line would prevent that.

gm4_liquid_minecarts/data/gm4_liquid_minecarts/functions/init_tank.mcfunction
```diff
@@ -1,22 +1,29 @@
...
+#setup tank data
+data modify block ~ ~ ~ CustomName set from entity @s Passengers[0].ArmorItems[0].tag.gm4_liquid_minecarts.tank.CustomName
 scoreboard players operation @s gm4_lm_data = @s gm4_lt_max
 scoreboard players operation @s gm4_lm_data *= #5 gm4_lm_data
 scoreboard players operation @s gm4_lm_data /= #2 gm4_lm_data
 scoreboard players operation @e[type=armor_stand,tag=gm4_liquid_tank,limit=1,sort=nearest,distance=..0.5] gm4_lt_max = @s gm4_lm_data
+
+#apply liquid tag
+data modify entity @e[type=armor_stand,tag=gm4_liquid_tank,limit=1,sort=nearest,distance=..0.5] Tags append from entity @s Passengers[0].ArmorItems[0].tag.gm4_liquid_minecarts.liquid_tag
+tag @e[type=armor_stand,tag=gm4_liquid_tank,limit=1,sort=nearest,distance=..0.5] remove gm4_lt_empty
```
misode, line 19: You are doing more modifications and on this newly created entity. The same problem as described above could happen. Since you need both entities, it's not an option to run a function `as` this new armor stand. I suggest you keep the `gm4_lm_needs_texture` tag around until the end of this function and maybe rename it.

gm4_liquid_minecarts/data/gm4_liquid_minecarts/functions/init_tank.mcfunction
SpecialBuilder32: The liquid tank itself isn't the entity with the `lm_needs_texture` tag. That's the newly summoned display armor stand. This whole function runs at the center of the block that is a liquid tank, so the only way I can imagine this is exploitable is if someone moves the location of the Marker armor stand, or if they broke the tank and killed the entity sub-tick

Though a distance check might be suitable as a contingency.

misode: Oh you're right. I had in my head that this was handling minecarts. So I guess it would be fine here.

gm4_liquid_minecarts/data/gm4_liquid_minecarts/functions/init_tank.mcfunction
SpecialBuilder32: Same as my last comment. Do you think it might be worth it to give this tank a tag early in the function and only refer to it by that, or would that not decrease the number of selectors

misode: If you use a distance check, a tag would not be required to optimize. So using `@e[type=armor_stand,tag=gm4_liquid_tank,limit=1,sort=nearest,distance=..0.5]` is fine.

PR #388, opened 2020-04-08
Add Chorus Fruit to Falling Stars

Bloo-dev:
Adds a ~7% chance for Falling Stars to carry chorus fruit. Also tweaks the drop chances of diamonds, gold, iron, endstone and endermites slightly.


Bloo-dev:
I was working on this and got it to work, however, the code for most of the in built expansions is pretty ancient and will need a rework to not hurt my eyes anymore.

Bloo-dev:
A 1.16 re-write would be good, yeah. The base is not too bad but the packaged expansions are weird.

PR #389, opened 2020-04-11
Minecraft namespace consistency 'fixes'

AndanteDevs:
I noticed this abomination

```
... run particle heart ...
... run particle flame ...
... run particle barrier ...
... run particle end_rod ...
... run particle happy_villager ...
... run particle minecraft:falling_dust snow ... <---
```
And overdid a fix lmao


gm4_ambrosia/data/gm4_ambrosia/functions/main.mcfunction
```diff
@@ -9,7 +9,7 @@ execute as @e[type=vex,tag=gm4_roaming_bee] run function gm4_ambrosia:beehavior/
 
 # wax smooshing
 scoreboard players reset found_item_on_anvil gm4_bee_data
-execute as @e[type=item,nbt={Item:{tag:{gm4_ambrosia:{item:"beeswax"}}}}] at @s if block ~ ~-1 ~ #minecraft:anvil run function gm4_ambrosia:beeswax/item_on_anvil
```
SpecialBuilder32, line 12: I am like, 80% sure that block tags actually require the namespace to function, unlike sound effects and status effects.

AndanteDevs: Nope, in 80% of instances the Minecraft namespace is assumed, including this one :p

Bloo-dev: I am like, 80% sure that we should try to use the namespace every time, so we should try to fix this when modules are updated :P

SpecialBuilder32: Certainly in tags yes I agree. With sounds and particles however I think it's less important.

AndanteDevs: > block tags actually require the namespace to function

Come to think of it, you're probably thinking of the actual tag JSON files, which I believe do require the namespace to function

Kroppeb:
I'm of the opinion that we should use `minecraft:` everywhere.

Bloo-dev:
I'm with kroppeb, `minecraft:` should be used everywhere and we should slowly start to enforce this standard when updating modules.

PR #391, opened 2020-04-13
Add ender hopper minecarts

misode:
Allows ender hoppers to also be created in hopper minecarts. Same recipe.


Bloo-dev:
I tested this on the public server and it works as expected, can you confirm that this version matches the public server one? @misode

misode:
> Can you confirm that this version matches the public server one?

Yes, this is the version that's currently on the public server.

gm4_ender_hoppers/data/gm4_ender_hoppers/loot_tables/destroy.json
```diff
@@ -0,0 +1,52 @@
+{
```
SpecialBuilder32, line 1: We should probably follow the precedent of the other modules and call this `destroy_ender_hopper`

PR #392, opened 2020-04-15
Standard Crafting Recipe Polish

SpecialBuilder32:
-Added slab un-crafting
-Removed uneeded smooth-block recipies
-Updated flint recipe to be 1-to-1, preventing loss.
-Added netherwart block uncrafting
    -This one is the main one to discuss before merging. 1.16 adds "trees" of netherwart blocks, and this might make nether wart too plentiful perhaps. 

Also open to additional recipe suggestions for things the vanilla game is missing. I didn't think there were many that didn't have other suitable methods of getting them though.


misode:
I'm highly against the wart block uncrafting recipe. It's similar to magma blocks. Since both blocks generate naturally (at least in 1.16+), you don't want the base material to become too cheap. There should still be an incentive for players to build a wart or magma cream farm.

Kroppeb:
> How often do players harvest large crops of nether-wart?
1) Makes a decently large farm
2) Creates the habit of flushing down the nether-warts to replant them so I can fully fill up the farm.
3) The farm is fully planted. Has 2 large chests of nether-wart storage
4) Sees the farm fully grown and flushes the warts out of habit before realizing I already have 16 double chests of the stuff.
5) Replants it anyway so when I ever run out I don't have to wait for them to grow again/find some new warts to restart the whole farm
6) Go back to step 4

sparks0:
Yeah I'm against uncrafting nether wart blocks.

PR #394, opened 2020-04-18
Fix End Fishing for 1.15

BluePsychoRanger:
Fixed it so loot can actually be gained and fishing bait is checked correctly. 

Changed the id system to be based on player UUID instead of procedurally increasing. 

Also modified some advancements, adding an advancement for finding every sword lore that can be obtained (20 possible lores).


gm4_end_fishing/data/gm4_end_fishing/functions/reel_in/lore/sword.mcfunction
```diff
@@ -1,23 +1,23 @@
...
+execute if score @s gm4_ef_data matches 65..69 run data merge entity @s {Item:{tag:{gm4_ef_sword:14b,display:{Lore:['"If found, please return"','"to William Wallace"']}}}}
+execute if score @s gm4_ef_data matches 70..74 run data merge entity @s {Item:{tag:{gm4_ef_sword:15b,display:{Lore:['"Reforged from the shards of Narsil"']}}}}
+execute if score @s gm4_ef_data matches 75..79 run data merge entity @s {Item:{tag:{gm4_ef_sword:16b,display:{Lore:['"Forged by the Iron Sisters"']}}}}
+execute if score @s gm4_ef_data matches 80..84 run data merge entity @s {Item:{tag:{gm4_ef_sword:17b,display:{Lore:['"Used to lead the revolutionary"','"army against the Telmarines"']}}}}
+execute if score @s gm4_ef_data matches 85..89 run data merge entity @s {Item:{tag:{gm4_ef_sword:18b,display:{Lore:['"Can only be wielded by a true Gryffindor"']}}}}
+execute if score @s gm4_ef_data matches 90..94 run data merge entity @s {Item:{tag:{gm4_ef_sword:19b,display:{Lore:['"Sought by Thrax"']}}}}
+execute if score @s gm4_ef_data matches 95..99 run data merge entity @s {Item:{tag:{gm4_ef_sword:20b,display:{Lore:['"Wielded by Lion-O"']}}}}
```
Bloo-dev, line 23: This works, however, the standard for custom NBT in Gamemode 4 is to have a compound called `gm4_<module id>` within `tag` that contains all module specific NBT.
So a better solution would be to have `{Item:{tag:{gm4_end_fishing:{sword:1b}}}}`

To make these changes work, the loot table would have to be modified as well.

BluePsychoRanger: Alright, I was following suit with the Audrey messages, do those need changing as well?

Bloo-dev: Oh they use that format as well? Yeah, ideally, but not really high priority - we can make those changes with 1.16, I don't wanna trigger the "Your module is outdated" message for a small backend change

AndanteDevs: We could fix the lack of translations here too?

BluePsychoRanger: > We could fix the lack of translations here too?

Will do, thanks for reminding me

misode: The "module is outdated" message is the least of our concerns. Existing items would no longer work, so in this case I would avoid changing them.

SpecialBuilder32: In this particular case, I think that's okay. The advancement has the intent of "catch all the swords", not just "touch all the swords". 

We should be more diligent about formats in the future. 
Maybe GM4 needs some update path stuff too. Don't know how that would work.

BluePsychoRanger: I believe misode is referring to the audrey bottles, as the End Fishing Swords advancement is new to 1.15, thus will have no change.

misode: Datafixers for items are hard to do with data packs. We should avoid it at all cost.

Bloo-dev:
@BluePsychoRanger This is the version that is being used on the public server, right?

BluePsychoRanger:
> @BluePsychoRanger This is the version that is being used on the public server, right?

Yes

PR #399, opened 2020-05-01
Update Ink Spitting Squid

BluePsychoRanger:
- main runs to check all squids in the world, rather than all players in the world
- previously players wouldn't get blinded if in swim mode and the block above them isn't water (at the top layer of water)
- extended range to 3 blocks around squids
- modified play sound command to play as hostile instead of master and play in a larger radius
- fix ink sac in advancement to use CMD rather than name to indicate the custom texture support

### DIFF
--- a/gm4_ink_spitting_squid/data/gm4/advancements/ink_spitting_squid.json
+++ b/gm4_ink_spitting_squid/data/gm4/advancements/ink_spitting_squid.json
@@ -2,7 +2,7 @@
   "display": {
     "icon": {
       "item": "minecraft:ink_sac",
-      "nbt": "{display:{Name:'\"ink_spitting_squid logo\"'}}"
+      "nbt": "{CustomModelData:1}"
     },
     "title": {
       "translate": "%1$s%3427655$s",
--- a/gm4_ink_spitting_squid/data/gm4_ink_spitting_squid/functions/main.mcfunction
+++ b/gm4_ink_spitting_squid/data/gm4_ink_spitting_squid/functions/main.mcfunction
@@ -1,3 +1,3 @@
-execute as @a[gamemode=!spectator] at @s if block ~ ~1 ~ water if entity @e[type=squid,distance=..2.2] run function gm4_ink_spitting_squid:squid_squirt
+execute at @e[type=squid] as @a[gamemode=!spectator,distance=..3] at @s positioned ^ ^ ^ if block ~ ~ ~ water run function gm4_ink_spitting_squid:squid_squirt
 
 schedule function gm4_ink_spitting_squid:main 16t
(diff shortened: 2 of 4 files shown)


gm4_ink_spitting_squid/data/gm4_ink_spitting_squid/functions/main.mcfunction
```diff
@@ -1,3 +1,3 @@
-execute as @a[gamemode=!spectator] at @s if block ~ ~1 ~ water if entity @e[type=squid,distance=..2.2] run function gm4_ink_spitting_squid:squid_squirt
+execute at @e[type=squid] as @a[gamemode=!spectator,distance=..3] at @s positioned ^ ^ ^ if block ~ ~ ~ water run function gm4_ink_spitting_squid:squid_squirt
```
misode, line 1: Don't you need `anchored eyes` here? Otherwise `positioned ^ ^ ^` does nothing and it will check at the player's feet.
```suggestion
execute at @e[type=squid] as @a[gamemode=!spectator,distance=..3] at @s anchored eyes positioned ^ ^ ^ if block ~ ~ ~ water run function gm4_ink_spitting_squid:squid_squirt
```

gm4_ink_spitting_squid/pack.mcmeta
```diff
@@ -20,12 +20,8 @@
...
         "Updated by": [
             [
-                "Bloo",
-                "https://www.twitter.com/Bloo_dev"
-            ],
-            [
-                "SpecialBuilder32",
-                "https://www.twitter.com/SpecialBuilder"
+                "BluePsychoRanger",
+                "https://www.twitter.com/BluPsychoRanger"
```
misode, line 24: I think we retain previous updaters? I may be wrong though.

gm4_ink_spitting_squid/data/gm4/advancements/ink_spitting_squid.json
```diff
@@ -2,7 +2,7 @@
   "display": {
     "icon": {
       "item": "minecraft:ink_sac",
-      "nbt": "{display:{Name:'\"ink_spitting_squid logo\"'}}"
+      "nbt": "{CustomModelData:1}"
```
misode, line 5: Added this to the document, all good!

gm4_ink_spitting_squid/pack.mcmeta
BluePsychoRanger: I'm not sure tbh, if we did that would eventually create a long list tho, right?

misode: The website only takes the credits from the last version (master). Updated By is also when you're doing module specific updates. So for example I think this could warrant as such, but updating all the attributes ids to the 1.16 names shouldn't.

SpecialBuilder32: I think it's okay to overwrite/replace old updated by information in the master. Old versions will still have the credit listed in the pack.mcmeta, just not displayed on the site itself.

sparks0: I guess this is a discussion to be had so we can have a standard way of showing updated-by on the site but in this case Special is ok with it and Bloo probably is too, so we'll just go with that.

PR #400, opened 2020-05-01
Lumos Shamir

Sheepe:
Adds new metallurgy shamir "Lumos" which places torches from your inventory when the light level is too low.

The metal used is Thorium Brass.

It checks to make sure you can actually place a torch on the block before allowing it to be placed, and the specific "visible" light levels it will place in is 0 to 5.

Idea was from misode from Gamemode 4 Labs and it was made by me and @InternetAlien

Issue to consider: Since it uses setblock, it can bypass protected regions.


sparks0:
Love the concept! Misode has some good feedback requests above :)

Sheepe:
The requested changes by @misode should have been completed.

gm4_lumos_shamir/data/gm4_lumos_shamir/functions/place_torch.mcfunction
```diff
@@ -0,0 +1,8 @@
+#run from active
+#@s = player with torches
+
+setblock ~ ~ ~ minecraft:torch
+clear @s torch 1
+
+playsound minecraft:block.wood.place block @a ~ ~ ~ 1 1
+execute align xyz positioned ~0.5 ~0.5 ~0.5 run particle minecraft:firework ~ ~0.4 ~ 0.08 0 0.08 0.01 3
```
sparks0, line 8: missing newline

Bloo-dev:
I really like this, but I'm already scared of maintaining the `torch_placeable.json`

misode:
> I'm already scared of maintaining the `torch_placeable.json`

There are lots of modules that have these kind of tags, so I don't see this as a problem. Although in the future we should think about having shared block tags for these, since a lot of modules have tags with the same meaning.

Sheepe:
> Here's a thought that could be entirely dumb: What if this shamir only only worked with the new soul torches. It would give a unique use to the soul torches, and might look really cool to see tunnels lit up with blue torches. They are slightly dimmer though so I don't know how that would affect its usefulness

Considering the torches are dimmer, I feel it would impact the usefulness quite a bit, as the point is that it's supposed to help light up areas and torches are the better thing for that. We do plan on letting players use soul torches with the shamir when 1.16 comes out though, just to keep it consistent in the fact that it places torches.

PR #401, opened 2020-05-03
Add slow_falling negation to desire lines

sparks0:
This just cancels the ability for desire lines to affect blocks around the player if they have the slow falling potion effect.

(This will allow soul glass beacons to cancel desire lines' effects around them)

Edit: Moved slow_falling check from nbt to predicate

### DIFF
--- a/gm4_desire_lines/data/gm4_desire_lines/functions/player.mcfunction
+++ b/gm4_desire_lines/data/gm4_desire_lines/functions/player.mcfunction
@@ -7,4 +7,5 @@ execute if score @s gm4_desire_lines matches 30.. run scoreboard players set @s 
 execute if score @s gm4_desire_lines matches 1.. run scoreboard players remove @s gm4_desire_lines 20
 
 function #gm4_desire_lines:expansion
+scoreboard players set @s[nbt=!{ActiveEffects:[{id:28}]}] gm4_desire_lines 0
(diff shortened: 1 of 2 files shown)


gm4_desire_lines/data/gm4_desire_lines/functions/player.mcfunction
```diff
@@ -7,4 +7,5 @@ execute if score @s gm4_desire_lines matches 30.. run scoreboard players set @s
 execute if score @s gm4_desire_lines matches 1.. run scoreboard players remove @s gm4_desire_lines 20
 
 function #gm4_desire_lines:expansion
+scoreboard players set @s[nbt={ActiveEffects:[{id:28}]}] gm4_desire_lines 0
```
AndanteDevs, line 10: Shouldn't this be `Id`?

misode: Ideally it should use a predicate to detect the effect. I believe this should do it.
```json
{
  "entity": "this",
  "predicate": {
    "effects": {
      "slow_falling": {}
    }
  }
}
```

sparks0: :'D Asks in a voice call that Misode is in whether I should use a predicate or NBT and told either is fine, use nbt 🗡️

PR #402, opened 2020-05-04
Add Fulcio Shamir

BluePsychoRanger:
New Fulcio Shamir allows players to place blocks in mid-air. Works really nicely with levity!

Metal is Barimium (Modification)
Goes on a Compass

To place blocks in mid-air, the player must sneak and hold the Fulcio Compass in the main or off hand. A structure void will appear 3 blocks in front of the player's eyes (if the block is air) and a particle will display where the hitbox is. Players can then right click on the structure void and the block will be replace the structure void. Liquids will not replace the structure void, but will instead be placed as if the structure void was a normal block (against the block instead of on the block)

Notes: 
The structure void can be broken in survival, but will not drop anything and will not break the module. 
This module does not mess with other possible structure voids in the world because a "fulcio block" can only be placed in air, and is marked with an AEC which handles setting the structure void back to air.
Fulcio is latin for "prop up" or "support"


SandSander:
A nice module, works perfectly! some small details / features that can be done differently?


*  A blue line on the custom head?
+ Can be on an elytra?

BluePsychoRanger:
> * A blue line on the custom head?
It's not a custom head, it's a structure void which is the only block (to my knowledge) that has the property of replacing it when placing a block upon it.

gm4_fulcio_shamir/data/gm4_fulcio_shamir/functions/create_block.mcfunction
```diff
@@ -0,0 +1,5 @@
+#run from active
+#@s = players holding a fulcio compass and looking at air positioned ^ ^ ^3 align xyz
+
+setblock ~ ~ ~ structure_void keep
+summon area_effect_cloud ~0.5 ~0.5 ~0.5 {Radius:0,Age:-2147483648,CustomName:'"gm4_fulcio_block"',Tags:["gm4_fulcio_block"],Particle:"composter"}
```
AndanteDevs, line 5: How about ~6.8 years:

`Age:-2147483648,Duration:-1,WaitTime:-2147483648`

gm4_fulcio_shamir/data/gm4_fulcio_shamir/loot_tables/band.json
```diff
@@ -0,0 +1,36 @@
...
+        {
+            "rolls": 1,
+            "entries": [
+                {
+                    "type": "minecraft:loot_table",
+                    "name": "gm4_metallurgy:barimium_band",
+                    "functions": [
+                        {
+                            "function": "minecraft:set_nbt",
+                            "tag": "{gm4_metallurgy:{stored_shamir:\"fulcio\",custom_model_data:115}}"
```
AndanteDevs, line 12: This should be `CustomModelData` (`custom_model_data` is what it's called in resource packs)

AndanteDevs: Also, the custom model data tag is stored inside the GM4 tag - I might've overlooked something here

AndanteDevs: Ok, apparently this is something the module uses to set the tag - ignore me :p

gm4_fulcio_shamir/data/gm4_fulcio_shamir/functions/create_block.mcfunction
BluePsychoRanger: Doesn't matter really, this technically should've been set to 17 cuz all AECs with this tag are killed every cycle, but I just wanted some extra time in case for some reason it wasn't killed when it should have died, thus leaving a random structure void in the world

adriaan1313:
> which is the only block (to my knowledge) that has the property of replacing it when placing a block upon it.

Grass, tall grass, ferns and large ferns also have this property.

BluePsychoRanger:
> Grass, tall grass, ferns and large ferns also have this property.

Oh yeah! Completely forgot about those, but ofc structure voids are still the best block for this

Bloo-dev:
Currently it still checks and works on Adventure (and Spectator) players should be easy to fix, though.

PR #404, opened 2020-05-06
Shiny Items

MichaelMiner137:
This module allows the player to apply the enchantment glow to most items in the game (look at function "blacklist_items" for a list of blacklisted items). This can be used by players to draw attention to items they find special and want to show off.

To get a "shiny" item the player must simply throw an item near xp orbs. Slowly the item will consume xp from the nearby orbs and once it has collected 150 points worth of xp, it will become "shiny".

(I have no idea how to use GitHub, sorry if this is in the wrong spot)

### DIFF
--- a/gm4_orbis/data/orbis/loot_tables/chests/dungeon_20.json
+++ b/gm4_orbis/data/orbis/loot_tables/chests/dungeon_20.json
@@ -1,0 +1,13 @@
+{
+    "pools": [
+        {
+            "rolls": 1,
+            "entries": [
+                {
+                    "type": "minecraft:loot_table",
+                    "name": "gm4_orbis:chests/dungeon_20"
+                }
+            ]
+        }
+    ]
+}
--- a/gm4_orbis/data/orbis/loot_tables/chests/dungeon_30.json
+++ b/gm4_orbis/data/orbis/loot_tables/chests/dungeon_30.json
@@ -1,0 +1,13 @@
+{
+    "pools": [
+        {
+            "rolls": 1,
+            "entries": [
+                {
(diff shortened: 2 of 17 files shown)


SandSander:
This module seems interesting,
have tested and works fine!


Maybe a small number of changes?


- The amount of xp you need to enchant just 1 block is quite a lot, maybe 150 --> 75 and stack 
350-400?
- The ability to do one stack at a time?
- Weapons, tools, water bottles, don't work?

shiny items (not ingame yet)

misode:
This includes files from Orbis, more specifically 1cc13f59795ca9344864fe686b0e50937572c4d5. You made your branch from `ver/1.15`, but we're using master for all new development. Either manually remove these files or re-upload the module files to a new branch.

MichaelMiner137:
To explain some stuff:
- I do plan on adding support for enchanting stacks and I do think 150 is too much, I just wanted to get everything working first, so I could get your opinions on what needs to be changed.
- The reason you can't enchant armor and tools is that when I add the enchant to the item, I also add the HideFlags:1 tag so it won't show the enchantment. So if you were to have a "shiny" diamond helmet and then added an enchantment onto it, then that enchantment would be hidden. I originally was gonna give it a blank enchantment id so I wouldn't have to use HideFlags:1 however when I tested it on a paper server it would just remove the enchant with no id. (it even crashed when I put an invalid Minecraft enchant id)
-I will probably rewrite the module now that I know how it needs to work, this time ill use the master branch and use multiple functions to avoid using ~ 120 @e selectors

Thank you guys for the comments, making a GM4 module isn't as difficult as I first thought it would be.

PR #407, opened 2020-05-07
Move Helious into Weighted Armor

Bloo-dev:
The Helious shamir requires Weighted Armor to function, therefore it should be moved into Weighted Armor.
This PR also fixes a spelling mistake that was present in all 1.13+ Metallurgy versions, where Helious was incorrectly spelled "Helios".


misode:
I'm worried about changing the shamir ID, existing shamirs should still work. 

If we want to rename the ID, you need to do the following:
* Have a function called by `#gm4_metallurgy:apply_band` that updates the shamir ID
* Add a line in `check_item_validity` with the old ID

With these measures players would just have to remove and re-add the band.

Bloo-dev:
> With these measures players would just have to remove and re-add the band.

That's a good idea, I'll work on that!

Bloo-dev:
@misode In case you're intrested, here are a couple of screenshots from Gamemode 4D's command room (paths dimension)
helious-lore
helious-tag
metallurgy-1-12

Bloo-dev:
This should be merger sooner rather than later. I've added some code that allows players to update their 'Helios' by simply re-applying it to their gear and tested the code thoroughly.
@misode could you take a quick look at the code?

misode:
I had a quick look at the code and I believe it's doing the right things. We should probably test this on the public server first.

PR #410, opened 2020-05-08
Allow defuse shamir to work on armor stands

Dennis-0:
With this addition you could place defuse on armor stands to protect your area, so players without defuse can enter it without accidentally letting a creeper blow up

Suggestion was approved a couple months ago on GM4 Labs https://discordapp.com/channels/515225138006851624/515578870540533771/652231317148401674


jpeterik12:
this would still be useful for say a creeper farm/where a player could potentially blow up the entire farm if they are not wearing the correct armor

jpeterik12:
this would still be useful for say a creeper farm where a player could potentially blow up the entire farm if they are not wearing the correct armor

Bloo-dev:
I'd like to limit this shamir to players only, not only does it greatly reduce lag but it also keeps it consistent with other shamirs

Dennis-0:
> I'd like to limit this shamir to players only, not only does it greatly reduce lag but it also keeps it consistent with other shamirs

I understand that, just wanted to get this suggestion done since a lot of people seemed to like it in GM4 Labs

Bloo-dev:
I'm closing this as two admins have issued concerns, it can always be reopened if necessary.

PR #413, opened 2020-05-11
Adds Mushrooms to Arborenda, also fixes continuity errors

Hero29:
Add Large Mushroom to Arborenda shamir, also fixes "#run from destroy_tree" to "#run from destroy_trunk" in fell_LOGTYPE function


misode:
This has a lot of unnecessary changes because the line endings were changed

Hero29:
@misode I'm not sure why the endings were changed, I didn't change any of them myself.

Hero29:
also through some more testing, in "destory_trunk", because of the way red mushrooms can spawn,
"execute store result score @s[tag=gm4_red_mushroom_feller] gm4_arb_stat run fill ~3 ~1 ~3 ~-3 ~1 ~-3 air replace red_mushroom_block" needs to be changed to 
"execute store result score @s[tag=gm4_red_mushroom_feller] gm4_arb_stat run fill ~3 ~1 ~3 ~-3 ~0 ~-3 air replace red_mushroom_block"
this seems to fix an issue of it not breaking the bottom mushroom blocks of the shortest possible red mushroom

misode:
By default git should convert the newlines to Unix style when you commit, if that's not the case you need to modify the config. You can read more about that here: <https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings>

PR #414, opened 2020-05-11
Adds Mushrooms to Arborenda shamir, fixes continuity errors

Hero29:
Adds Large Mushroom to Arborenda shamir, also fixes "#run from destroy_tree" to "#run from destroy_trunk" in fell_LOGTYPE function


sparks0:
Going to close this. We had a big debate in admin about this requiring silk touch to work on mushrooms. Since this is an addition instead of a bug fix, and arborenda needs mushroom support for 1.16 anyway, we can re-evaluate this as a 1.16 update in future.

PR #418, opened 2020-05-31
Add Washing Tanks

adriaan1313:
### Washing Tanks
This module adds even more functionality to the water tank!
It lets you remove the colour and ink off of things.
Washing something consumes ⅓ of a bucket of water.

You can wash:
-Stained glass → glass
-Stained glass panes → glass panes
-Coloured terracotta → terracotta
-Coloured wool* → white wool
-Coloured beds* → white bed (ik it’s useless)
-Coloured carpets* → white carpet
-Filled maps → map
-Dyed leather armour → leather armour
-Written books → book
~~-Coloured glazed terracotta* → white glazed terracotta (ik it’s useless)~~
-Coloured shulker box → shulker box
It only works in the first slot like all liquid tanks things, it also only works with one item at a time

I think this is how adding on to liquid tanks works? Maybe not. If so, sorry!
\* Except for white, otherwise you would just drain the tank
** I also still need a good description, I'm bad at doing stuff that isn't a semi-funny blurb about dyes
***I did not make concrete/concrete powder washable because of cement mixers.
**** **USE THE GM4DEV VERSION OF LT/BASE, THE VERSION FROM MY FORK IS FROM 2 YRS AGO**


gm4_washing_machines/data/gm4_washing_machines/functions/wash.mcfunction
```diff
@@ -0,0 +1,122 @@
+#@s = water tank with item in first slot 
+#run from item_fill
+
+#glass
+execute if block ~ ~ ~ hopper{Items:[{Slot: 0b, id: "minecraft:white_stained_glass"}]} run function gm4_washing_machines:washing_recipes/glass
+execute if block ~ ~ ~ hopper{Items:[{Slot: 0b, id: "minecraft:orange_stained_glass"}]} run function gm4_washing_machines:washing_recipes/glass
+execute if block ~ ~ ~ hopper{Items:[{Slot: 0b, id: "minecraft:magenta_stained_glass"}]} run function gm4_washing_machines:washing_recipes/glass
+execute if block ~ ~ ~ hopper{Items:[{Slot: 0b, id: "minecraft:light_blue_stained_glass"}]} run function gm4_washing_machines:washing_recipes/glass
```
misode, line 8: This function has a lot of NBT checks. It would be better to move this to a loot table where you can use item tags to detect the input and drop the resulting washed item.

adriaan1313: I'm gonna have to look into that

adriaan1313:
- [] update the namespace to `washing tanks`
- [x] remove unused advancement
- [ ] re-check file structure

BluePsychoRanger:
> what do you mean by this?

When renaming the namespace, you’ll want to double check that all files use the correct namespace and that functions point to the correct files

gm4_washing_tanks/pack.mcmeta
```diff
@@ -0,0 +1,26 @@
...
+    },
+    "module_name": "Washing Tanks",
+    "module_id": "washing_tanks",
+    "creation_meta":"Created using the Gamemode 4 Module Template Generator (gm4.co/modules/generator)",
+    "site_description": "Dye! Die!",
+    "site_categories": [
+	    "Liquid Tanks"
+    ],
+    "video_link": "YOUTUBE_VIDEO_URL",
+    "wiki_link": "WIKI_URL",
```
BluePsychoRanger, line 14:
```suggestion
    "video_link": "",
    "wiki_link": "wiki.gm4.co/wiki/Liquid_Tanks/Washing_Machines",
```

adriaan1313:
> > what do you mean by this?
> 
> When renaming the namespace, you’ll want to double check that all files use the correct namespace and that functions point to the correct files

I mean, it works?

gm4_washing_tanks/data/gm4_washing_tanks/functions/load.mcfunction
```diff
@@ -0,0 +1,5 @@
+execute if score gm4 load matches 1 run scoreboard players set gm4_washing_tanks load 1
```
EpyonProjects, line 1:
```suggestion
execute if score gm4 load matches 1 if score gm4_liquid_tanks load matches 1 run scoreboard players set gm4_washing_tanks load 1
```
If Liquid Tanks is not installed, Washing Tanks shouldn't attempt to install.

EpyonProjects (changes requested):
Hi!   I hope you don't mind me adding some input.  I couldn't actually get it to run, so I'm just basing it off of what I read.  I tried running this first in 1.16.5, then 1.15.2, using only Base, Liquid Tanks, and Washing Tanks.  If you could take a look at this again and update it to 1.16.2+ that would be lovely!  

Changes:
* predicates that wash down to a white version shouldn't check if the white version is in the first slot, because it would (probably) constantly drain, due to checking that it has a white variant, then going white -> white over and over.


Personally, I don't think it makes much sense to wash glass, books, or maps.  In terms of fabrics (wool, carpet, beds), I think it would make the most sense if there was some sort of bleach or acid to remove the dye, instead of water, but that does not speak about the functionality.

adriaan1313:
> I tried running this first in 1.16.5, then 1.15.2, using only Base, Liquid Tanks, and Washing Tanks. 
 It works for me in 1.15.2, I will maybe update it later this week. I don't know why it doesn't work for you.
> Personally, I don't think it makes much sense to wash glass, books, or maps. In terms of fabrics (wool, carpet, beds), I think it would make the most sense if there was some sort of bleach or acid to remove the dye, instead of water, but that does not speak about the functionality.

I made this because I once accidentally crafted a white stained glass instead of a black one and I wanted to re-dye it or something bc I was out of sand or something.
Also, I was going to do something with bleach or something, but then I had to think of a logical way to get bleach and add a new liquid etc. so I didn't

Bloo-dev:
> Personally, I don't think it makes much sense to wash glass, books, or maps. In terms of fabrics (wool, carpet, beds), I think it would make the most sense if there was some sort of bleach or acid to remove the dye, instead of water, but that does not speak about the functionality.

Washing the glass actually does make some sense in a weird double-unrealstic way.
In real life you dye glass by adding traces of metal into it, the suspended particles then cause wavelength selective interference. E.g. you get red glass by adding trace amounts of gold (this is also the reason why you've never seen cheap red glassware, except for the one with an organically dyed layer on the surface).

Organic dyes simply wouldn't work to tint glass as the high temperatures would carbonise all organics.
Minecrafts way of crafting colored glass implies that it uses organics and therefore Minecraft's stained glass is normal glass with a layer of organic paint on the surface - which should be able to be washed off.

Though, I do agree that washing paper based items isn't a good idea.

adriaan1313:
For me, it works in 1.16.5, so should I just bump the pack format?

gm4_washing_tanks/pack.mcmeta
```diff
@@ -0,0 +1,32 @@
+{
+    "pack": {
+        "pack_format": 5,
```
BluePsychoRanger, line 3:
```suggestion
        "pack_format": 6,
```

gm4_washing_tanks/pack.mcmeta
```diff
@@ -0,0 +1,32 @@
...
+                "Bunnygamers",
+                "https://twitter.com/BunnygamersMC"
+            ]
+        ],
+	"With help from": [
+	    [
+	        "Misode",
+		"https://twitter.com/misoloo"
+	    ]
+	]
```
BluePsychoRanger, line 30: This may look weird on the website, might wanna consider just adding misode to the creators? Additionally, misode's twitter link is outdated (misode_).

gm4_washing_tanks/pack.mcmeta
```diff
@@ -0,0 +1,32 @@
+{
+    "pack": {
+        "pack_format": 5,
+        "description": "A Gamemode 4 Module"
+    },
+    "module_name": "Washing Tanks",
+    "module_id": "washing_tanks",
+    "creation_meta":"Created using the Gamemode 4 Module Template Generator (gm4.co/modules/generator)",
+    "site_description": "Dye! Die!",
```
BluePsychoRanger, line 9: Should this be "Die! Dye!" instead? As in kill the dyes?

gm4_washing_tanks/data/gm4_washing_tanks/functions/washing_recipes/wool.mcfunction
```diff
@@ -0,0 +1,6 @@
+#@s = water tank with one coloured wool in the first slot
+#run from wash
+
+data modify block ~ ~ ~ Items[0].id set value white_wool
```
BluePsychoRanger, line 4: these aren't consistent with current liquid tank expansion packs. it should use the `replaceitem` command (tho this will be changed to `item modify` in 1.17). Any one that just sets the item to a new id should use `replaceitem`. The ones where you actually modify item data is good to use `data remove`

gm4_washing_tanks/pack.mcmeta
adriaan1313: yes

gm4_washing_tanks/data/gm4_washing_tanks/functions/washing_recipes/wool.mcfunction
adriaan1313: I did this to preserve any names/enchantments/other stuff that might be on there, but seeing as other ones use replaceitem I should do that as well

BluePsychoRanger: Oh that's probably a better idea; people may throw compressed items in here

adriaan1313: oh yes, that is also a problem, you would be able to wash 64 for the price of 1

adriaan1313: also shulker items

sparks0 (approved):

SandSander (approved):

Bloo-dev (approved):
This will be merged by the 1st of May unless someone issues a concern.

adriaan1313:
If maps and books are being removed (y?), the predicates and the tags should go too

Issue #416, opened 2020-05-27
Levity Affects Spectators


Bloo-dev:
**Describe the bug**
Levity works on spectators. This means that spectators leave (visible!) cloud trails and get the slow falling effect.

**To Reproduce**
1. Put on some levity gear
2. Switch to spectator
3. Sneak

**Expected behavior**
Levity should ignore spectators.

**Fix**
Add a `gamemode=!spectator` selector.

**Environment**
 - Minecraft version: 1.15.2
 - Server version: 1.15.2 (Public Server 6)

misode:
This also affects enderpuffs

PR #419, opened 2020-06-01
Improves Metallurgy's Spectator Handler

Bloo-dev:
This change allows the module to revoke the "gm4_has_shamir" tags from spectators, whilst not allowing these tags to be re-applied onto spectators. This should make sure all base-packaged shamirs only work on non-spectator players, thereby fixing #416 .
Not that this only affects shamirs packaged within the metallurgy base.


Bloo-dev:
@misode Since you re-wrote metallurgy, is this change reasonable?

misode:
This makes sense yes. I don't think this would break anything.

PR #425, opened 2020-06-23
Add Endermen Deactivators

hozz8528:
Endermen Deactivators:

- Create by placing a cobblestone wall (unconnected), a grass block on it, and a wither skull on top of that, and offering it an emerald and an eye of ender.
- Will give Endermen in an area an unplaceable block to stop them from griefing.
- Will deactivate after 10 mins (half a minecraft day). Can be reactivated by throwing an emerald and an eye of ender at its base.


Bloo-dev:
Thank you for submitting this PR! Sadly, this is not the solution we are looking for.

PR #426, opened 2020-06-25
Add Server Utility Modules (Multiplayer Sleeping, AFK Detector, Increased Shulker Skells, Better Head Drops)

mcpeachpies:
**Addition for 4 new server utilities modules**

- Multiplayer Sleeping
- AFK Detector
- Increased Shulker Shells
- Better Head Drops

**Multiplayer Sleeping**
Allow for players to pass the night when 25% of online players are sleeping.
Only accounts for players in the overworld, compatible with AFK Detector

**AFK Detector**
Detect and announce AFK players to the server
Default role with AFK prefix, compatibility with additional roles can be set for prefixes
Recommended for Multiplayer Sleeping

**Increased Shulker Shells**
Guarantee a shell drop from shulkers, with a 25% of a second shell per looting level

**Better Head Drops**
Obtain player head drops from charged Creepers, can also be dropped when killed by another player with the same drop rate as Wither Skeleton skulls
Default entity skulls (Zombie, Skeleton, Creeper) have the same drop rate as Wither Skeleton skulls
Rare Steve player heads can be found in a range of structure loot chests

### DIFF
--- a/gm4_afk_detector/data/gm4/advancements/get_back_to_work.json
+++ b/gm4_afk_detector/data/gm4/advancements/get_back_to_work.json
@@ -1,0 +1,29 @@
+{
+  "display": {
+    "icon": {
+      "item": "bell",
+      "nbt": "{CustomModelData:1}"
+    },
+    "title": {
+      "translate": "%1$s%3427655$s",
+      "with": [
+        "Get back to work!",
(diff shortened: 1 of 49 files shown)


Bloo-dev:
Closing this as it contains more than 1 module submission in a single PR.

PR #428, opened 2020-06-26
Add AFK Detector

mcpeachpies:
_Server Utility Module_
**Gamemode 4 AFK Detector**
Detect, tag, and announce players away from their keyboards

Players who have not moved for 15 minutes will be announced and tagged as AFK

Players without set teams will receive an AFK prefix attached to their name
Can be configured to work with already established teams.
See the following mcfunction files under /data/gm4_afk_detector/functions/
afk_away.mcfunction
afk_back.mcfunction
init_team.mcfunction

gm4_afk tag used for players detected as AFK

Compatible with gm4_multiplayer_sleeping to exclude AFK players required sleeping players calculations

### DIFF
--- a/gm4_afk_detector/pack.mcmeta
+++ b/gm4_afk_detector/pack.mcmeta
@@ -1,0 +1,25 @@
+{
+    "pack": {
+        "pack_format": 5,
+        "description": "AFK Detector"
+    },
+    "module_name": "AFK Detector",
+    "module_id": "afk_detector",
+    "creation_meta":"Created using the Gamemode 4 Module Template Generator (gm4.co/modules/generator)",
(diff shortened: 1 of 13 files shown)


gm4_afk_detector/pack.mcmeta
```diff
@@ -0,0 +1,25 @@
+{
+    "pack": {
+        "pack_format": 5,
+        "description": "AFK Detector"
+    },
+    "module_name": "AFK Detector",
+    "module_id": "afk_detector",
+    "creation_meta":"Created using the Gamemode 4 Module Template Generator (gm4.co/modules/generator)",
```
AndanteDevs, line 8:
```suggestion
```

PR #429, opened 2020-06-28
Add Cozy Campfires (1.16)

Bloo-dev:
This module provides absorption to players near a campfire. This increases their Health by 8HP (4 ❤) for 12 minutes.
Players are **not** affected if they've recently taken damage or if they are further than 3 blocks away from a campfire, though in the later case the hearts remain for 12 minutes.

campfire-lit
(Slighlty outdated screenshot. The module now uses yellow absorption hearts.)

The module is designed with the intense use of campfires (particularly unlit) as decoration blocks in mind and runs on a slow 0.25 Hz clock. Checks on unlit campfires are kept to a minimum.

Attached are two debug reports from my test world.
```
Minecraft Version: 1.16.1
Installed Datapacks: Custom Crafters, Cozy Campfires
One campfire was present during the tests.
```

[profile-results-cozy-campfires-lit.txt](https://github.com/Gamemode4Dev/GM4_Datapacks/files/4842240/profile-results-cozy-campfires-lit.txt)
[profile-results-cozy-campfires-unlit.txt](https://github.com/Gamemode4Dev/GM4_Datapacks/files/4842241/profile-results-cozy-campfires-unlit.txt)


sparks0:
Honestly not super keen on this implementation of the idea. if I have to be within a few blocks of a campfire to get extra hearts and they vanish the moment I leave, I don't think I'd ever really use that. I would either go for regeneration, temporary golden hearts that last for 5 minutes or so, or both.

Regarding just getting regen not being UHC friendly - we have a lot of modules that aren't design for UHC at all that UHC players just wouldn't install.

grian32:
id like to know from what height above/below the campfire u can obtain the effect

Bloo-dev:
> Honestly not super keen on this implementation of the idea. if I have to be within a few blocks of a campfire to get extra hearts and they vanish the moment I leave, I don't think I'd ever really use that. I would either go for regeneration, temporary golden hearts that last for 5 minutes or so, or both.
> 
> Regarding just getting regen not being UHC friendly - we have a lot of modules that aren't design for UHC at all that UHC players just wouldn't install.

Thanks for the feedback. Yeah, I can see that absorption might be the better choice here. I'll make some changes on another day!

AndanteDevs:
> id like to know from what height above/below the campfire u can obtain the effect

It's a radius of 4 blocks from the campfire, so 4 blocks if you're directly above the campfire

grian32:
> > id like to know from what height above/below the campfire u can obtain the effect
> 
> It's a radius of 4 blocks from the campfire, so 4 blocks if you're directly above the campfire

iiiiii dont know if i like that
considering u can just put them underground and get health literally everywhere that way

AndanteDevs:
@grian32: 4 blocks? Too far
@Bloo-dev: [`8fca3bd`](https://github.com/Gamemode4Dev/GM4_Datapacks/pull/429/commits/8fca3bdf79b1b62482bf493bb6ebd7b7b7a72eba) `Increase range to 12 blocks`

grian32:
> @grian32: 4 blocks? Too far
> @Bloo-dev: [`8fca3bd`](https://github.com/Gamemode4Dev/GM4_Datapacks/pull/429/commits/8fca3bdf79b1b62482bf493bb6ebd7b7b7a72eba) `Increase range to 12 blocks`

i did say in the y axis 4 blocks was too much
but i assume this applies to all axis
so... yeah

gm4_cozy_campfires/pack.mcmeta
```diff
@@ -0,0 +1,20 @@
+{
+    "pack": {
+        "pack_format": 5,
```
Nathan22211, line 3: pack format will need to be changed to 6 for 1.16.2

Bloo-dev: Fixed!

SandSander:
Add Cozy Campfires 1.16.2 Test:

- Effect Whenever your at the campfire **Yes.**
- Campfire & Soul Campfire Working **Yes.**
Suggestion: Sound whenever you gets the effect?

SandSander:
Suggestion: 1 Advancement whenever the effect is activated at the campfire?

Bloo-dev:
I was gonna add a sound, but that would either require a tag, or a scoreboard to not be spammy. I decided against it since there are two visual queues already, so the performance loss wouldn't be justified.

PR #431, opened 2020-07-03
Add Guardian -> Elder Guardian

BluePsychoRanger:
Guardians must be splashed with potion of the turtle master and thrown a sponge to be converted into an elder guardian.

There is no farming of sponges possible with this, as elder guardians always drop only one wet sponge upon a player kill.


sparks0:
Closing and rejecting submission due to discussed aging module that would achieve this effect.

PR #432, opened 2020-07-04
Add Silverfish creation to lightning rods

BluePsychoRanger:
When a lightning rod is struck on a chiseled stone bricks block, it will be converted into a infested chiseled stone bricks block


hozz8528:
@SpecialBuilder32 Vanishing Soul Essences from the Orb of Ankou module only drop from Wandering Traders (which are extremely rare) and Silverfish (which we are locked out of the only spawner in the world being destroyed and infested blocks not existing in the world). In this current situation, it took 6 days of doing nothing but AFKing at a platform designed to catch Wandering Traders to get the necessary 13 souls (and I already had some)

Bloo-dev:
So this is a Public Server issue? These should not be on this repo then.

BluePsychoRanger:
Eh, it was suggested by someone on the server. I thought it was an interesting idea, but I suppose the use cases are very limited

Bloo-dev:
Closing this PR as it is a Public Server issue. We might add a Silverfish Spawner on the Public Server to solve that issue.

PR #436, opened 2020-07-21
Update Enchantment Extractors

BluePsychoRanger:
- Moved randomizers to loot tables and `/loot insert`

- Cursed items will extract ONLY the curse

- moved randomizer to UUID type

- Items will only have their enchantments extracted if the enchantment has an id; i.e. will not check enchantments with `Enchantments:[{}]`

- items with enchantments extracted now get the entire Enchantments[] data removed, rather than set to an empty list


BluePsychoRanger:
Yeah, should I just throw it into the `loot_tables/extract` folder as `generate.js`? Note that it is a modified version of code written by @Sheepe so we should probably get permission from him too

PR #437, opened 2020-07-23
Fix Trapped Signs Delay

BluePsychoRanger:
Currently, trapped signs do not power anything because the redstone dust is only powered for 1 game tick which is not long enough for any redstone component.

This change will extend that duration to 2 gameticks which is one redstone tick.


sparks0:
Couldn't reproduce. Tested with website version and directly on the gm4 public server. dust next to the sign could power repeaters, observers and pistons.

BluePsychoRanger:
Several people said it was broken on PS6, this fix would most likely fix the issue. (I didn’t test if it was broken, just added a guarantee— hopefully— fix)

PR #439, opened 2020-07-25
Add CustomModelData to Obsidian Casts

hozz8528:
Right now, Obsidian Casts with a metal band hold the number of the CustomModelData they will give the item they are applied to in the gm4_metallurgy tag, but don't have CustomModelData themselves (except for mundane bands). Note that Fulcio, Helious and Corripio would still need this change to be applied. (If this is merged, keep in mind that both Percurro and Defuse currently have the tag gm4_metallurgy.custom_model_data:106, and when this is used to set CustomModelData to the Obsidian Cast it would cause a problem for resource packs. I suggest changing the custom model data of percurro to a different, unused number).

### DIFF
--- a/gm4_lumos_shamir/data/gm4_lumos_shamir/loot_tables/band.json
+++ b/gm4_lumos_shamir/data/gm4_lumos_shamir/loot_tables/band.json
@@ -9,7 +9,7 @@
 					"functions": [
 						{
 							"function": "minecraft:set_nbt",
-							"tag": "{CustomModelData:114,gm4_metallurgy:{stored_shamir:\"lumos\",custom_model_data:114}}"
+							"tag": "{CustomModelData:114,gm4_metallurgy:{stored_shamir:\"lumos\"}}"
 						},
 						{
 							"function": "minecraft:set_lore",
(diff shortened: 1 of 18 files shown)


AndanteDevs:
You added `CustomModelData` to the player head itself? I didn't notice

sparks0:
in-game testing by Hozz: http://puu.sh/Gbibh/98158acebb.mp4

misode:
Since both the band and the tool now have the `CustomModelData` tag, it might be better to get rid of the `gm4_metallurgy.custom_model_data` tag. It only existed on bands so you just have to make sure to check if the tag exists when smooshing a band on a tool.

hozz8528:
I was only able to remove the gm4_metallurgy.custom_model_data tag from the obsidian casts/bands, but I had to still keep that tag in the items with bands to avoid problems with other datapacks with items with custom model data. I made a video testing the changes. http://puu.sh/GdACx/0159bf4d18.mp4

mstuyven:
Awesome! Did you make sure that old bands and old tools still work as expected? Old bands don't have `CustomModelData` but they do have `custom_model_data`.

misode:
Did you make sure that old items still work? Old bands only have the `custom_model_data` tag and not the `CustomModelData`.

hozz8528:
I made a video testing old shamirs and making sure they work correctly. http://puu.sh/GeGBR/89f66447b3.mp4

PR #441, opened 2020-08-05
Update Relocators

BluePsychoRanger:
- Relocators no longer use raycasting, making it much more reliable

- Each relocated machine block has its own set of functions meaning each machine block can choose what to allow as pickup-able and what happens when a relocator is placed on it

- Each relocated block has a custom player head instead of the block that makes up the actual block


BluePsychoRanger:
I will remake this PR for 1.16

PR #442, opened 2020-08-05
Update Relocators

BluePsychoRanger:
- Relocators no longer use raycasting, making it much more reliable

- Each relocated machine block has its own set of functions meaning each machine block can choose what to allow as pickup-able and what happens when a relocator is placed on it

- Each relocated block has a custom player head instead of the block that makes up the actual block


BluePsychoRanger:
> With how you're doing fill commands to find where people placed player heads, if theyre low enough in the world the fill command will fail due to trying to work in the void, and the relocator will just break. I don't know if this is a known bug that we're just going to live with though as I don't see any nice way to get around this.

It would also break at the top of the world afaik. The only fix I can think of would be to basically have each y-level scanned individually if the first one fails (execute store success), but that would add 13*5 commands to each machine block that supports it and would make it a lot harder to make a machine block compatible with future modules that I haven't already implemented (i.e. if someone else were to make it).

BluePsychoRanger:
@SpecialBuilder32 I've fixed that annoying fill bug; basically it checks each y-level individually.

BluePsychoRanger:
Updated version: #478

PR #449, opened 2020-08-20
Give old zauber armor CustomModelData

hozz8528:
Gives Zauber Armour pieces without Custom Model Data the value 2 (same as when created) when dropped on the floor.


hozz8528:
Oh woops. We can at least get this command on the server, right? Poor Michael just made every Zauber Armour set.

Bloo-dev:
I agree, especially since this isn't a breaking change.

misode:
This patch was put on the server but we will not be including this in the module on the site.

PR #451, opened 2020-08-29
Fix double band creation bug

### LINKED ISSUE #450: Getting 2 Bands from 1 mould in metallurgy
**Describe the bug**
When you time it right you can break the sand block the mould is in and it will return 2 bands

**To Reproduce**
1. Create a mould
2. Either manually drop 23 ores onto it or just set them
3. Set the sidebar to the gm4_ml_heat scoreboard
4. Break the sand on 1 and it can return 2 bands

**Expected behavior**
It only return 1 ban

**Fix**
Add an extra check to not return a band when the heat equals 0 if there is no sand

**Screenshots**

**Environment**
 - 1.15.2 / 1.16.2
 - Server version: 1.15.2 (public server)
 - Installed Datapacks: Metallurgy and all its addons, Custom Crafters, Ziprails, Enchant Extractors, Block Compressors, Tinkering Compressors
 - Installed Mods: none

misode:
Fixes #450


kruthers:
This looks to have fixed the issue tested it about 10 times both manually and with command blocks

PR #466, opened 2020-08-30
[Trapped Signs] Added support for crimson and warped signs

kruthers:
Added the crimson and warped signs as trapped signs.

Also changed the secret sign to drop a black dye as that's what you craft it with


sparks0 (approved):

misode (approved):

SpecialBuilder32 (approved):
Tested and it works.

PR #473, opened 2020-09-06
Fix Netherite Moneo

BluePsychoRanger:
updates MaxDurability tag of tools that had moneo before being upgraded to netherite


SpecialBuilder32 (approved):
I've tested these changes and they all look good
I did noticethe usage of the tag `MaxDurability` was a bit odd since we usually put stuff in out own tags (though that's probably leftover from the original coding). Not sure if this is a good opportunity to change that or not, or even if its worth addressing. I don't know what it may break.

gm4_moneo_shamir/data/gm4_moneo_shamir/predicates/holding_moneo.json
```diff
@@ -1,29 +1,12 @@
...
-      }
-    },
-    {
-      "condition": "minecraft:entity_properties",
-      "entity": "this",
-      "predicate": {
-        "equipment": {
-          "offhand": {
-            "tag": "gm4_moneo_shamir:tools",
-            "nbt": "{gm4_metallurgy:{has_shamir:1b,active_shamir:'moneo'}}"
```
misode, line 23: Players can also use up durability of tools in the offhand, for example by stripping wood with an axe.

misode (changes requested):
Fix works.

gm4_moneo_shamir/data/gm4_moneo_shamir/predicates/holding_moneo.json
BluePsychoRanger: I removed the offhand check as the rest of the code did not use it at all; `active_tool` only checks for the mainhand, so the offhand was never being checked. Additionally, mining fatigue would not have an effect on right click abilities, so it doesn't really matter to check the offhand for that.

misode (approved):

PR #477, opened 2020-09-16
Metallurgy band recycling

hozz8528:
Makes it possible to recycle the metal from existing bands to make new ones. You only get the equivalent of 12 lumps from recycling a band (9 from the primary metal and 3 from the secondary in the case of alloys). Also added something for when you try to "melt" other player head items.

Here's a video of this in action (I've changed the position of the summoned items a bit since then)
https://puu.sh/Gt7B1/2de9d2609c.mp4


Bloo-dev (changes requested):
This PR has **a lot of duplicate code and some unintuitive features**; Both of these issues could be solved by integrating the mechanic into the ore melting mechanic.
- Adding metal to a cast is usually done by dropping the ore on, I don't see a reason to alter that behavior for bands, especially since players are discouraged from placing bands in normal gameplay due to the loss of the shamir upon breaking.
- Metallurgy does already have a system that handles the jump animation and subsequent addition of metal to a cast. One could tap into that system and allow bands to be used as well; and using a switch case (`ore ^ band`) add metal accordingly.

Bloo-dev:
Closing as this is being implemented by #487 .

PR #479, opened 2020-09-19
Update relocators for reliability

BluePsychoRanger:
- removes raycasting to be more reliable
- each machine block gets a custom player head rather than the relocated block being the actual block it was made of
- each machine block runs its own checks and removals to ensure it works properly with relocators
- if a machine block doesn't have code to support interactions with relocators, it will act as if it's a normal block that can't be relocated

Each machine block need 5 files:
  - 2 functions
  - 1 function tag
  - 1 loot table


BluePsychoRanger:
I've tested this pretty extensively and the only issue I've not been able to fix is that if two players place a relocator on the same block within the same tick, one relocator will be deleted (the relocated block will not, so it's just a loss in relocator resources - popped chorus, diamond, silk touch book)

BluePsychoRanger:
This PR is not ready to be merged; there is a bug (currently only for Liquid Tanks) that doesn't allow modules to run separate commands for picking up the machine block (so filled liquid tanks will leave the display armor stand, cuz it doesn't get killed initially)

BluePsychoRanger:
The module should now be ready to be merged

misode (approved):

PR #481, opened 2020-09-27
Add Trash Cans

MichaelMiner137:
Purpose:

- Trash Cans provide an easy way to destroy items without taking up much space. It only deletes items when the player has walked away, thus protecting the player from accidentally destroying valuables.

Creation:

- It is created in a custom crafter using 7 cobblestone in the shape of a U with a magma block on top and a lava bucket in the middle.

Features:

- By default, items within the Trash Can will not be deleted unless there are no players within 15 blocks. They will then be deleted at a rate of 16 items every 16 ticks.
- If the Trash Can is powered it will delete items at a rate of 64 items every 16 ticks, regardless of players nearby.
- If the Trash Can contains any netherite based items, it will be disabled as netherite can not be burned in lava.
- If the Trash Can is touched by water it will be "cooled down" preventing items from being deleted. This serves as a way for players to disable it.


MichaelMiner137:
There was a mistake with the repository so I had to delete it and resubmit it.

grian32:
Could you feed into this with a hopper? I think it'd be a good thing to have to distance this from vanilla and make it it's own thing.

hozz8528:
I'm trying really hard to come up with ways to justify needing this. I really like how it's implemented and that it can be used as a redstone component to detect players, but I you can still do everything this can in vanilla.

I really like how netherite items can turn it off, and I think there's potential to expand on this function to make this worth using over a vanilla redstone thing. Maybe something could work as an interface to prevent other kinds of items from being destroyed. Maybe that or some other way to further differentiate this from what you can already do would be help.

grian32:
> I'm trying really hard to come up with ways to justify needing this. I really like how it's implemented and that it can be used as a redstone component to detect players, but you can still do everything this can do in vanilla.
> 
> I really like how netherite items can turn it off, and I think there's potential to expand on this function to make this worth using over a vanilla redstone thing. Maybe something could work as an interface to prevent other kinds of items from being destroyed. Maybe that or some other way to further differentiate this from what you can already do would be help.

Shadow suggested this on discord:
image
IMO I think it's great, and it also stops people's gear from getting destroyed if something happens.

Bloo-dev:
I'm closing this PR as it has been stale for a while and has not seen general approval by the admins and our testers. A trash can can easily be set up using survival means.

PR #485, opened 2020-11-12
Add Teleportation Anchors

BluePsychoRanger:
Adds a block that can prevent teleportation and another block that can force teleportation to a specific location

https://docs.google.com/document/d/1EytQ5IcxTRx885PJk6NWVhBy0_kww1RJF5lj9icZZII/edit?usp=sharing


BluePsychoRanger:
I found another bug that is pretty annoying to fix... if a Jammed Enderman or Shulker goes through an end gateway, they are marked as teleported, but that form of teleportation isn't chorus-based and should be ignored. There's already a tag to ignore entities, but running a check for the block every tick doesn't work cuz end gateways teleport entities immediately. There are three options: 
A) Ignore this as it won't happen too often (lore could be that end gateways are in fact chorus based; it could be valid, as end gateways are similar to chorus in that it has a range and will somewhat randomly move entities to a location within that range)
B) Make it so all portal types negate nearby jammers
C) Make it so end gateways negate nearby jammers

Also, it seems that inter-dimensional travel doesn't have this issue, likely because the entity doesn't exist for some time while it's being moved to the other dimension.

Bloo-dev:
> A) Ignore this as it won't happen too often (lore could be that end gateways are in fact chorus based; it could be valid, as end gateways are similar to chorus in that it has a range and will somewhat randomly move entities to a location within that range)

This actually seems very reasonable to me in this case, though I'd love to hear what @SpecialBuilder32 thinks.

BluePsychoRanger:
> I was always kind-of of the mindset that all teleportation is end-based, whether that was gotten to from the plants that grow there, the creatures that grow there, or... rocks that grow there. (This list was more concise than I expected). What was the reason for not preventing ender-pearling? (I'm okay with not disabling percurro, but that's cause it feels more like an anime-zoom-sword-slash than actual teleportation)

(Other than the fact that it'd be hard to detect...) Ender pearls are different than chorus based teleportation. Chorus-based teleportation is random and doesn't need a direct line of sight to the block that the player is teleported. Endermen and Shulkers follow this rule as they can teleport from a completely enclosed space and their teleportation is random and has limitations to the destination (no block collision and no liquids). 

Ender pearls on the other hand don't have the same destination limitations (can TP into liquids/blocks) and need a direct line of sight (players cannot teleport out of a completely enclosed space, other than glitching through blocks). Nether/End portals don't work like this because their locations are exact and there's a determined destination (also they're inter dimensional).

BluePsychoRanger:
Oh... I just remembered the end gateway thing only currently works with endermen and shulkers ATM, not with players. Should I try to add that jammed players are prevented from end gateway teleportation?

gm4_teleportation_anchors/data/gm4_teleportation_anchors/functions/player/visuals_jam.mcfunction
```diff
@@ -0,0 +1,8 @@
+# @s = player that was jammed
+# run from player/find_aec
+
+particle minecraft:angry_villager ~ ~1.3 ~ 0.3 0.2 0.3 1 5
```
SpecialBuilder32, line 4: Maybe use local coordinates to force this particle infront of the player? That way it is a little more visible.

BluePsychoRanger: Made this change with the latest commit

sparks0:
How did we feel this module worked in testing on the last server? Did players use it? What did they use it for? Did we get any feedback?

I haven't played in survival with it myself but personally I'd love to use it to make pipes in my base to get around different sections of it quickly, but I think the current cost of the lodestone based anchor and the 20 second cooldown makes that a little impractical. I'd be in favour of maybe swapping out the lodestone part of the recipe for a respawn anchor and possibly halving the tp cooldown... what are people's thoughts?

BluePsychoRanger:
This module was put on the server while not many people were playing PS6 anymore, so unfortunately we didn't get much feedback. The only player I know of that used it was Hozz, for prevention of endermen near his base. I would agree with halving the tp cooldown, but as for the recipe change, using a respawn anchor (as the block itself) will allow players to use the respawn anchor as a normal respawn anchor, and idk if that would make sense lore-wise. A lodestone was used for a few reasons: it doesn't have an inventory; it's an immovable block, so pistons can't destroy them; it represents navigation throughout the pipes; right clicking it would still make sense lore-wise as you can either teleport to navigate or just use the old-fashioned compass.

Bloo-dev:
We'll wait for people to use this on the current public server to gather feedback, once that has happend this is ready to be merged.

Bloo-dev:
I don't think this will be merged for 1.17. Our new aim is to get this merged in January 2022, once it ran on the public server for a couple of weeks.

gm4_zauber_cauldrons/data/gm4_zauber_cauldrons/functions/player/wormhole_targeting/prepare_teleport.mcfunction
```diff
@@ -2,6 +2,9 @@
 # at @s
 # run from advancement zauber_cauldrons:use/wormhole
 
+# compatibility with teleportation anchors
+execute if score gm4_teleportation_anchors load.status matches 1.. run function gm4_teleportation_anchors:player/used_chorus
```
misode, line 6: ZC should add a function tag instead

gm4_zauber_cauldrons/data/gm4_zauber_cauldrons/functions/player/wormhole_targeting/set_dimension.mcfunction
```diff
@@ -16,4 +16,4 @@ tp @a[tag=gm4_zc_wormhole_consumer,limit=1] ~.5 ~.7 ~.5
 # play arrival animation
 particle minecraft:portal ~.5 ~1.3 ~.5 .25 .25 .25 0 100 force @a[distance=..32]
 execute positioned ~.5 ~.7 ~.5 run playsound minecraft:entity.enderman.teleport player @a[distance=..8] ~ ~ ~ 1 .3
-execute unless entity @e[type=area_effect_cloud,tag=gm4_zauber_cauldron,dx=0,dy=0,dz=0] run summon tnt ~.5 ~.7 ~.5
+execute unless entity @a[tag=gm4_zc_wormhole_consumer,tag=gm4_ta_jammed,limit=1] unless entity @e[type=area_effect_cloud,tag=gm4_zauber_cauldron,dx=0,dy=0,dz=0] run summon tnt ~.5 ~.7 ~.5
```
misode, line 19: I'd much rather see a `gm4_zc_prevent_explosion` tag, not specific to TA

gm4_teleportation_anchors/data/gm4_teleportation_anchors/advancements/join.json
```diff
@@ -0,0 +1,10 @@
+{
+    "criteria": {
+        "consume chorus": {
```
misode, line 3: Wrong criterion name

misode (changes requested):
This PR needs to be updated to 1.17 and 1.18, in particular, #627 moves the `set_dimension` function which this PR modifies, the `gm4_ta_jammed` tag doesn't work anymore in the current state. I'd recommend pulling from master and resolving the conflicts.

gm4_zauber_cauldrons/data/gm4_zauber_cauldrons/functions/player/wormhole_targeting/set_dimension.mcfunction
BluePsychoRanger: This has been changed to a score check for ZC. I was going to implement it as a tag, but then modules would try to remove the tag and there's no way to verify that another module still wants the tag set

Bloo-dev (approved):
Passed in-game testing in 1.19.

PR #494, opened 2020-12-02
Add stone_button to smelteries

hozz8528:
This is slightly bigger than the other stone_button PRs, but it does essentially the same (allows me to retexture/remodel the block, but in this case it allows me to keep the display head). Some CMD switching had to happen in find_furnace because of the rotation, and it still glitches when it tries to find a new furnace, but hey, what can you do (at least it lasts only a fraction of a second. You can see it in the video below)

Also, it gives the display blocks custom model data so they can be retextured

Here's a preview of what can be done with the button (the files for this are in a PR in resources):
http://puu.sh/GThH2/3c74f2efee.mp4

### DIFF
--- a/gm4_smelteries/data/gm4_smelteries/functions/find_furnace.mcfunction
+++ b/gm4_smelteries/data/gm4_smelteries/functions/find_furnace.mcfunction
@@ -2,12 +2,30 @@
 # @s - smeltery that isn't pointing to a furnace
 # called by process
 
-execute at @s if entity @s[tag=gm4_bf_has_furnace] run data merge entity @s {HandItems:[{},{id:"minecraft:stone_button",Count:1b,tag:{CustomModelData:5}}],Pose:{Head:[0f, 0f, 0f],LeftArm:[0f, 0f, 0f]}}
-execute at @s if entity @s[tag=gm4_bf_has_furnace] run tp @s ~ ~ ~ -45 0
+# save initial state change
+scoreboard players add furnace_state_initial gm4_bf_data 0
+execute if entity @s[tag=gm4_bf_has_furnace] run scoreboard players add furnace_state_initial gm4_bf_data 1
+
+# check for furnaces
 tag @s remove gm4_bf_has_furnace
 execute if block ~1 ~ ~ furnace[facing=east] run tp @s ~ ~ ~ -90 0
 execute if block ~-1 ~ ~ furnace[facing=west] run tp @s ~ ~ ~ 90 0
 execute if block ~ ~ ~1 furnace[facing=south] run tp @s ~ ~ ~ 0 0
 execute if block ~ ~ ~-1 furnace[facing=north] run tp @s ~ ~ ~ 180 0
 execute at @s if block ^ ^ ^1 furnace run tag @s add gm4_bf_has_furnace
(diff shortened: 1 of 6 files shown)


gm4_smelteries/data/gm4_smelteries/functions/find_furnace.mcfunction
```diff
@@ -8,4 +8,6 @@ execute if block ~-1 ~ ~ furnace[facing=west] run tp @s ~ ~ ~ 90 0
 execute if block ~ ~ ~1 furnace[facing=south] run tp @s ~ ~ ~ 0 0
 execute if block ~ ~ ~-1 furnace[facing=north] run tp @s ~ ~ ~ 180 0
 execute at @s if block ^ ^ ^1 furnace run tag @s add gm4_bf_has_furnace
-execute at @s unless block ^ ^ ^1 furnace run tp @s ~ ~ ~ 45 0
+execute at @s if block ^ ^ ^1 furnace run data merge entity @s {HandItems:[{},{id:"minecraft:stone_button",Count:1b,tag:{CustomModelData:4}}],Pose:{Head:[0f, 0f, 0f],LeftArm:[0f, 0f, 0f]}}
+execute at @s unless block ^ ^ ^1 furnace run tp @s ~ ~ ~ -45 0
+execute at @s unless block ^ ^ ^1 furnace run data merge entity @s {HandItems:[{},{id:"minecraft:stone_button",Count:1b,tag:{CustomModelData:5}}],Pose:{Head:[0f, 0f, 0f],LeftArm:[0f, 0f, 0f]}}
```
Bloo-dev, line 13: This needs to be implemented more carefully; `find_furnace` is clocked at 1.25Hz so we need to limit costly operation such as NBT operations.
This item updating only needs to happen once, right? As in, once the armor stand is holding the correct button it doesn't need be given the same button again 16 ticks later.
We should limit these NBT operations so that they only happen once the "gm4_bf_has_furnace" state changes.

gm4_smelteries/data/gm4_smelteries/loot_tables/smeltable_display.json
```diff
@@ -10,6 +10,12 @@
                         {
                             "type": "minecraft:item",
                             "name": "minecraft:gold_ore",
+							"functions": [
+								{
+									"function": "minecraft:set_nbt",
+									"tag": "{CustomModelData:1}"
+								}
+							],
```
Bloo-dev, line 18: This is an amazing idea! Allows us to pixel align the display 👍

Bloo-dev (changes requested):

gm4_smelteries/data/gm4_smelteries/functions/find_furnace.mcfunction
hozz8528: I *think* I took care of it... at least it still works.

Bloo-dev:
I don't think [4f3cd0a](https://github.com/Gamemode4Dev/GM4_Datapacks/pull/494/commits/4f3cd0a269793b880d288df47f305d25d0944ce5)is sufficient, as it still does at least one NBT operation per pulse. I'll work on a solution!

Bloo-dev (approved):

PR #497, opened 2020-12-06
Add Ancient Debris to Smelteries

BluePsychoRanger:
Ancient Debris is a metal ore, so smelteries should be able to double it.


BluePsychoRanger:
Nevermind, it was discussed that this should not be added.

PR #500, opened 2020-12-07
Tunnel bores model cosmetics

hozz8528:
DEMO: http://puu.sh/GV19J/180728e048.mp4

This will change the piston's orientation depending on where the minecart is facing and will change the orientation back to "up" if it fails to place a rail or runs out. It makes it much easier to see when the tunnel bore is set up properly and looks nicer overall (except for the fraction of a second when it turns some corners or goes up or down, but all minecarts with block passengers do that, and tunnel bores rarely do ¯\_(ツ)_/¯). I tried to limit the amount of NBT operations made by adding temporary tags when rotating so that the NBT operations happen once per turn (and tunnel bores don't really turn all that much in practice) and similarly when clearing the orientation back to up. Feel free to (please!) make suggestions on how to improve this.

### DIFF
--- a/gm4_tunnel_bores/data/gm4_tunnel_bores/functions/bore/direction/update_rotation.mcfunction
+++ b/gm4_tunnel_bores/data/gm4_tunnel_bores/functions/bore/direction/update_rotation.mcfunction
@@ -2,6 +2,11 @@
 # at world spawn
 # called by gm4_tunnel_bores:update_direction
 
+execute if entity @s[tag=!gm4_tunnel_bore_south] if score @s gm4_bore_data matches 0 at @s run function gm4_tunnel_bores:bore/direction/update_rotation_south
+execute if entity @s[tag=!gm4_tunnel_bore_west] if score @s gm4_bore_data matches 90 at @s run function gm4_tunnel_bores:bore/direction/update_rotation_west
+execute if entity @s[tag=!gm4_tunnel_bore_east] if score @s gm4_bore_data matches -90 at @s run function gm4_tunnel_bores:bore/direction/update_rotation_east
+execute if entity @s[tag=!gm4_tunnel_bore_north] if score @s gm4_bore_data matches 180 at @s run function gm4_tunnel_bores:bore/direction/update_rotation_north
+
 execute if score @s gm4_bore_data matches 0 at @s run tp @s ~ ~ ~ 0 ~
 execute if score @s gm4_bore_data matches 90 at @s run tp @s ~ ~ ~ 90 ~
 execute if score @s gm4_bore_data matches -90 at @s run tp @s ~ ~ ~ -90 ~
--- a/gm4_tunnel_bores/data/gm4_tunnel_bores/functions/bore/create.mcfunction
+++ b/gm4_tunnel_bores/data/gm4_tunnel_bores/functions/bore/create.mcfunction
@@ -4,7 +4,7 @@
 
 # kill existing cart and summon new cart
 kill @s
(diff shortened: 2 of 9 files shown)


BluePsychoRanger (approved):

gm4_tunnel_bores/data/gm4_tunnel_bores/functions/bore/direction/update_rotation.mcfunction
```diff
@@ -2,6 +2,11 @@
 # at world spawn
 # called by gm4_tunnel_bores:update_direction
 
+execute if entity @s[tag=!gm4_tunnel_bore_south] if score @s gm4_bore_data matches 0 at @s run function gm4_tunnel_bores:bore/direction/update_rotation_south
+execute if entity @s[tag=!gm4_tunnel_bore_west] if score @s gm4_bore_data matches 90 at @s run function gm4_tunnel_bores:bore/direction/update_rotation_west
+execute if entity @s[tag=!gm4_tunnel_bore_east] if score @s gm4_bore_data matches -90 at @s run function gm4_tunnel_bores:bore/direction/update_rotation_east
+execute if entity @s[tag=!gm4_tunnel_bore_north] if score @s gm4_bore_data matches 180 at @s run function gm4_tunnel_bores:bore/direction/update_rotation_north
+
 execute if score @s gm4_bore_data matches 0 at @s run tp @s ~ ~ ~ 0 ~
```
BluePsychoRanger, line 10: I believe these tp commands can go in `gm4_tunnel_bores:bore/direction/update_rotation_<N/E/S/W>`

hozz8528: I tried that, but this caused a bug when detecting where to place rails for some reason. I'll show you a video of what happened earlier. I'll try again and investigate.

hozz8528: Yeah, I tried again and it's still there :/. I really have no idea why that change would have that effect. I only tested when changing directions to south, but it was enough to make me do things this way since i had no idea what the origin of the issue was (btw, took me forever to figure out exactly what change caused it since I couldn't see the correlation between the change and the effect ;__;)

misode (approved):
Needing these 4 tags is a bit annoying, but it's probably the best and most efficient solution, maybe except a scoreboard. I quite like the visual changes!

Bloo-dev:
I'm kinda scared to commit this without very thorough testing. Tunnel Bores sadly has to do some weird stuff to combat minecarts, just looking at it the wrong way may break stuff.

misode:
> I'm kinda scared to commit this without very thorough testing
This PR only ever modifies the `DisplayState` and doesn't touch the other code.

hozz8528:
> 
> 
> > I'm kinda scared to commit this without very thorough testing
> 
> This PR only ever modifies the `DisplayState` and doesn't touch the other code.

Minecarts do seem to work in mysterious ways, though. I remember I tried to optimize some code and the module broke in ways that I could not connect to the changes I made (there's a video of the error somewhere in the conversation). I tested the changes in their current state quite a bit, but couldn't find any more bugs.

SpecialBuilder32 (approved):
Tested and it all looks good to me. It does only change the visuals of the cart so I'm not worried about it breaking things. 

However, it will change the texturepack-less appearance of the cart. See the attached picture.
image

@Bloo-dev What are your thoughts on this new appearance, as you were the one who decided on the current appearance?

sparks0:
We have now placed this module on the server for testing to see if anything weird happens! We'll review in a week to push if nothing seems untoward.

PR #503, opened 2020-12-09
Pack.png and in-game descriptions for Data Packs

BluePsychoRanger:
Makes data packs visually appealing when opening them in the Data Pack Selection screen.

- Adds icons for all modules
- Makes module description have the module name and module versio


misode (commented):
I think it was decided that we're not translating module names, so the pack.mcmeta files should not use any translations.

Bloo-dev (approved):
This looks good, I'm not opposed to not translating the module names, but maybe leaving in the translation keys isn't too bad? @misode

PR #505, opened 2020-12-24
Add Apple Trees

Bloo-dev:
Apple trees adds trees that grow apples once every 3 in game days.
Apple trees can be found in orbis generated chunks or can be grown using saplings, which are obtained from apple bearing leaves. Oaks and dark oaks will no longer drop apples.
Apple trees are fully deterministic and the seed is obtained via the following options:
- sapling based generation: The seed is random (extracted from UUID)
- world generation: The seed is constructed from the world seed and the position of the apple tree

The included base (Fruiting Trees) provides a simple interface for seed extraction and structure generation. Fruiting trees provides support for 16 bit seeds and n-layer layered structure generation.
Seed interpretation and structure generation have to be implemented by expansions.

This module requires orbis.


gm4_apple_trees/data/gm4_fruiting_trees/functions/load.mcfunction
```diff
@@ -0,0 +1,7 @@
+execute if score gm4 load matches 1 if score gm4_orbis load matches 1 run scoreboard players set gm4_fruiting_trees load 1
+execute unless score gm4 load matches 1 run data modify storage gm4:log queue append value {type:"missing",module:"Apple Trees",require:"Gamemode 4"}
+execute unless score gm4_orbis load matches 1 run data modify storage gm4:log queue append value {type:"missing",module:"Apple Trees",require:"Orbis"}
```
misode, line 3: These checks should either be removed or moved to gm4_apple_trees:load, to avoid duplicate errors.

BluePsychoRanger:
Also, this PR changes a few other data packs:
* Podzol Rooting Soil requires lib-forceload and can now support custom saplings
    * Apple Tree Saplings support being planted by Podzol Rooting Soil
* lib-forceload now contains a command block

gm4_apple_trees/data/gm4_fruiting_trees/functions/load.mcfunction
BluePsychoRanger: oh.. we forgot to change the names. Those are supposed to say "Fruiting Trees", not "Apple Trees"

BluePsychoRanger, line 3: @misode this is what the commands _should_ be. This is correct then, right?
```suggestion
execute unless score gm4 load matches 1 run data modify storage gm4:log queue append value {type:"missing",module:"Fruiting Trees",require:"Gamemode 4"}
execute unless score gm4_orbis load matches 1 run data modify storage gm4:log queue append value {type:"missing",module:"Fruiting Trees",require:"Orbis"}
```

misode: But fruiting trees and apple trees are supposedly always packaged together. Only the actual name of the module should announce errors.

PR #511, opened 2021-01-20
Add an armor stand with a CMD stone button to metallurgy casting

hozz8528:
This adds an armor stand holding a stone button when casting. The stone button is completely invisible without using a resource pack (it's completely inside the head), and I tried implementing it in a way that doesn't add too many commands. It DOES require adding an invisible armor stand to make it work, but this shouldn't be an issue since it's only 1 and it only lasts a few moments (you guys just added apple trees, c'mon guys). I also made it so that when adding ores/bands the head/CMD change only happens when the vex doesn't have the gm4_contains_metal tag to limit the number of NBT reads (I think this makes it slightly faster)

EDIT: This is so models can be added on top of the sand ring, like this: http://puu.sh/H8SZi/0b1f0cca41.mp4


gm4_metallurgy/data/gm4_metallurgy/functions/casting/sustain_mould.mcfunction
```diff
@@ -27,6 +27,8 @@ execute if score @s[tag=!gm4_contains_metal] gm4_ml_heat matches 48..49 run data
 #cool ring with metal
 execute if score @s[tag=gm4_contains_metal] gm4_ml_heat matches 48..49 run data merge entity @s {ArmorItems:[{},{},{},{id:"minecraft:player_head",Count:1,tag:{CustomModelData:25,SkullOwner:{Id:[I;140191582,-1151712444,-1271443335,826741136],Properties:{textures:[{Value:"eyJ0aW1lc3RhbXAiOjE0ODE0MTE1NjYwMjQsInByb2ZpbGVJZCI6Ijk4NWIyNzVlYmI1YTQzNDRiNDM3Njg5NTI4NjNhNjNmIiwicHJvZmlsZU5hbWUiOiJTcGFya3MiLCJ0ZXh0dXJlcyI6eyJTS0lOIjp7InVybCI6Imh0dHA6Ly90ZXh0dXJlcy5taW5lY3JhZnQubmV0L3RleHR1cmUvZjM3ZjE2Zjc2MTk1ZWM3OWI4MzFmMzc5ZjkzYjYwYzg2MWZkNGNhYmYyZTI5ZDNmZTlkYTJmODU4OTJkY2EifSwiQ0FQRSI6eyJ1cmwiOiJodHRwOi8vdGV4dHVyZXMubWluZWNyYWZ0Lm5ldC90ZXh0dXJlLzVjM2NhN2VlMmE0OThmMWI1ZDI1OGQ1ZmE5MjdlNjNlNDMzMTQzYWRkNTUzOGNmNjNiNmE5Yjc4YWU3MzUifX19"}]}}}}]}
 
+#update display armor stand
+execute if score @s gm4_ml_heat matches 48..51 as @e[type=armor_stand,tag=gm4_sand_ring_display,limit=1,sort=nearest,distance=..1.2] run data modify entity @s HandItems[0].tag.CustomModelData set from entity @e[type=vex,tag=gm4_sand_ring,limit=1,distance=..1.2] ArmorItems[3].tag.CustomModelData
```
Bloo-dev, line 31: I don't fully know how Custom Model Data in Metallurgy works, but is the Custom Model Data on the vex still needed after this PR?

gm4_metallurgy/data/gm4_metallurgy/functions/casting/create_mould.mcfunction
```diff
@@ -2,6 +2,8 @@
...
+summon armor_stand ~.65 ~-0.545 ~.5 {Small:1,NoGravity:1,Marker:1,Invisible:1,Silent:1,DisabledSlots:2039552,Tags:["gm4_no_edit","gm4_sand_ring_display"],Fire:200000,CustomName:'"gm4_sand_ring_display"',HandItems:[{id:"minecraft:stone_button",Count:1b},{}],Pose:{RightArm:[0f, 0f, 0f]}}
+execute as @e[type=armor_stand,tag=gm4_sand_ring_display,limit=1,sort=nearest,distance=..1.2] run data modify entity @s HandItems[0].tag.CustomModelData set from entity @e[type=vex,tag=gm4_sand_ring,limit=1,distance=..1.2] ArmorItems[3].tag.CustomModelData
```
Bloo-dev, line 6: I don't know if this would make a huge difference but as moulds are snapped to the block grid using
`align xyz <...> @e[dx=0<...>` might be safer than `distance=..1.2` and would allow getting rid of the sorting.
However, having to align the commands might be more of a hassle than it is worth and/or might add extra lag when a new execute command needs to be added.

misode:
Why not put the stone button on the vex?

hozz8528:
the vexes arms move :) (even with noAI. I posted a video a while ago)

gm4_metallurgy/data/gm4_metallurgy/functions/casting/sustain_mould.mcfunction
hozz8528: I think you're right. It might be a better idea to remove the CMD from the vex head altogether and just have the display stone button have the CMD. What do you/the rest think about this?

gm4_metallurgy/data/gm4_metallurgy/functions/casting/create_mould.mcfunction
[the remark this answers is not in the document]
hozz8528: Right. I'll make it appear with the CMD instead of copying it from the vex head.

gm4_metallurgy/data/gm4_metallurgy/functions/casting/add_metal/add_aluminium.mcfunction
```diff
@@ -2,8 +2,9 @@
...
+execute if entity @s[tag=!gm4_contains_metal] as @e[type=armor_stand,tag=gm4_sand_ring_display,limit=1,sort=nearest,distance=..1.2] run data modify entity @s HandItems[0].tag.CustomModelData set value 26
 tag @s add gm4_contains_metal
-data merge entity @s {ArmorItems:[{},{},{},{id:"minecraft:player_head",Count:1,tag:{CustomModelData:26,SkullOwner:{Id:[I;140191582,-1151712444,-1271433465,-1758951935],Properties:{textures:[{Value:"eyJ0aW1lc3RhbXAiOjE0ODE0MTE2NTQxODksInByb2ZpbGVJZCI6Ijk4NWIyNzVlYmI1YTQzNDRiNDM3Njg5NTI4NjNhNjNmIiwicHJvZmlsZU5hbWUiOiJTcGFya3MiLCJ0ZXh0dXJlcyI6eyJTS0lOIjp7InVybCI6Imh0dHA6Ly90ZXh0dXJlcy5taW5lY3JhZnQubmV0L3RleHR1cmUvODZkMTEwNmNjZDhhY2E3NDg4N2Q1ODIwNTU0YmZiMzM5YzdhM2NmOTJjNDY1MTU2MmM2NmZiNzZkMTQ0In0sIkNBUEUiOnsidXJsIjoiaHR0cDovL3RleHR1cmVzLm1pbmVjcmFmdC5uZXQvdGV4dHVyZS81YzNjYTdlZTJhNDk4ZjFiNWQyNThkNWZhOTI3ZTYzZTQzMzE0M2FkZDU1MzhjZjYzYjZhOWI3OGFlNzM1In19fQ=="}]}}}}]}
```
misode, line 6: Can these commands not be in `initialize` on line 8? It seems like these are the same for each type. I'm unsure why this was done in the past. Is there ever a scenario where `if data storage gm4_metallurgy:temp/item/ore gm4_metallurgy.metal{type:"copper"}` is not true for any of the types?

hozz8528: *shrug*

I can make the changes and test if everything works when I get some time later.

misode: I looked into it and I'd rather leave it like this actually. The way the jumping works is pretty complicated and it's better not to mess anything up.

misode (approved):

sparks0:
After a lot of discussion we've decided not to merge this PR. Adding another entity adds some complexity to a module that's already very complex and confusing and will create extra work if it needs an update in future when there is already a custom texture for the sand ring. The strength of the resource pack is foremost for items to make them look different from the vanilla items, secondly for custom blocks like crafters and I think is least important in situations where a custom texture already exists because of a custom skull. If it had been possible to add this without another entity like in smelteries I'd probably have been in favour of it.

PR #513, opened 2021-01-29
Standardized GM4 color team names

EpyonProjects:
* replaced Team names from gm4_ht_color to gm4_color

In the event future datapacks use in-game Teams for colors, removing the datapack-specific naming scheme allows every datapack to use a GM4-standard team name


misode:
Would there be a problem if every module that needed a team for each color added its own? Because you're going to end up with just as much code in the end.

EpyonProjects:
Each module that needs to create a team will have team(/color) creating code, however this reduces the total number of teams.  If their sole function is to add a color, it seems like a waste to have so many teams.

PR #514, opened 2021-02-01
Repair Relocated Blocks

BluePsychoRanger:
- If moving too quickly, placing down relocated blocks can place a command block in the world

- This fixes the issue by allowing the placed down command block to be powered which will run a repair function


misode (approved):
I've had a difficult experience reproducing the original issue, but the fix seems to make sense when looking over it.

BluePsychoRanger:
> I've had a difficult experience reproducing the original issue, but the fix seems to make sense when looking over it.

Reproducing the original issue can be done by looking at the top part of a stair from the side then placing a relocator. You'll place the relocated block to the side of the stair and the raycast will fail because you're not looking at the relocated block.

misode:
I managed to reproduce and could confirm that this PR fixes the issue.

PR #515, opened 2021-02-05
Health Boost potion effect -> Attribute modifiers

EpyonProjects:
* also had to change the `if score <` to `unless score =` because it would not update sometimes.


misode (approved):
Tested and works fine!

PR #518, opened 2021-02-07
Speed Paths' speed boost: potion effect -> attribute modifier

EpyonProjects:
Making the speed boost an attribute modifier allows the player to stack additional speed effects (vanilla potions or beacons, and Zauber speed armor)


gm4_speed_paths/data/gm4_speed_paths/functions/main.mcfunction
```diff
@@ -1,3 +1,4 @@
-execute as @a at @s if block ~ ~-0.9 ~ grass_path unless entity @s[nbt={ActiveEffects:[{Id:1b,ShowParticles:1b}]}] run effect give @s speed 2 0 true
+execute as @a at @s run attribute @s minecraft:generic.movement_speed modifier remove dc33007e-5da4-4fad-a850-9c5a058c22ba
```
BluePsychoRanger, line 1:
```suggestion
execute as @a run attribute @s minecraft:generic.movement_speed modifier remove dc33007e-5da4-4fad-a850-9c5a058c22ba
```

misode (changes requested):
The current implementation gives the speed effect for 2 seconds for lag and jumping reasons. If attributes are used it needs a similar measurement.

misode:
@EpyonProjects After a few modifications we'd be happy to merge this. 
1. The module didn't have particles before, so it shouldn't artificially be added
2. The duration of the effect is a bit too long, it should be just enough so you keep the effect when jumping on paths with lag. Before the effect was given for 2 seconds and that seemed to work.
2.1. Any reason the scoreboard counts in increments of 5?

SpecialBuilder32 (approved):
Tested and Approved

PR #519, opened 2021-02-08
Health Boost Effect Compatibility

BluePsychoRanger:
This change makes it so that heart canisters only clear the health boost effect if the player doesn't already have the effect, improving compatibility with modules that do use health boost as an effect.


misode:
Doesn't this mean that if a player has a health boost effect, their health will not update?

misode:
Merging this anyways since it is a significant improvement, we can consider further changes later.

PR #521, opened 2021-02-12
Poses pack: add paste and move

Dennis-0:
This update adds 2 new functions:
* `paste` 
A copied armor stand item can be thrown at another armor stand, besides the book with `paste`, to copy that pose onto this armor stand.
* `move` 
Stand inside an armor stand and throw this book at it to move the armor stand to your position

Small addition/change: 
* `flip` works with `body` to rotate the body 180 degrees around its shoulders (this makes it possible for the body to be rotated downwards)
* `body` + `reset` is renamed to `body` + `default` to keep it consistent with the default command for the base module


SpecialBuilder32 (approved):
Tested, reviewed. All looks good to me :)

PR #522, opened 2021-02-12
Use data storage to optimize better armour stands

misode:
This updates the major version of this module to 2, since it will be incompatible with the current expansions. I've only updated the core module for now.


misode:
Decided to do this with the 1.17 update

PR #524, opened 2021-02-15
Metallurgy Update

MichaelMiner137:
- Converted all smooshing related nbt checks to predicates
- Converted all compressing related nbt checks to storage checks


misode:
I'm sorry for your hard work, but the predicates in this PR don't optimize anything, if anything they add extra overhead. Additionally, we don't tend to add people to "Updated By" for a one line change in a module.

PR #531, opened 2021-02-23
Re-worked Levity to fixed 2 1.16 bugs (#529, #533) and #233

### LINKED ISSUE #233: Metallurgy Levity can overwrite or clear an existing levitation effect
This allows players with the levity shamir equipped to completely avoid getting a levitation effect. 

A solution would be to check if the player already has a levitation effect before trying to apply it, and only clear a level 255 levitation effect.

### LINKED ISSUE #529: Metallurgy Levity not working in 1.16+ due to MC-199326 (you cant escape your block)
**Describe the bug**
Due to MC-199326 you cant escape the block your are on when shifting even with levitation 255

**To Reproduce**
1. Try to use levity

**Expected behavior**
To be able to escape the block you are on and start floating away

**Fix**
Make it so you can momentarily un-shift to escape the block

**Environment**
 - Minecraft version: 1.16.5
 - Server version: 1.16.5
 - Installed Datapacks: Metallurgy 
 - Installed Mods: None

**Additional context**
This bug report is just for the record

### LINKED ISSUE #530: Free flight with Levity, no boots no limit (up to 8 mins)
**Describe the bug**
When using levity, you can removed your boots while holding shift (thanks to toggle shit, 1.16+) this will not register as disabling levity. This then allows you to release shift and just continue to float with no cost, you can also go as far as you want. This will only stop when the effects clear

**To Reproduce**
Steps to reproduce the behaviour:
1. Enable toggle shift in Options > Accessibility Options > Sneak: Toggle
2. When standing on a block hit shift once
3. While crouching open your inventory and remove your boots.
4. Now you can un-shift and keep the effects

**Expected behaviour**
Your levitation to be cancelled as if you were just un-shifting with the boots on

**Fix**
Add a check in gm4_metallurgy:player so that if they are still floating, and do not have the boots to cancel them.

**Environment**
 - Minecraft version: 1.16.5
 - Server version: 1.16.5
 - Installed Datapacks: Metallurgy

kruthers:
Fixed #529 by allowing the player to release shift for less then a second when moving off a block (can be fine-tuned and adjusted later)
Fixed #530 by adding a new check in gm4_metallurgy:player so that if they have the floating tag but not the tag for the boot the effects will be cleared correctly.
Fixed #233 (wow this is old) by having a predicate that checks when before it may do any action that effects levitation if they have levities effects of different effects.

I renamed the gm4_levity_started_floating tag to be gm4_levity_is_floating, to make it a bit clearer what it means.


sparks0:
This PR did not pass server testing and needs to be reworked to resolve the issues.

PR #532, opened 2021-02-24
OoA: Spectators + Predicates + Tick

BluePsychoRanger:
- Spectators no longer activate OoA
- predicate checks are used instead of data checks when checking the player's offhand
- functions running every tick were moved to a dedicated function clock

### DIFF
--- a/gm4_orb_of_ankou/data/gm4_orb_of_ankou/functions/main.mcfunction
+++ b/gm4_orb_of_ankou/data/gm4_orb_of_ankou/functions/main.mcfunction
@@ -15,24 +15,18 @@ execute as @e[type=armor_stand,tag=gm4_soul_forge] at @s if predicate gm4_orb_of
 
 ## PNEUMA STUFF ##
 
-# bubbly
-execute as @a[gamemode=!spectator,tag=gm4_pneuma_bubbly] at @s anchored eyes unless block ^ ^ ^ #gm4:water unless block ^ ^ ^ #gm4:waterloggable[waterlogged=true] run effect give @s water_breathing 121 0
-
 # gliding
 effect give @a[gamemode=!spectator,tag=gm4_pneuma_gliding,scores={gm4_oa_swim=1..}] dolphins_grace 3 1 true
 scoreboard players reset @a gm4_oa_swim
 
-# synergetic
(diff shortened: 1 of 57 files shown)


gm4_orb_of_ankou/data/gm4_orb_of_ankou/functions/main.mcfunction
```diff
@@ -16,33 +16,27 @@ execute as @e[type=armor_stand,tag=gm4_soul_forge] at @s if predicate gm4_orb_of
...
-execute as @a[tag=gm4_oa_soaring_active] at @s run function gm4_orb_of_ankou:pneumas/soaring/apply
+effect give @a[gamemode=!spectator,tag=gm4_oa_soaring_off_ground] jump_boost 2 255 true
+execute as @a[gamemode=!spectator,tag=gm4_oa_soaring_active] at @s run function gm4_orb_of_ankou:pneumas/soaring/apply
 
 # sneaking stuff
-execute as @a[scores={gm4_oa_sneak=0},tag=gm4_oa_sneaking] run function gm4_orb_of_ankou:pneumas/sneak/stopped
-execute as @a[scores={gm4_oa_sneak=1..},tag=gm4_has_pneuma] run function gm4_orb_of_ankou:pneumas/sneak/check
+execute as @a[gamemode=!spectator,scores={gm4_oa_sneak=0},tag=gm4_oa_sneaking] run function gm4_orb_of_ankou:pneumas/sneak/stopped
+execute as @a[gamemode=!spectator,scores={gm4_oa_sneak=1..},tag=gm4_has_pneuma] run function gm4_orb_of_ankou:pneumas/sneak/check
 scoreboard players set @a gm4_oa_sneak 0
```
Bloo-dev, line 36: Wouldn't it be worth putting these into a function that's called @a and then replacing all these @a's with @s?

BluePsychoRanger: In most cases, yes. Where it's `execute as @a`, those should be in a separate function, but the ones without `execute` should remain in main because it'd run less commands per player. I'll fix it.

sparks0:
Once this becomes live on the server, we'll add this PR version for testing.

BluePsychoRanger:
This PR was merged with #639

PR #533, opened 2021-02-25
Added tnt projectiles

TimeIntensive:
Added

-tnt projectile to crossbow cartages
-firing arrow with tnt in offhand will summon primed tnt at location of arrows landing
-added tnt priming sound when firing cartridge 
-added explosion effect when firing cartridge

Tips

-If used with Piercing enchantment it is possible to use on entities


gm4_crossbow_cartridges/data/gm4_crossbow_cartridges/functions/tnt/tnt_loop.mcfunction
```diff
@@ -0,0 +1,3 @@
+#check for arrow that places a tnt or crop
+execute as @e[type=minecraft:arrow,tag=gm4_cb_tnt,nbt={inGround:1b}] at @s run function gm4_crossbow_cartridges:tnt/tnt_check
+execute if entity @e[type=minecraft:arrow,tag=gm4_cb_tnt] run schedule function gm4_crossbow_cartridges:tnt/tnt_loop 16t
```
BluePsychoRanger, line 3:
```suggestion
execute if entity @e[type=minecraft:arrow,tag=gm4_cb_tnt,limit=1] run schedule function gm4_crossbow_cartridges:tnt/tnt_loop 16t
```

gm4_crossbow_cartridges/data/gm4_crossbow_cartridges/functions/projectile/tnt.mcfunction
```diff
@@ -0,0 +1,20 @@
+#run from check_projectile
+# @s = player using the crossbow
+
+#add tag to arrow
+tag @e[type=minecraft:arrow,distance=..2,limit=1,sort=nearest] add gm4_cb_tnt
+
+#add fire to arrow
+data merge entity @e[type=minecraft:arrow,distance=..2,limit=1,sort=nearest] {Fire:2000s}
```
BluePsychoRanger, line 8:
```suggestion
data merge entity @e[type=minecraft:arrow,tag=gm4_cb_tnt,distance=..2,limit=1,sort=nearest] {Fire:2000s}
```

gm4_crossbow_cartridges/data/gm4_crossbow_cartridges/functions/projectile/tnt.mcfunction
```diff
@@ -0,0 +1,20 @@
...
+
+#take item from player
+clear @s[gamemode=!creative] tnt 1
```
BluePsychoRanger, line 11:
```suggestion
clear @s[gamemode=!creative,gamemode=!spectator] tnt 1
```

BluePsychoRanger:
I think this module would make more sense if the tnt was primed (summon it with `Fuse:80s`) and use the same implementation as the fish. Doesn't really make sense for the arrow to wait until hitting the ground for the tnt to spawn. This would also be more visually appealing, as players can actually see the tnt flying

EpyonProjects:
> 
> 
> I think this module would make more sense if the tnt was primed (summon it with `Fuse:80s`) and use the same implementation as the fish. Doesn't really make sense for the arrow to wait until hitting the ground for the tnt to spawn. This would also be more visually appealing, as players can actually see the tnt flying

I suggested it just summon upon impact.  Due to the current implementation, entities hit do not activate the tnt, but it would allow for a range of distances.   What about a primed tnt projectile, but only visually, so it would have the flying tnt but keep the explosion-on-impact mechanic?

Bloo-dev:
We've checked this out during our dev day and have discovered two things:
- Arrows that are stuck in the ground next to you will take priority over the arrow shot by the crossbow. This means that standing next to old arrows will make you explode. However, this is an issue which persists throughout crossbow cartridges and requires a module-wide fix using `Age:0`
- Entities hit by the projectile are not exploded, which is somewhat sad. This should be implemented using an AEC to cache the arrows location, and tnt should be summoned once the arrow is either gone or `inGround`
- we've removed the `explosion` particle as it was barely visible and didn't make too much sense to us

BluePsychoRanger:
> Entities hit by the projectile are not exploded, which is somewhat sad. This should be implemented using an AEC to cache the arrows location, and tnt should be summoned once the arrow is either gone or inGround

I suggested simply having the tnt thrown with the motion the arrow originally had (possibly with a longer fuse), which would mirror every other crossbow cartridges projectile (except string and torch, but those are special cases). IMO this shouldn't be "TNT Arrows" but rather "TNT Launching"

Bloo-dev:
Closing this PR as we have not heared from you in 7 days.

PR #538, opened 2021-03-13
Add automatic data pack linter + style fixes

misode:
This PR adds a Github workflow that validates master and pull requests.
It uses https://github.com/ChenCMD/datapack-linter which internally uses the Data-Pack Helper Plus VSCode extension https://github.com/SPYGlassMC/SPYGlass.

An example of the feedback it gives: https://github.com/misode/GM4_Clone3/runs/2104255623
image

This fixes some bugs, including but not limited to:
- falling stars did not give you levitation
- nightmare nights did not kick players out of their bed
- metallurgy vex was not invulnerable
- soul glass destroyed item did not have a correct pickup delay 
- orb of ankou fire sprite still has old attribute names
- failed extraction still used UUIDMost for randomization
- disassemblers destroyed item did not apply lore correctly to crafting table


Bloo-dev (approved):
Checked the modified files and all modifications seem valid.
It's probably out of the scope of this PR, but it would be nice to use something like this to enforce header comment and comment style.

PR #539, opened 2021-03-16
Better Armour Stands: Add head/hand interaction

Dennis-0:
Throw a book with one of the following commands and an item at and armour stand to place this item in this slot, if it's currently empty
* equip head
* equip hand
* equip offhand


SpecialBuilder32 (approved):
Tested and Approved for Merge

PR #540, opened 2021-03-16
Crossbow Cartridges: 1.16 update 2

Dennis-0:
* add soul/redstone torch
* update block tag to use #gm4:full_collision


gm4_crossbow_cartridges/data/gm4_crossbow_cartridges/functions/check_projectile.mcfunction
```diff
@@ -17,6 +17,8 @@ execute if entity @s[predicate=gm4_crossbow_cartridges:lingering_potion] run fun
 
 #torch
 execute if entity @s[gamemode=!adventure,predicate=gm4_crossbow_cartridges:torch] run function gm4_crossbow_cartridges:projectile/torch
+execute if entity @s[gamemode=!adventure,predicate=gm4_crossbow_cartridges:redstone_torch] run function gm4_crossbow_cartridges:projectile/redstone_torch
+execute if entity @s[gamemode=!adventure,predicate=gm4_crossbow_cartridges:soul_torch] run function gm4_crossbow_cartridges:projectile/soul_torch
 
 #fireball
 execute if entity @s[gamemode=!adventure,predicate=gm4_crossbow_cartridges:fire_charge] run function gm4_crossbow_cartridges:projectile/fireball
```
Bloo-dev, line 24: I think it might be a good idea to implement a fake player alongside with a score check here, as only one of these condition can evaluate to `true` in a given call.
To do this, set a fake player to `1` in the `projectile/` functions and check that score before executing any of the `if entity` calls. Those should only be run if the score is still `0`.

Bloo-dev:
Closing this PR as we have not heard from you for a while.

PR #541, opened 2021-03-16
Metallurgy: Defuse to work on armour stands

Dennis-0:
Each armour stand will be able to defuse creepers just like players, if it has a defuse shamir equiped

This idea was accepted in the labs discord 
https://discord.com/channels/515225138006851624/515578870540533771/652231317148401674


Bloo-dev:
This is not something we desire to implement as both of the following outcomes are not ideal:
-  Scenario 1: This is used by players a lot, resulting in very poor performance as the number of armor stands is usually significantly higher than the amount of players.
- Scenario 2: The feature is barely used. However, it will still cause a lot of lag as every armor stand needs to be checked.
Hence I'll close this PR.

PR #543, opened 2021-03-22
Vertical Rails - Optimizations (Main, clock speed)

EpyonProjects:
* main function put into advancement
* Clock speed reduced from 1t to 4t

### DIFF
--- a/gm4_vertical_rails/data/gm4/advancements/vertical_rails.json
+++ b/gm4_vertical_rails/data/gm4/advancements/vertical_rails.json
@@ -23,7 +23,25 @@
   "parent": "gm4:root",
   "criteria": {
     "ride_vertical_rail": {
-      "trigger": "minecraft:impossible"
+      "trigger": "minecraft:location",
+      "conditions": {
+        "player": [
+          {
+            "condition": "minecraft:entity_properties",
+            "entity": "this",
+            "predicate": {
+              "location": {
+                "block": {
+                  "block": "minecraft:ladder"
+                }
+              },
+              "vehicle": {
+                "type": "minecraft:minecart"
+              }
+            }
+          }
+        ]
+      }
     }
   }
 }
--- a/gm4_vertical_rails/data/gm4_vertical_rails/functions/init.mcfunction
+++ b/gm4_vertical_rails/data/gm4_vertical_rails/functions/init.mcfunction
@@ -1,7 +1,6 @@
 execute unless score vertical_rails gm4_modules matches 1 run data modify storage gm4:log queue append value {type:"install",module:"Vertical Rails"}
 scoreboard players set vertical_rails gm4_modules 1
 
-schedule function gm4_vertical_rails:main 1t
 schedule function gm4_vertical_rails:tick 1t
 
 #$moduleUpdateList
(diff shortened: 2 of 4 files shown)


Bloo-dev:
We might want to increase the vertical motion as the clock runs slower (To be balanced).

RitikShah:
One issue I find with the minecart-related modules is that minecart chests spawn naturally and they cause unnecessary lag to these modules (they are usually the first thing I have to disable in my worlds). I wonder if it might be worthwhile to explicitly mark those with loot tables and remove them from the `@e`?

Bloo-dev:
> One issue I find with the minecart-related modules is that ~minecart chests~ spawn naturally and they cause unnecessary lag to these modules (they are usually the first thing I have to disable in my worlds). I wonder if it might be worthwhile to explicitly mark those with vanilla loot tables and remove them from the `@e`'s?
> 
> Oh, I see that only minecarts are detected here, I often see an entity tag of minecarts in my debug files so I wonder where that's from.

Sadly, it'd be most likely more expensive to check that a minecart is a naturally generated minecart instead of simply having the module run on those carts and then abort when there is no ladder nearby :/

Bloo-dev:
I have noticed some weirdness about vertical rails on the current public server; When you reach the top of a vertical rail the cart sometimes doesn't make it onto the horizontal rail and you end up doing short hops until you're back on the horizontal rail.

RitikShah:
> > One issue I find with the minecart-related modules is that ~minecart chests~ spawn naturally and they cause unnecessary lag to these modules (they are usually the first thing I have to disable in my worlds). I wonder if it might be worthwhile to explicitly mark those with vanilla loot tables and remove them from the `@e`'s?
> > Oh, I see that only minecarts are detected here, I often see an entity tag of minecarts in my debug files so I wonder where that's from.
> 
> Sadly, it'd be most likely more expensive to check that a minecart is a naturally generated minecart instead of simply having the module run on those carts and then abort when there is no ladder nearby :/

Not necessarily, a check for naturally generated minecart (which are only chest minecarts) would be 1 time, and u can tag the minecart and remove it from the `@e` check. The issue with aborting when theres' no ladder nearby is that it is a ticking operation while just checking for a loot table would be one time. You can even add an `interacted_with_entity` advancement trigger and remove the tag when someone opens the chest.

Bloo-dev:
I will leave this PR unmerged for a bit, as we have discovered some weirdness with short vertical rails. Epyon tweaked some numbers, so we'll see how that looks.

Bloo-dev (changes requested):
Having the main clock of this module run at `4t` causes minecarts to behave weirdly at the top of vertical rails.
We've switched to `2t` on the current Public Server which reduces this weirdness, however, it does not fully represent the old `1t` behavior.

Bloo-dev:
We've talked about this and we think the best solution would be the following:
- Have a function tick at `8t` (this value will need to be adjusted depending on how things feel) that checks for ladders at a minecart.
- If a ladder is found, a 20Hz is started that keeps rescheduling itself until there is no more minecart at a ladder. A fake player should be used in `momentum.mcfunction` to remember that there is at least one minecart still on a ladder; This will make sure that the block check for ladders is only done once. If the fake player is set, the 20Hz re-schedules itself.

Bloo-dev:
Closed since the issues above have not been fixed.

PR #545, opened 2021-04-02
Give Brewer and Congealed Enchantments' name a section symbol at the end

hozz8528:
EDIT: Now puts the character at the end, so it simply acts as a dummy invisible character that the player can't input for Optifine to detect. 10000000000% better.

OLD: Makes it so that Resource Packs can use Optifine's Random Entities to detect the white color format on top of the name when retexturing the entities. This makes it so entities renamed by the player won't change their texture as well when using the RP. This method could become a way to allow for some mobs to be retextured with Resource Packs (with Optifine).


hozz8528:
Since it relies on the symbol. I just placed the symbol at the end instead. Now it's just a dummy symbol and won't change the format at all.

sparks0 (approved):

PR #546, opened 2021-04-02
Added GM4 Intro Song to base

EpyonProjects:
* Plays upon first movement after joining, or
* Plays upon installation for creative players / first join.

### DIFF
--- a/base/data/gm4-1.1/functions/load.mcfunction
+++ b/base/data/gm4-1.1/functions/load.mcfunction
@@ -14,5 +14,13 @@ execute unless score $cooldown gm4_creative matches 0.. run scoreboard players s
 # Setting this to 0 will always show logs to all players
 execute unless score $global_cooldown gm4_creative matches 0.. run scoreboard players set $global_cooldown gm4_creative 10
 
+# Intro song
+scoreboard objectives add gm4_intro_song dummy
+scoreboard objectives add gm4_intro_song_t dummy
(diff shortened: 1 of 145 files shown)


base/data/gm4-1.1/functions/load.mcfunction
```diff
@@ -14,5 +14,13 @@ execute unless score $cooldown gm4_creative matches 0.. run scoreboard players s
 # Setting this to 0 will always show logs to all players
 execute unless score $global_cooldown gm4_creative matches 0.. run scoreboard players set $global_cooldown gm4_creative 10
 
+# Intro song
+scoreboard objectives add gm4_intro_song dummy
+scoreboard objectives add gm4_intro_song_t dummy
+
+scoreboard objectives add gm4_intro_played dummy
```
SpecialBuilder32, line 21:
```suggestion
```

SpecialBuilder32 (approved):
I love it. There are some oddities with the function tree layout as a product of the noteblock generator, but since it only runs once I think they're no problem.

Bloo-dev (approved):
I've looked through this before and it looks fine, I just have two minor concerns:

1. This uses two objectives for the song position (`gm4_intro_song` and `gm4_intro_song_t`) and part of me is certain it can be done with a single objective
2. I'm not sure about the walking check, how much of a difference does it make?

PR #547, opened 2021-04-02
Updated Heart Cans' skull textures

EpyonProjects:


misode:
The `load` score and the `gm4_modules` score should be left untouched as this is not a breaking change

Bloo-dev:
Why does this PR remove the part of the Datapack that repairs placed down skulls?

PR #548, opened 2021-04-02
Update Enderpuff texture

hozz8528:
See title. I didn't change the (what I'm guessing is the) version number in init and load like Epyon did for Heart Canisters as I didn't know if it was considered necessary in this case. Let me know if I have to and I'll change that.

### DIFF
--- a/gm4_end_fishing/data/gm4_end_fishing/functions/main.mcfunction
+++ b/gm4_end_fishing/data/gm4_end_fishing/functions/main.mcfunction
@@ -6,6 +6,7 @@ execute as @a[tag=gm4_has_enderpuff] run function gm4_end_fishing:enderpuff/proc
 scoreboard players reset @a gm4_ef_sneak
 
 #fix enderpuff head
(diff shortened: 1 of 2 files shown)


gm4_end_fishing/data/gm4_end_fishing/functions/main.mcfunction
```diff
@@ -6,6 +6,6 @@ execute as @a[tag=gm4_has_enderpuff] run function gm4_end_fishing:enderpuff/proc
 scoreboard players reset @a gm4_ef_sneak
 
 #fix enderpuff head
-execute as @e[type=item,nbt={Item:{id:"minecraft:player_head",Count:1b,tag:{SkullOwner:{Id:[I;2008953110,-1021689564,-2086641039,1315516205]}}}}] run data merge entity @s {Item:{tag:{CustomModelData:14,gm4_end_fishing:{enderpuff:1b},display:{Name:'{"translate":"%1$s%3427655$s","with":["Enderpuff",{"translate":"item.gm4.enderpuff"}],"italic":false}'}}}}
+execute as @e[type=item,nbt={Item:{id:"minecraft:player_head",Count:1b,tag:{SkullOwner:{Id:[I;-749248143,1561284060,-1964497486,1336566238]}}}}] run data merge entity @s {Item:{tag:{CustomModelData:14,gm4_end_fishing:{enderpuff:1b},display:{Name:'{"translate":"%1$s%3427655$s","with":["Enderpuff",{"translate":"item.gm4.enderpuff"}],"italic":false}'}}}}
```
Bloo-dev, line 9: This should be moved into a function as we need to keep the old repair command for a couple of versions. I'd be best to pull the `Id` into storage and then run the appropriate command based on that.

Bloo-dev (changes requested):

PR #554, opened 2021-04-12
Ziprails - Optimizations (2t clock, remove main, reduce @e calls)

EpyonProjects:
* Advancement granted using Vehicle nbt
* Tick branches on @e minecart once, instead of 2 @e's per tick
* Tick is now main, and schedules on a 2t clock.  Speeds adjusted accordingly


gm4_ziprails/data/gm4/advancements/ziprails.json
```diff
@@ -8,22 +8,43 @@
...
+          "condition": "minecraft:entity_properties",
+          "entity": "this",
+          "predicate": {
+            "vehicle": {
+              "type": "minecraft:minecart",
+              "nbt": "{Tags:[\"gm4_linked\"]}"
+            },
+            "player": {
+              "advancements": {
+                "gm4:ziprails": false
```
misode, line 42: Adding this check is useless and hurts performance. When the player already has this advancement it won't trigger it again and even start to check conditions. For players that don't have the advancement it's extra work that always gives the same result. (Also side note please avoid formatting the document using `[{`, line 32).

gm4_ziprails/data/gm4_ziprails/functions/checks/cable.mcfunction
```diff
@@ -0,0 +1,2 @@
+execute if block ~ ~1 ~ #gm4_ziprails:cable[attached=true] run function gm4_ziprails:ziprail/zipping
```
misode, line 1: Please add a heading comment for this function. Also I find it a bit strange to create a folder for this one function. Especially since "checks" is not very descriptive. Most functions check something and then do something with the result.

misode (commented):
I like the optimizations so far. While it's good the main function is removed, the main lag causer is the tick function. Ideally there would be a slower clock detecting when a minecart needs to be linked and then a faster clock can be scheduled.

Bloo-dev:
This PR looks great!
We might to increase motion as the clock speed changed (to be balanced).

PR #556, opened 2021-04-13
Added saddle crafting recipe to standard crafting

EpyonProjects:


sparks0:
We've excluded saddles from standard crafting for many years because we wanted to encourage exploration (and I guess fishing now?). Did I miss a conversation because this is still how I feel about crafting saddles.

PR #559, opened 2021-04-18
Weighted Armour: potion effect -> attribute

EpyonProjects:
* Reduces player movement speed in an attribute
* Allows slowness to stack with slowness potion effect


Bloo-dev:
After extensive server testing we've come to the conclusion that using an attribute (and thereby hiding the amount of slow-down they experience) is very unintuitive and makes the weight system really frustrating to handle.

PR #563, opened 2021-04-24
Revise boots of ostara functionality

Luewd:
* As discussed in the Gamemode 4 Discord, the Boots of Ostara's ability to place crops was never really useful as the module ticks too slowly to be better than placing by hand, and pig tractors exists for automated usage. Hence, that functionality was removed. The ability for players and armor stands to spread grass blocks remains.

 * The custom crafting recipe for Boots of Ostara now preserves the NBT of the leather boots used in the recipe, such as `CustomModelData`, enchantments, and shamirs. The `Lore` tag is prepended to instead of overriden, as well, to preserve existing lore when present.

 * Instead of checking whether the Boots of Ostara are equipped separately from the player holding a grass block, there is now a predicate that checks for both at once, since holding a grass block while wearing Boots of Ostara is the only remaining feature other than `block_desire_lines` (which isn't part of the `main` function).


Bloo-dev (approved):
A very clean implementation. We might end up implementing a `lib_lore` to manage lore in the future, however, that can still be changed later.

Bloo-dev (approved):
Tested and reviewed.

PR #564, opened 2021-04-24
Desire Lines: Add Celaro Shamir

Luewd:
Celaro is a new type of Aluminium Band that can be placed on boots, included in the Desire Lines module when loaded alongside Metallurgy. When applied to the player's boots, the Celaro shamir will simply suppress the the effects of Desire Lines, just like the Boots of Ostara. There is also a fun easter egg suggested by Bloo if you stand in tall flowers while wearing Celaro boots.

### DIFF
--- a/gm4_desire_lines/data/gm4/tags/items/boots.json
+++ b/gm4_desire_lines/data/gm4/tags/items/boots.json
@@ -1,0 +1,10 @@
+{
+  "values": [
+    "minecraft:netherite_boots",
+    "minecraft:diamond_boots",
+    "minecraft:golden_boots",
+    "minecraft:chainmail_boots",
+    "minecraft:iron_boots",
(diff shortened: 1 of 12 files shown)


Bloo-dev (commented):
Shamir names usually follow a latinesque naming scheme, so we may want to reconsider the name. Provided we find a good onr that doesn't feel forced.

ToffeeMax:
> Shamir names usually follow a latinesque naming scheme, so we may want to reconsider the name. Provided we find a good onr that doesn't feel forced.

If a suggestion is wanted -- a play on "invisible" comes to mind -- which works as Caecus

Luewd:
The shamir is now named Celaro instead of Delicate as suggested in #module-discussion on the discord.

Bloo-dev (approved):
Tested and reviewed.

PR #565, opened 2021-04-25
Update Zauber Armour creation

hozz8528:
Updates the way zauber armour pieces are created so that the items keep any tags, including the CMD, that they may have had before zaubering. Also placed the easter egg lore before any pre existing lore to avoid the 25% Magic Shamirs bug (also looks nicer when it's before lore IMO).


hozz8528:
I'm closing this for now as this is closely related to the way metallurgy applies bands and I think I want to discuss the lore order with a few others, so then I can make a new PR that takes into account both modules

PR #567, opened 2021-04-25
Update How shamirs add and remove lore

hozz8528:
Made it so shamirs prepend their lore to existing lore when added, updated removing shamirs so that it copies the lore to storage, finds the "Shamir" line and removes that and the surrounding lines before applying it back to the item, and also updated the way the finalized shamir item gets the lore from the item so that it gets the type of shamir in a similar way to how I remove specific lore.

Basically, adding and removing shamirs won't mess up your lore now.


BluePsychoRanger:
#576 does this in a less specific way, utilizing a new library dealing with lore

hozz8528:
closing since lib_lore makes this unnecessary

PR #576, opened 2021-05-04
Add lib_lore

BluePsychoRanger:
lib_lore allows for easy searching, removal, inserting, and replacing of lore lines. 

- [x] remove
- [x] search
- [x] insert
- [x] replace

--

REMOVE - removes a specified line of lore
To set up the removal of lore, the following steps should be taken:

1. Set the Source. This should be a list
`data modify storage gm4_lore:temp Source set from <...>`
2. Set the Target. This should be a single json item.
`data modify storage gm4_lore:temp Target set from <...>`
3. (optional) set the start/offset. This determines where the removal will start when it finds the target line; can be negative for before the line or positive for after the line. Defaults to 0.
`scoreboard players set $start gm4_lore <value>`
4. (optional) set the extra lines. This determines how many more lines to remove after the first target line is removed. Must be positive. Defaults to 0. If set to -1, every line after the first target line will be removed.
`scoreboard players set $extra gm4_lore <value>`
5. run the removal command
`function #gm4_lore:remove`

This will give you 3 things to work with: the Source, the Dump, and the index.
The Source is the modified lore with the removed lines, and the Dump is the removed lines. The index is the location of the first removed line
`data get storage gm4_lore:temp Source`
`data get storage gm4_lore:temp Dump`
`scoreboard players get $index gm4_lore`

--

SEARCH - gets the index of a specified line of lore
To set up the searching of lore, the following steps should be taken:

1. Set the Source and Target as before (1 and 2 from REMOVE)
2. (optional) set the start/offset. This determines where the search will start; must be a positive number. Defaults to 0.
3. run the search command
`function #gm4_lore:search`

Source will be unchanged, and Dump will be the line that was found by the search function (so the Target). The index is the location of the line that was found.

--

INSERT - inserts lines of lore after a specified line of lore
To set up the insertion of lore, the following steps should be taken:

1. Set the Source and Target as before (1 and 2 from REMOVE)
2. Set the Input. This should be a list
`data modify storage gm4_lore:temp Input set from <...>`
4. (optional) set the start/offset. This determines where the insertion will start when it finds the target line; can be negative for before the line or positive for after the line. Defaults to 0.
5. run the search command
`function #gm4_lore:insert`

Source will be have the inserted lore, and Dump will be the lines that were added. The index is the location of the first line that was added.

--

REPLACE - replaces a specified line of lore with 1 or more new lines of lore
To set up the replacement of lore, the following steps should be taken:

1. Set the Source, Target, and Input as before (1 and 2 from INSERT)
2. (optional) set the start/offset. This determines where the search will start; must be a positive number. Defaults to 0.
3. run the search command
`function #gm4_lore:replace`

Source will be have the new replaced lore, and Dump will be the line that was replaced. The index is the location of the first line that was replaced.


misode (changes requested):
The way versioning is done makes it sort of useless. The `gm4_lore:remove` function is always going to be overwritten by the last data pack defined by data pack load order. While the `resolve_load` functions makes it so only the `load` of the most recent version runs. 

The way to fix it would be to move the `insert`, `remove`, etc functions to the versioned namespace.  If you want dependencies to call a non versioned function you can make that a function tag, which will then check the `gm4_lore load` score similar to how `load` is resolved. This for each of the 4 calls.

Additionally, `gm4_metallurgy` should check for `gm4_lore` in its `load` function.

Forgot to say: I was a bit skeptical of this approach at first, but now I'm really confident this is the best way to solve these issues, so good job!

BluePsychoRanger:
Thanks Misode, I've made those changes

hozz8528:
I've tested the following:

Adding/removing shamirs to/from Unbreaking V items with Lore from Sky Dimension before and after making it Zauber.
Adding/removing shamirs to/from leather boots before and after making them into Boots of Ostara.
Manually removing lore lines from Orbs of Ankou using soul essences to determine the lines to remove.

Haven't found any issues so far.

PR #577, opened 2021-05-04
Translation path fixes (mostly)

hozz8528:
- Changed a few translation paths to be more in line with the rest of modules (changed an item. to text., added the name of the module to the path in some cases, etc.)
- Added a few missing translation paths (and fixed 1 broken one)
- Added section symbols (and 1 CMD) to Orb of Ankou mobs to allow for retexturing since I was already modifying those files anyways.


BluePsychoRanger (approved):
Looks good, though have you looked through every module for incorrect translation keys?

gm4_zauber_liquids/data/gm4_zauber_liquids/functions/liquid_init/harming.mcfunction
```diff
@@ -1,4 +1,4 @@
-data merge block ~ ~ ~ {CustomName:'"Zauber Harming Potion Tank"'}
+data merge block ~ ~ ~ {CustomName:'{"translate":"%1$s%3427655$s","with":["Zauber Harming Potion Tank", {"translate":"block.gm4.zauber_liquids.harming"}]}'}
```
misode, line 1: To be consistent with potion tanks, the translation key should be `block.gm4.zauber_tank.harming`. See https://github.com/Gamemode4Dev/GM4_Resources/blob/master/gm4_translations/assets/gm4/lang/en_us.json#L111

gm4_everstone/data/gm4_everstone/loot_tables/everstone.json
```diff
@@ -17,12 +17,16 @@
...
                           "function": "set_name",
-                          "name": [
-                              {
-                                  "text": "Everstone",
-                                  "italic": false
-                              }
-                          ]
+                          "name": {
+							"translate": "%1$s%3427655$s",
+							"with": [
```
misode, line 22: Mixed tabs and spaces. I know we aren't always consistent, but at least consistency on a file base would be nice.

misode (changes requested):

misode (approved):

PR #578, opened 2021-05-04
Tnt landmines fix

hozz8528:
For some reason, The recipe for Soft Defuser wouldn't detect the potions of harming IV until I removed `minecraft:` and the `b` after the numbers. I also fixed the lore path to be the same as the other landmines


misode:
I'd actually prefer to remove the changed recipe when Zauber Cauldrons is installed. I don't think it adds much other than possible confusion.

hozz8528:
> 
> 
> I'd actually prefer to remove the changed recipe when Zauber Cauldrons is installed. I don't think it adds much other than possible confusion.

Done. We might as well rethink the recipe a bit so it doesn't use another landmine. It always made very little sense to me, and BPR and Bloo agree

Bloo-dev:
> I'd actually prefer to remove the changed recipe when Zauber Cauldrons is installed. I don't think it adds much other than possible confusion.

I'm ok with this change.

BluePsychoRanger (approved):

misode (approved):

PR #579, opened 2021-05-04
Move tower_structures and dangerous_dungeons files into their own modules

MichaelMiner137:
Update:
- Moves all tower_structures and dangerous_dungeons files out of orbis and puts them in their own modules

Notes:
- I was able to get all the towers and dungeons to spawn including the chest loot
- The wiki says mob loot is disabled and the loot tables are empty however this and the master branch version have mob loot


misode:
Thanks for your contribution, but these are not changes that we want to make. The tower and dungeon specific functions in the Orbis base are to allow other expansions that use these and wouldn't have to start from scratch with the chunk generate tag.

Additionally, moving and renaming loot tables would break existing placed spawners and chests.

PR #580, opened 2021-05-05
Villager Telltale Module

MichaelMiner137:
Features
- When holding a work block near a villager they will look at their job site and particles will be displayed
- The same effect will happen when holding a bell or bed
- If the site of interest is missing then a barrier particle will be displayed

Notes
- If the villagers bell or bed is broken the nbt will not change so they will look at the last known location


gm4_villager_telltale/data/gm4/advancements/vt_find_workstation.json
```diff
@@ -0,0 +1,29 @@
+{
```
BluePsychoRanger, line 1: This advancement should be named more consistently with the rest of GM4 modules. Modules with only a single advancement should be named their own module name. (So this should be renamed to `villager_telltale`)

BluePsychoRanger (changes requested):
This looks like a good module. However there are some optimizations to be made, especially utilizing predicates for the mainhand check. What happens if the villager doesn't have a bed/workstation/bell? Do they still look somewhere?

gm4_villager_telltale/data/gm4_villager_telltale/functions/main.mcfunction
```diff
@@ -0,0 +1,6 @@
+
+schedule function gm4_villager_telltale:main 3s
+
+execute as @e[type=villager,tag=gm4_vt_looking] run function gm4_villager_telltale:villager/stop_looking
+
+execute as @a at @s run function gm4_villager_telltale:player
```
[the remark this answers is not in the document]
BluePsychoRanger: Correction, you need 3 item tags cuz you don't need an item tag for just the bell. The "all" item tag will just use the other two item tags, plus the bell.

gm4_villager_telltale/data/gm4_villager_telltale/functions/villager/locate_bed.mcfunction
```diff
@@ -0,0 +1,14 @@
...
+execute unless data storage gm4_villager_telltale:temp/villager NBT.Brain.memories."minecraft:home" run data remove storage gm4_villager_telltale:temp/villager Look
+execute store result storage gm4_villager_telltale:temp/villager Look[0] double 1 run data get storage gm4_villager_telltale:temp/villager NBT.Brain.memories."minecraft:home".value.pos[0]
+execute store result storage gm4_villager_telltale:temp/villager Look[1] double 1 run data get storage gm4_villager_telltale:temp/villager NBT.Brain.memories."minecraft:home".value.pos[1]
+execute store result storage gm4_villager_telltale:temp/villager Look[2] double 1 run data get storage gm4_villager_telltale:temp/villager NBT.Brain.memories."minecraft:home".value.pos[2]
+
+# show target
+execute if data storage gm4_villager_telltale:temp/villager Look[2] run function gm4_villager_telltale:villager/show_target
+
+# look at target
+execute if data storage gm4_villager_telltale:temp/villager Look[2] run function gm4_villager_telltale:villager/begin_looking
```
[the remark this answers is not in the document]
MichaelMiner137: yes, it is checking if their is a job/bed/bell

gm4_villager_telltale/data/gm4_villager_telltale/functions/main.mcfunction
```diff
@@ -0,0 +1,7 @@
+
+schedule function gm4_villager_telltale:main 3s
+
+execute as @e[type=villager,tag=gm4_vt_looking] run function gm4_villager_telltale:villager/stop_looking
+
+tag @a remove gm4_tv_holding_villager_block
```
BluePsychoRanger, line 6: This tag should be removed at the end of the `player` function. There's no reason to run it every clock cycle.

BluePsychoRanger: Also why is it `tv` instead of `vt`?

PR #581, opened 2021-05-06
Metallurgy Arborenda: Disable by sneaking

Dennis-0:


misode (approved):
Works fine and the change seems useful 👍

SpecialBuilder32 (approved):
Nice clean simple change. Hozz tested and it worked as expected

PR #582, opened 2021-05-06
Rational Villagers Module

MichaelMiner137:
Features
- Limits the major_positive gossip to 5 (normally 100)
- Limits the minor_positive gossip to 75 (normally 200)


Bloo-dev:
Whilst this fix is (in my opinion) a good idea, it doesn't fit Gamemode 4. Our modules usually add Gameplay elements and are not "bugfix" or "utility" modules.

PR #583, opened 2021-05-06
Make band recycling not give you the same band again

hozz8528:
Makes the vex mould store the CMD of the recycled band as a score, then kills the band with the same CMD when they are summoned before selecting one.

### DIFF
--- a/gm4_metallurgy/data/gm4_metallurgy/functions/casting/finish_band.mcfunction
+++ b/gm4_metallurgy/data/gm4_metallurgy/functions/casting/finish_band.mcfunction
@@ -1,6 +1,10 @@
 # @s = mould requesting a band
 # run from any function in casting/summon_band
 
+# kill band with the same CMD as the CMD of possible recycled band
+execute as @e[type=item,distance=..0.1,nbt={Age:0s,Item:{tag:{gm4_metallurgy:{has_shamir:1b}}}}] run execute store result score @s gm4_ml_sh_id run data get entity @s Item.tag.CustomModelData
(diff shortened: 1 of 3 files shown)


gm4_metallurgy/data/gm4_metallurgy/functions/casting/finish_band.mcfunction
```diff
@@ -1,6 +1,10 @@
 # @s = mould requesting a band
 # run from any function in casting/summon_band
 
+# kill band with the same CMD as the CMD of possible recycled band
+execute as @e[type=item,distance=..0.1,nbt={Age:0s,Item:{tag:{gm4_metallurgy:{has_shamir:1b}}}}] run execute store result score @s gm4_ml_sh_id run data get entity @s Item.tag.CustomModelData
+execute as @e[type=item,distance=..0.1,nbt={Age:0s,Item:{tag:{gm4_metallurgy:{has_shamir:1b}}}}] if score @s gm4_ml_sh_id = @e[type=vex,tag=gm4_sand_ring,distance=..0.1,limit=1] gm4_ml_sh_id run kill @s
```
Bloo-dev, line 6: This double invocation of `@e[type=item,distance=..0.1,nbt={Age:0s,Item:{tag:{gm4_metallurgy:{has_shamir:1b}}}}]` should be avoided by the use of a function.

Bloo-dev (changes requested):

Bloo-dev:
I think we should avoid using CMD for purposes other than visuals.

hozz8528:
> 
> 
> I think we should avoid using CMD for purposes other than visuals.

well, are there any other ways this could be achieved then? or other alternatives to reduce the excessive randomness of casting?

PR #584, opened 2021-05-07
Sign Inscriber Module

MichaelMiner137:
Featues
- Allows the player to copy text from a book onto a sign item which can then be placed with the stored text

Crafting
- Place a written book in the top left of a custom crafter and place a single sign in the top middle. This will move the sign to the bottom right slot with each line of text correlating to the first 4 pages in the book.


Bloo-dev:
We don't think the interface this module provides (an empty GUI once you place the sign) is intuitive. The concept of this is neat but the execution isn't ideal.
It's probably best to hold this back until the GUI either updates properly or doesn't open upon placement of a pre-written sign.

PR #591, opened 2021-05-09
Poses Pack: Improve reliability of pose modification

Luewd:
Dennis recently fixed an issue where pose would not be updated properly due to the pose for `Head`/`Body` first being overwritten to `[0f,0f,0f]`, which would cause the NBT tag to be removed, which makes the NBT path used to copy the new pose fail.

However, this does not fix the case where the first element is exactly `0f`: when it is copied to the entity as part of pose modification, the whole pose will be deleted in the same way it was with the erroneous code setting the pose to `[0f,0f,0f]`, in turn making it impossible to copy the remaining elements as the NBT path fails.

This commit constructs the new pose in storage and then copies it to the entity all at once to avoid NBT being reset partway which would lead to same issue as Dennis partially fixed. Theoretically this could be done for `Body`/`Head` and not the other body parts (as arms and legs have different defaults), but it is better to be consistent and safe for the other body parts, so this commit applies this fix to every pose as a failsafe.

### DIFF
--- a/gm4_poses_pack/data/gm4_poses_pack/functions/pose/player/arm_left.mcfunction
+++ b/gm4_poses_pack/data/gm4_poses_pack/functions/pose/player/arm_left.mcfunction
@@ -1,5 +1,11 @@
 # @s = armor_stand ..1 from writable_book
+
+# Initialize scoreboard and storage with pose data.
 function gm4_poses_pack:pose/update_values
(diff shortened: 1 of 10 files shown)


misode (approved):
There is some redundancy in these functions but there was already some inconsistency regarding the `gm4_pose_changed` tag so I think it's fine.

PR #592, opened 2021-05-14
Include lib_player_heads in metallurgy

Bloo-dev:
This PR updates metallurgy to use `lib_player_heads` instead of its own skull repair functionality.
This is an urgent update in preperation for the copper-bismuth conversion that will take place in 1.17 and should be merged ASAP. With this update, old copper ore and bands will remain copper items and will not brake when placed down.


Luewd (changes requested):
This pull request makes it so that *any* Obsidian Cast item on the ground will be converted into a Mundane Band, including freshly-minted shamirs or items dropped by the player. I suggest that this pull request is closed in favor of one I will be making soon, which updates `lib_player_heads` to support repairing skulls based on the `SkullOwner.Name` tag (which can hold a string identifying the shamir so that placed/broken Obsidian Casts do not become Mundane Bands).

Note that Metallurgy will always require custom logic to repair Obsidian Casts as they currently do not hold any `SkullOwner.Name` tag, and `lib_player_heads` always merges its item data on top of whatever the item already has. I can fix both of these issues, but all previously existing items will still have the old format and hence require special handling by Metallurgy.

PR #595, opened 2021-05-20
Add Vecto Shamir (temporary boat/minecart)

Dennis-0:
Copper shamir for boots - It creates a temporary boat/minecart

- Sneak on rails to summon a minecart
- Sneak on top of water next to a solid block to summon a boat (with enough water around it)

After 2 seconds it will start checking if the boat/minecart is being used by the player that summoned it. If this player is not in the minecart at the next clock cycle, it will be destroyed.

The 'solid block' check will include more blocks in the future, besides just #gm4:full_collision)


Luewd:
> I'm not completely satisfied with the way boats are spawned, because its rotation in water isn't applied properly and the chain of `if block water ...` checks doesn't look ideal (but it works, which is what counts)

1) This does not handle waterlogged blocks such as kelp or slabs: you may want to use `if predicate` to check for varying possibilities (this is how Soul Glass' beacon passable predicate works).
2) If you want to cut down on the number of `execute if block` subcommands, you could actually merge them into a single predicate (also how Soul Glass' beacon passable predicate works).

gm4_vecto_shamir/data/gm4_vecto_shamir/predicates/sneaking.json
```diff
@@ -0,0 +1,14 @@
...
+  "condition": "minecraft:entity_properties",
+  "entity": "this",
+  "predicate": {
+    "flags": {
+      "is_sneaking": true
+    },
+    "equipment": {
+      "feet": {
+        "nbt": "{gm4_metallurgy:{has_shamir:1b,active_shamir:'vecto'}}"
+      }
```
Luewd, line 11:
```suggestion
      "feet": {
        "tag": "gm4:boots",
        "nbt": "{gm4_metallurgy:{has_shamir:1b,active_shamir:'vecto'}}"
      }
```
If you do not check for a specific item or item tag, you will encounter a bug where the predicate incorrectly succeeds despite the boots being unequipped. The solution used for the Celaro shamir was to add a tag to `data/gm4/tags/items/boots.json` and check for it in the predicate. The contents of this tag are as follows:
```json
{
  "values": [
    "minecraft:netherite_boots",
    "minecraft:diamond_boots",
    "minecraft:golden_boots",
    "minecraft:chainmail_boots",
    "minecraft:iron_boots",
    "minecraft:leather_boots"
  ]
}
```

Luewd (changes requested):

gm4_vecto_shamir/data/gm4_vecto_shamir/predicates/sneaking.json
BluePsychoRanger: This must be why Enderpuffs kept breaking

Dennis-0: So I need to copy that tag, or should those armor tags be added to the gm4 base, if we need it more than once?

BluePsychoRanger: I think item tags for each of the armor items is a good idea; #gm4:armor should also be added in case we want all armor items

gm4_vecto_shamir/data/gm4_vecto_shamir/functions/vehicle/check_water.mcfunction
```diff
@@ -0,0 +1,13 @@
+#run from active
```
BluePsychoRanger, line 1: nearly all of this function can be replaced with a predicate. The predicate can check for `fluid:"water"`. I think just checking a 3x3 for water or `#gm4_no_collision` will suffice.

gm4_vecto_shamir/data/gm4_vecto_shamir/functions/vehicle/check.mcfunction
```diff
@@ -0,0 +1,5 @@
+#run from vehicle/spawn_(boat/minecart) and vehicle/check
+
+execute as @e[tag=gm4_ml_vehicle] at @a[tag=gm4_ml_invehicle,predicate=gm4_vecto_shamir:not_in_vehicle] if score @s gm4_ml_vecto_id = @p[tag=gm4_ml_invehicle,predicate=gm4_vecto_shamir:not_in_vehicle,distance=..1,sort=nearest] gm4_ml_vecto_id run function gm4_vecto_shamir:vehicle/remove
+execute unless entity @e[tag=gm4_ml_vehicle] run tag @a remove gm4_ml_invehicle
+execute if entity @e[tag=gm4_ml_vehicle] run schedule function gm4_vecto_shamir:vehicle/check 16t
```
BluePsychoRanger, line 5: Note that this means if all vehicles are unloaded (e.g. if a player dies in the boat and immediately respawns), the clock will stop until the next time a boat/minecart is spawned which may cause issues.

gm4_vecto_shamir/data/gm4_vecto_shamir/functions/vehicle/check.mcfunction
```diff
@@ -0,0 +1,5 @@
+#run from vehicle/spawn_(boat/minecart) and vehicle/check
+
+execute as @e[tag=gm4_ml_vehicle] at @a[tag=gm4_ml_invehicle,predicate=gm4_vecto_shamir:not_in_vehicle] if score @s gm4_ml_vecto_id = @p[tag=gm4_ml_invehicle,predicate=gm4_vecto_shamir:not_in_vehicle,distance=..1,sort=nearest] gm4_ml_vecto_id run function gm4_vecto_shamir:vehicle/remove
```
BluePsychoRanger, line 3: `type=#vecto:vehicles` should be added in the `@e` selector, in which `vecto:vehicles` contains boats and minecarts

BluePsychoRanger: `@p` already sorts by nearest, so the `sort=nearest` is unnecessary

gm4_vecto_shamir/data/gm4_vecto_shamir/functions/vehicle/spawn_boat.mcfunction
```diff
@@ -0,0 +1,15 @@
...
+#@s = player with shamir equipped, sneaking above water
+
+tag @s add gm4_ml_invehicle
+
+# spawn boat
+summon minecraft:boat ~ ~ ~ {Invulnerable:1b,Tags:["gm4_ml_vehicle"]}
+execute rotated as @p[tag=gm4_ml_invehicle] run tp @e[tag=gm4_ml_vehicle] ~ ~ ~ ~ ~
+
+# link player to boat with id
+execute store result score @s gm4_ml_vecto_id run data get entity @s UUID[0]
```
BluePsychoRanger, line 11: players should get their `gm4_ml_vecto_id` score via a join advancement, so it only ever happens once per player

gm4_vecto_shamir/data/gm4_vecto_shamir/functions/vehicle/check.mcfunction
```diff
@@ -0,0 +1,5 @@
...
+execute unless entity @e[tag=gm4_ml_vehicle] run tag @a remove gm4_ml_invehicle
```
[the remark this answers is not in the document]
BluePsychoRanger: I think the best way to do this is to tag the player when they summon a boat/minecart and then have a clock check if they have the tag, but are not in a temp vehicle, then from there kill all boats/minecarts that have the same `gm4_ml_vecto_id` score as the player.

gm4_vecto_shamir/data/gm4_vecto_shamir/functions/vehicle/spawn_minecart.mcfunction
```diff
@@ -0,0 +1,14 @@
...
+
+# spawn minecart
+summon minecraft:minecart ^ ^.1 ^ {Invulnerable:1b,Tags:["gm4_ml_vehicle"]}
+
+# link player to minecart with id
+execute store result score @s gm4_ml_vecto_id run data get entity @s UUID[0]
+scoreboard players operation @e[type=minecart,tag=gm4_ml_vehicle,distance=..1,sort=nearest,limit=1] gm4_ml_vecto_id = @s gm4_ml_vecto_id
+
+# start loop for check not in vehicle
+schedule function gm4_vecto_shamir:vehicle/check 40t
```
BluePsychoRanger, line 14: In theory this should give players 2 seconds to get into the vehicle, but in a multiplayer setting, this will not be the case. temp vehicles should get a timer score that goes counts to 3, and once it's at 3 it will start checking if there's a player inside it.

gm4_vecto_shamir/data/gm4_vecto_shamir/functions/vehicle/remove.mcfunction
```diff
@@ -0,0 +1,10 @@
+#run from active
```
BluePsychoRanger, line 1: header is incorect; this doesn't run from active; it runs from vehicle/check

gm4_vecto_shamir/data/gm4_vecto_shamir/functions/vehicle/spawn_minecart.mcfunction
```diff
@@ -0,0 +1,14 @@
+#run from active
+#@s = player with shamir equipped, sneaking on rails
+
+tag @s add gm4_ml_invehicle
+
+# spawn minecart
+summon minecraft:minecart ^ ^.1 ^ {Invulnerable:1b,Tags:["gm4_ml_vehicle"]}
+
+# link player to minecart with id
+execute store result score @s gm4_ml_vecto_id run data get entity @s UUID[0]
```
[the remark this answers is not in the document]
Dennis-0: What's that advancement? I don't see it in the list

gm4_vecto_shamir/data/gm4_vecto_shamir/functions/vehicle/spawn_minecart.mcfunction
Dennis-0: I know it resets the schedule timer when it's run again, but I've thought about it, and the alternative seems to be 2 more commands in main for the counter and checks. I'm not sure if that's worth it, since this "issue" would only occur if 2 or more players use this feature within 2 seconds of eachother, with the first player not actually getting in their vehicle, and if this does happen, it only means the vehicle is destroyed 2 seconds later than technically intended.

gm4_vecto_shamir/data/gm4_vecto_shamir/functions/vehicle/check.mcfunction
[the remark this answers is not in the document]
Dennis-0: > this removes the tag `gm4_ml_invehicle` tag from all players? Is this intentional?

That was indeed intentional, because no player is supposed to have the tag if no temp vehicles exist.

the `gm4_ml_invehicle` tag was meant as "this player is supposed to be in a temp vehicle. If not, kill the temp vehicle with their id"

gm4_vecto_shamir/data/gm4_vecto_shamir/predicates/sneaking.json
Luewd: I don't think the tag needs to be in base, but it should be in the `gm4` namespace as the tag is generic and multiple modules may do the same thing, so it makes sense to have a standardized name for a tag that you can copy-paste between modules. This is similar to how every pack includes the `gm4:root` advancement.

gm4_vecto_shamir/data/gm4_vecto_shamir/functions/vehicle/spawn_minecart.mcfunction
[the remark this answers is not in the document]
Dennis-0: Moved it to a `tick` advancement. Hope that's what you meant

gm4_vecto_shamir/data/gm4_vecto_shamir/predicates/sneaking.json
Dennis-0: Does this bug not happen for other shamirs? Or should this be changed in the others too eventually

SpecialBuilder32: We already have a number of common block tags in the base, I don't see why we shouldn't also make common item tags like for boots and such

gm4_vecto_shamir/data/gm4_vecto_shamir/functions/vehicle/check.mcfunction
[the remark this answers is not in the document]
Dennis-0: > I think the best way to do this is to tag the player when they summon a boat/minecart and then have a clock check if they have the tag, but are not in a temp vehicle, then from there kill all boats/minecarts that have the same `gm4_ml_vecto_id` score as the player.

From what I can see, that's exactly what this function does here. In the next commit I'll rename the tag so it's a bit more clear what it does.

When a boat/minecart is summoned, the player gets a tag and the check loop starts, which keeps running until no temp vehicles exist

gm4_vecto_shamir/data/gm4_vecto_shamir/predicates/sneak_validblock.json
```diff
@@ -0,0 +1,63 @@
...
+      }
+    ]
+  },
+  {
+    "condition": "minecraft:inverted",
+    "term": {
+      "condition": "minecraft:location_check",
+      "predicate": {
+        "fluid": {
+          "tag": "minecraft:water"
```
Bloo-dev, line 58: These should probably use `#gm4:water` instead as there are some water-like blocks that aren't actually water, e.g. bubble columns or seagrass.

gm4_vecto_shamir/data/gm4_vecto_shamir/predicates/check_water/below.json
```diff
@@ -0,0 +1,64 @@
...
+        "offsetX": 0,
+        "offsetY": -1,
+        "offsetZ": 0,
+        "predicate": {
+          "block": {
+            "tag": "minecraft:trapdoors",
+            "state": {
+              "open": "false",
+              "half": "bottom",
+              "waterlogged": "true"
```
Bloo-dev, line 57: Instead of listing a couple of waterloggable blocks, these should probably use `#gm4:waterloggable` to make the module more maintainable.
The same goes for the other predicates in `check_water/`.

gm4_vecto_shamir/data/gm4_vecto_shamir/functions/vehicle/spawn_boat.mcfunction
```diff
@@ -0,0 +1,20 @@
...
+
+# link player to boat with id
+scoreboard players operation @e[type=minecraft:boat,tag=gm4_ml_vecto_vehicle,distance=..1,sort=nearest,limit=1] gm4_ml_vecto_id = @s gm4_ml_vecto_id
+
+# set boat type
+execute if predicate gm4_vecto_shamir:biome/acacia run data merge entity @e[type=minecraft:boat,tag=gm4_ml_vecto_vehicle,distance=..1,sort=nearest,limit=1] {Type:"acacia"}
+execute if predicate gm4_vecto_shamir:biome/birch run data merge entity @e[type=minecraft:boat,tag=gm4_ml_vecto_vehicle,distance=..1,sort=nearest,limit=1] {Type:"birch"}
+execute if predicate gm4_vecto_shamir:biome/dark_oak run data merge entity @e[type=minecraft:boat,tag=gm4_ml_vecto_vehicle,distance=..1,sort=nearest,limit=1] {Type:"dark_oak"}
+execute if predicate gm4_vecto_shamir:biome/jungle run data merge entity @e[type=minecraft:boat,tag=gm4_ml_vecto_vehicle,distance=..1,sort=nearest,limit=1] {Type:"jungle"}
+execute if predicate gm4_vecto_shamir:biome/spruce run data merge entity @e[type=minecraft:boat,tag=gm4_ml_vecto_vehicle,distance=..1,sort=nearest,limit=1] {Type:"spruce"}
```
Bloo-dev, line 20: Referencing the freshly spawned boat by location and sorting is unsafe. Instead, give the freshly spawned boat a tag (e.g. `Tags:["gm4_vecto_vehicle","gm4_new_vecto_vehicle"]`) and target it by that. At the end of the function, remove that tag from all boats.

gm4_vecto_shamir/data/gm4_vecto_shamir/functions/init.mcfunction
```diff
@@ -0,0 +1,8 @@
+scoreboard objectives add gm4_ml_vecto_id dummy
```
Bloo-dev, line 1:
```suggestion
scoreboard objectives add gm4_vecto_id dummy
```
Shamirs should not use the `ml_` prefix. This change needs to be made in multiple sections of code throughout the module.

gm4_vecto_shamir/data/gm4_vecto_shamir/functions/vehicle/spawn_boat.mcfunction
```diff
@@ -0,0 +1,20 @@
+#run from vehicle/check_water and active
+#@s = player with shamir equipped, sneaking above water OR ice
+
+tag @s add gm4_ml_vecto_active
+
+# spawn boat
+execute if entity @s[y_rotation=-45..45] run summon minecraft:boat ~ ~ ~ {Invulnerable:1b,Tags:["gm4_ml_vecto_vehicle"],Rotation:[0f,0f]}
+execute if entity @s[y_rotation=45..135] run summon minecraft:boat ~ ~ ~ {Invulnerable:1b,Tags:["gm4_ml_vecto_vehicle"],Rotation:[90f,0f]}
+execute if entity @s[y_rotation=135..225] run summon minecraft:boat ~ ~ ~ {Invulnerable:1b,Tags:["gm4_ml_vecto_vehicle"],Rotation:[180f,0f]}
+execute if entity @s[y_rotation=-135..-45] run summon minecraft:boat ~ ~ ~ {Invulnerable:1b,Tags:["gm4_ml_vecto_vehicle"],Rotation:[270f,0f]}
```
Bloo-dev, line 10:
```suggestion
execute if entity @s[y_rotation=-45..45] run summon minecraft:boat ~ ~ ~ {Invulnerable:1b,Tags:["gm4_vecto_vehicle"],Rotation:[0f,0f]}
execute if entity @s[y_rotation=45..135] run summon minecraft:boat ~ ~ ~ {Invulnerable:1b,Tags:["gm4_vecto_vehicle"],Rotation:[90f,0f]}
execute if entity @s[y_rotation=135..225] run summon minecraft:boat ~ ~ ~ {Invulnerable:1b,Tags:["gm4_vecto_vehicle"],Rotation:[180f,0f]}
execute if entity @s[y_rotation=-135..-45] run summon minecraft:boat ~ ~ ~ {Invulnerable:1b,Tags:["gm4_vecto_vehicle"],Rotation:[270f,0f]}
```
Shamirs should not use the `ml_` prefix. This change needs to be made in multiple sections of code throughout the module.

Bloo-dev (changes requested):
There are some minor formatting issues that need to be resolved before this can advance to server testing:
- [ ] No longer use `ml_` as a prefix for vecto-internal tags and scoreboards
- [ ] Use `#gm4:water` and `#gm4_waterloggable`

gm4_vecto_shamir/data/gm4_vecto_shamir/predicates/sneak_validblock.json
Dennis-0: > These should probably use `#gm4:water` instead as there are some water-like blocks that aren't actually water, e.g. bubble columns or seagrass.

Both are checks for "water fluid", so any water-like block should be detected. I assume it's the same as both #gm4:water and #gm4:waterloggable combined. The first check is to make sure there's water below the player and the second check is to make sure you're not underwater.
The reason it checks for any "water fluid" here, is that this predicate is checked from main, so I assumed having more block checks.

Bloo-dev: That makes sense! Sorry, I don't know enough about minecraft's barebones fluid system.

gm4_vecto_shamir/data/gm4_vecto_shamir/predicates/check_water/below.json
Dennis-0: Most blocks in #gm4:waterloggable are not valid spawning spaces for boats. The blocks it checks for shouldn't intersect with the spawning area of the boat (bottom slabs and trapdoors are fine for example, but their "top" versions are not)

Bloo-dev: That is a valid point, though we are trading accuracy for maintainability.

Bloo-dev (approved):
This will now advance to server testing.

PR #597, opened 2021-05-20
Poses Pack advancement renamed

EpyonProjects:
* The display advancement should be the name of the module, if possible.


misode:
Renaming an advancements will cause players to lose it, and can break resource pack translation keys, or other data packs that reference this advancement.

misode:
Fuck it why not? We're in 1.17 and this isn't gameplay breaking.

PR #599, opened 2021-05-25
Base 1.2

misode:
Fixes a problem where players in survival would get reload logs.


Bloo-dev (commented):
This is kinda hard to read, what exactly does a score in `gm4_creative` mean?

misode:
> This is kinda hard to read, what exactly does a score in `gm4_creative` mean?

Essentially it stores the number of reloads ago that the player was seen in creative. This is used to make the player keep seeing logs if they had been in creative before. And is also used to make the logs visible to everyone if no creative player has been seen for a while.
This PR fixes the score for players that have never been in creative. Previously it would start at 1, and thus they would get logs shown for 10 reloads. This PR sets their store to 11 if they didn't have a score and are not in creative.

base/data/gm4-1.2/functions/log_start.mcfunction
```diff
@@ -1,4 +1,5 @@
 scoreboard players set @a[gamemode=creative] gm4_creative 0
+execute as @a[gamemode=!creative] unless score @s gm4_creative matches 0.. run scoreboard players operation @s gm4_creative = $cooldown gm4_creative
```
Bloo-dev, line 2: Shouldn't `$cooldown` be `#cooldown` as it is a constant? Our current use case for `#` is
> fake players used to store numbers for scoreboard operations should be named as a "#" followed by that number

However, I've been using `#` to marks constants in general.
Not that it would matter a lot, most of the current Gamemode 4 modules do not currently follow these standards but we should aim to change this over time.

Bloo-dev (commented):
I think what's been confusing me here was the choice of "cooldown" for a variable that does not tick down.

Apart from that this change seems to achieve what it is meant to do.

PR #602, opened 2021-05-29
Block Compressors: Arbitrary Compression

jpeterik12:
This started as just getting rid of the CMD of compressing items, and then morphed into the ability for a compressor to work with stack sizes other than just 16 and 64.

I haven't tested this extensively, but from what I've tested it works.
Also not sure if `item.mcfunction` is optimized well enough. I wrote it at 3:30 AM so probably could be done better.

99% sure it's compatible with old items and compressors


Bloo-dev:
More formal commit messages would be appreciated, yes. Your commits may show on the "version history" which is accessible via the website. E.g. [Book Binders](https://github.com/Gamemode4Dev/GM4_Datapacks/commits/master/gm4_book_binders).

gm4_block_compressors/data/gm4_block_compressors/functions/item.mcfunction
```diff
@@ -5,9 +5,30 @@
...
-execute if data storage gm4_block_compressors:temp/compressor {Items:[{Count:16b}]} if data storage gm4_block_compressors:temp/item_stack Item{Count:16b} unless data storage gm4_block_compressors:temp/item_stack Item.tag.gm4_compressed run function gm4_block_compressors:compress/16
-execute if data storage gm4_block_compressors:temp/compressor {Items:[{Count:1b}]} if data storage gm4_block_compressors:temp/item_stack Item{Count:1b,tag:{gm4_compressed:{value:64}}} run function gm4_block_compressors:decompress/64
-execute if data storage gm4_block_compressors:temp/compressor {Items:[{Count:1b}]} if data storage gm4_block_compressors:temp/item_stack Item{Count:1b,tag:{gm4_compressed:{value:16}}} run function gm4_block_compressors:decompress/16
+execute store success score @s gm4_bc_data if data storage gm4_block_compressors:temp/compressor {Items:[{Count:1b}]} if data storage gm4_block_compressors:temp/item_stack Item{Count:1b,tag:{gm4_compressed:{}}} run function gm4_block_compressors:item/decompress
```
Bloo-dev, line 8: This `score @s gm4_bc_data` does not seem to be kept between ticks, so why not use a fake player to increase readability?

gm4_block_compressors/data/gm4_block_compressors/functions/item.mcfunction
```diff
@@ -5,9 +5,30 @@
...
+execute if data storage gm4_block_compressors:temp/item_stack Item{Count:1b} run scoreboard players set @s gm4_bc_data 2
```
Bloo-dev, line 9: The check `if data storage gm4_block_compressors:temp/item_stack Item{Count:1b}` is evaluated twice in these two lines. Consider using a function to make sure it's only evaluated once.

gm4_block_compressors/data/gm4_block_compressors/functions/item/compress.mcfunction
```diff
@@ -0,0 +1,23 @@
+#@s = item entity to be compressed
+#run from block_compressors:item
```
Bloo-dev, line 2: Comments should be formatted as `# <comment>`.
Heads should also include the execution position `# at <location>` as well as - preferably - a short description of the function.

gm4_block_compressors/data/gm4_block_compressors/functions/item.mcfunction
```diff
@@ -5,9 +5,30 @@
...
+
+
+
+execute if score @s gm4_bc_data matches 0 unless data storage gm4_block_compressors:temp/item_stack Item.tag.gm4_compressed run function gm4_block_compressors:item/compress
+
+
+
 
 data remove storage gm4_block_compressors:temp/compressor Items
+data remove storage gm4_block_compressors:temp/item_stack Items
```
Bloo-dev, line 34: This should be `Item`.

gm4_block_compressors/data/gm4_block_compressors/functions/item.mcfunction
```diff
@@ -5,9 +5,30 @@
...
+execute if score @s gm4_bc_data matches 1 if data storage gm4_block_compressors:temp/compressor Items[5] store success score @s gm4_bc_data run data modify storage gm4_block_compressors:temp/compressor Items[5].Count set from storage gm4_block_compressors:temp/item_stack Item.Count
+execute if score @s gm4_bc_data matches 1 if data storage gm4_block_compressors:temp/compressor Items[6] store success score @s gm4_bc_data run data modify storage gm4_block_compressors:temp/compressor Items[6].Count set from storage gm4_block_compressors:temp/item_stack Item.Count
+execute if score @s gm4_bc_data matches 1 if data storage gm4_block_compressors:temp/compressor Items[7] store success score @s gm4_bc_data run data modify storage gm4_block_compressors:temp/compressor Items[7].Count set from storage gm4_block_compressors:temp/item_stack Item.Count
+execute if score @s gm4_bc_data matches 1 if data storage gm4_block_compressors:temp/compressor Items[8] store success score @s gm4_bc_data run data modify storage gm4_block_compressors:temp/compressor Items[8].Count set from storage gm4_block_compressors:temp/item_stack Item.Count
```
Bloo-dev, line 20: What is this for? The NBT `storage gm4_block_compressors:temp/compressor Items[8].Count` gets set here but the module doesn't seem to read this NBT anywhere, so why set it in the first place?

gm4_block_compressors/data/gm4_block_compressors/functions/item/compress.mcfunction
```diff
@@ -0,0 +1,23 @@
...
+
+#copy existing data for restoration at decompression
+data modify storage gm4_block_compressors:temp/item_stack Item.tag.gm4_precompression_tag set from storage gm4_block_compressors:temp/item_stack Item.tag
+
+#if no tag data to save, make a note
+execute unless data storage gm4_block_compressors:temp/item_stack Item.tag run data modify storage gm4_block_compressors:temp/item_stack Item.tag.gm4_precompression_tag set value "no_tag_data_to_preserve"
```
Bloo-dev, line 8: I know you adapted this from the current version of Block Compressors, but whilst you're updating this we should consider standardizing it.
This is a weird and non-standard way of introducing this custom tag. The NB tag should be under a tag named after the modules namespace. That would also prevent the tag from being of an ambiguous type. My suggest is to the following structure:
`gm4_block_compressors:{has_sealed_tag:<1/0b>,preserved_tag:{<data>}}`
The tag `has_sealed_data` might not even be necessary.

Now to the ugly: This would require implementing an update path for old items, but we already have one implemented in `decompress.mcfunction`, so that shouldn't be too bad.

gm4_block_compressors/data/gm4_block_compressors/functions/item/compress.mcfunction
```diff
@@ -0,0 +1,23 @@
...
+
+#add new data to indicate store compression state
+execute unless data storage gm4_block_compressors:temp/item_stack Item.tag.CustomModelData run data modify storage gm4_block_compressors:temp/item_stack Item.tag.CustomModelData set value 16
+execute store result storage gm4_block_compressors:temp/item_stack Item.tag.gm4_compressed.value int 1 run data get storage gm4_block_compressors:temp/item_stack Item.Count
+data merge storage gm4_block_compressors:temp/item_stack {Item:{Count:1b,tag:{HideFlags:1,Enchantments:[{id:"minecraft:protection",lvl:0s}]}}}
+
+#lore
+execute as @e[type=armor_stand,limit=1,tag=gm4_block_compressor_processing] run loot replace entity @s weapon.offhand loot gm4_block_compressors:compressed_lore
+execute as @e[type=armor_stand,limit=1,tag=gm4_block_compressor_processing] run data modify storage gm4_block_compressors:temp/item_stack Item.tag.display.Lore append from entity @s HandItems[1].tag.display.Name
```
Bloo-dev, line 17: Write `tag` before `limit`.

gm4_block_compressors/data/gm4_block_compressors/functions/item/decompress.mcfunction
```diff
@@ -12,5 +12,6 @@ execute if entity @s[nbt={Item:{tag:{gm4_precompression_tag:"no_tag_data_to_pres
 
 #restore original item tag, overwriting tags from compression
 data modify entity @s Item.tag set from entity @s Item.tag.gm4_precompression_tag
+data merge entity @s {PickupDelay:4}
```
Bloo-dev, line 15: Note to self: Check whether `merge` or `modify` is faster for constants.

Bloo-dev (changes requested):
I approve of the idea of this update, however, this is very convoluted and I feel like `item.mcfunction` takes a very complicated approach to a simple task that I can't comprehend.

gm4_block_compressors/data/gm4_block_compressors/functions/item.mcfunction
jpeterik12: I wasn't sure if the overhead of calling another function was worse than doing another `if data storage`.

gm4_block_compressors/data/gm4_block_compressors/functions/item.mcfunction
jpeterik12: This is how I do the comparisons.
The `if score` makes sure it only runs until it finds a match, and doesn't overwrite the match if later ones do not match.
The `if data` checks if the that slot exists, as otherwise it'll fail regardless.
The actual comparison is done by the `store success` and the `data modify`. If you try to copy a nbt value, but the original value is the same as the new value, you it will fail, and you'll get a success of 0.  If it ever gets set to zero, the following steps get skipped, and it'll run `compress.mcfunction`

An alternative way of doing this would be to copy the item's `Count` to a score, copy each slot's `Count` to a score, then to compare each of those scores with the items score. I'm not sure which would be better.

Breakdown of the number of commands used for each method in the worst case (no matches)

Example comparison for both
NBT:
```
execute if score action gm4_bc_data matches 1 if data storage gm4_block_compressors:temp/compressor Items[1] store success score action gm4_bc_data run data modify storage gm4_block_compressors:temp/compressor Items[1].Count set from storage gm4_block_compressors:temp/item_stack Item.Count
```
1 `if score ... matches`, 1 `if data`, 1 `score success score`, 1 `data merge ... from ...`

Scores
```
execute unless score action gm4_bc_data matches 1 store result score slot_1 gm4_bc_data run data get storage gm4_block_compressors:temp/compressor Items[0].Count
execute unless score action gm4_bc_data matches 1 if score slot_1 gm4_bc_data = item_count gm4_bc_data run function gm4_block_compressors:item/compress
```
2 `if score ... matches`, 1 `if score ... = ...`, 1 `score result score`, 1 `data get`

Bloo-dev: So this is to ensure all stacks inside the compressors are the same size?
I think previously compressors just looked at the last item, which is very useful.

jpeterik12: This will check if any item in the compressor matches the stack size of the item. It checks if it can copy the stacksize from the item to slot n. If that copy fails, it knows that both the item and slot n have the same count, and stops further checks.


The current version does the same thing. The current version allows you to put 64, 16, and 1 item into the same compressor, and it'll be able to do all 3 actions, no matter where those items are in the compressor. It can do this using `Items[{Count:64b}]`, which checks if any slot has 64 items. I can't do that because the 64 has to be dynamic, so direct comparison of each one slot is the best that can be done as far as I know.

gm4_block_compressors/data/gm4_block_compressors/functions/item.mcfunction
```diff
@@ -1,13 +1,13 @@
...
-execute if data storage gm4_block_compressors:temp/compressor {Items:[{Count:16b}]} if data storage gm4_block_compressors:temp/item_stack Item{Count:16b} unless data storage gm4_block_compressors:temp/item_stack Item.tag.gm4_compressed run function gm4_block_compressors:compress/16
-execute if data storage gm4_block_compressors:temp/compressor {Items:[{Count:1b}]} if data storage gm4_block_compressors:temp/item_stack Item{Count:1b,tag:{gm4_compressed:{value:64}}} run function gm4_block_compressors:decompress/64
-execute if data storage gm4_block_compressors:temp/compressor {Items:[{Count:1b}]} if data storage gm4_block_compressors:temp/item_stack Item{Count:1b,tag:{gm4_compressed:{value:16}}} run function gm4_block_compressors:decompress/16
+execute if data storage gm4_block_compressors:temp/item_stack Item{Count:1b} run function gm4_block_compressors:item/try_decompress
+execute unless score action gm4_bc_data matches 2 unless data storage gm4_block_compressors:temp/item_stack Item.tag.gm4_block_compressors unless data storage gm4_block_compressors:temp/item_stack gm4_compressed run function gm4_block_compressors:item/try_compress
```
misode, line 10: The `action gm4_bc_data` score doesn't make sense semantically anymore. Consider changing it to a simple boolean `$compressed` and you can store the success of the command on line 9.

gm4_block_compressors/data/gm4_block_compressors/functions/item/compress.mcfunction
```diff
@@ -0,0 +1,22 @@
...
+# at location of dropper (compressor)
+# run from block_compressors:item
+# compresses item by setting count, storing current item tag, and adding lore/enchantment glint
+
+# copy existing data for restoration at decompression, second line fixes bug where gm4_block_compressors gets saved to old_tag
+data modify storage gm4_block_compressors:temp/item_stack Item.tag.gm4_block_compressors.old_tag set from storage gm4_block_compressors:temp/item_stack Item.tag
+data remove storage gm4_block_compressors:temp/item_stack Item.tag.gm4_block_compressors.old_tag.gm4_block_compressors
+
+# add tag to indicate store compression size
+execute store result storage gm4_block_compressors:temp/item_stack Item.tag.gm4_block_compressors.compression_level int 1 run data get storage gm4_block_compressors:temp/item_stack Item.Count
```
misode, line 11: I believe you can use an item modifier to replace most of this function. (`set_count` with 1, `set_lore` using the score, `copy_nbt` for old_tag and for `compression_level`, `set_nbt` for enchantment glint)

gm4_block_compressors/data/gm4_block_compressors/functions/item/decompress.mcfunction
```diff
@@ -0,0 +1,21 @@
...
+# at location of dropper (compressor)
+# run from block_compressors:item
+# decompresses item by setting count and restoring old item tag
+
+# code automatically handles archaic items, this newer old items
+data modify storage gm4_block_compressors:temp/item_stack Item.tag.gm4_block_compressors.compression_level set from storage gm4_block_compressors:temp/item_stack Item.tag.gm4_compressed.value
+data modify storage gm4_block_compressors:temp/item_stack Item.tag.gm4_block_compressors.old_tag set from storage gm4_block_compressors:temp/item_stack Item.tag.gm4_precompression_tag
+
+# set item count
+data modify storage gm4_block_compressors:temp/item_stack Item.Count set from storage gm4_block_compressors:temp/item_stack Item.tag.gm4_block_compressors.compression_level
```
misode, line 11: Again, you can probably use an item modifier here.

misode (changes requested):
I agree that `gm4_block_compressors.compression_level` is technically the correct way of doing it, but there is nothing wrong with `gm4_compressed.value` so I'd rather we stick with that to avoid needing even more upgrade paths and stacking issues.

gm4_block_compressors/data/gm4_block_compressors/functions/item/compress.mcfunction
jpeterik12: Talked to misode, said not to do this because one can't use item modifiers on items entities, and I don't know if copying to the armor stand's hand and then using a modifier would really be an improvement.

misode (approved):

gm4_block_compressors/data/gm4_block_compressors/functions/process.mcfunction
```diff
@@ -1,6 +1,8 @@
 # @s = block compressor
 # run from main
+tag @s add gm4_block_compressor_processing
 execute positioned ~ ~1 ~ align xyz as @e[type=item,dx=0] positioned ~0.5 ~-0.9 ~0.5 run function gm4_block_compressors:item
+tag @s remove gm4_block_compressor_processing
 
 particle portal ~ ~1.1 ~ 0 0 0 0.1 20
 data merge entity @s {Fire:2000}
```
SpecialBuilder32, line 8: Strangely, Github is not showing any conflicts from this line, though [this commit](https://github.com/Gamemode4Dev/GM4_Datapacks/commit/b83816d695326a3589ddb749cc48249c49157a23) should create one. 

Unless my Github is just acting odd, this should be changed to use `HasVisualFire`

SpecialBuilder32 (approved):
I like this update; and agree that it will make for some very cool wormhole factories. Also a big fan of the challenge advancement 💯

gm4_block_compressors/data/gm4_block_compressors/functions/process.mcfunction
BluePsychoRanger: I believe there’s no conflicts because this isn’t changing any of those lines, it’s simply adding lines before and after line 3, which the PR you linked to doesn’t touch. If it was the other way around it might cause a conflict (i.e. if this one was merged first, then the visual fire PR was created, then the lines it’s changing would be different)

SpecialBuilder32: Oh excellent look at git working the way it should :tada:

gm4_block_compressors/data/gm4_block_compressors/functions/item/decompress.mcfunction
BluePsychoRanger: I've tested it:
`modify`: AVG tps: 329
`merge`: AVG tps: 283.6

`merge` is worse. I suspect it's because when you do merge, it has to parse the entire nbt to check if it's already set to that, but modify only checks that specific nbt tag?

gm4_block_compressors/data/gm4_block_compressors/functions/item/compress.mcfunction
```diff
@@ -0,0 +1,22 @@
...
+
+# set count to 1 and add compressed item visuals 
+data merge storage gm4_block_compressors:temp/item_stack {Item:{Count:1b,tag:{HideFlags:1,Enchantments:[{id:"minecraft:protection",lvl:0s}]}}}
+execute as @e[type=armor_stand,tag=gm4_block_compressor_processing,limit=1] run loot replace entity @s weapon.offhand loot gm4_block_compressors:compressed_lore
+execute as @e[type=armor_stand,tag=gm4_block_compressor_processing,limit=1] run data modify storage gm4_block_compressors:temp/item_stack Item.tag.display.Lore append from entity @s HandItems[1].tag.display.Name
+
+# clone data to original item entity
+data modify entity @s Item set from storage gm4_block_compressors:temp/item_stack Item
+data merge entity @s {PickupDelay:4}
```
BluePsychoRanger, line 20:
```suggestion
data modify entity @s PickupDelay set value 4
```

misode: It's fine in this case, but please don't go around the entire repo to replace all our `data merge` commands for a marginal performance benefit.

Bloo-dev (approved):
Changes approved, good to merge.

PR #605, opened 2021-06-01
Module Icon Update

BluePsychoRanger:
- Update all Icons to 512x512 to allow the website to extract from Github rather than rely on external (easily outdated) database
- Update some Icons that Hozz revamped
- Add credit for all icon designs
- changed "Creator" to "Creators" in the credits section if there are more than one person listed
- fixed some instances of "accidentalgames" and "Bluefire"


Bloo-dev (approved):
I've manually checked out all files and corrected some minor crediting errors. This seems fine.
Waiting on @sparks0 to confirm that the current website can handle this.

misode (changes requested):
The current site actually uses SVGs for module icons. It would be much better to store both a pack.png (with the maximum size Minecraft can use) and a pack.svg here for each module.

BluePsychoRanger:
> The current site actually uses SVGs for module icons. It would be much better to store both a pack.png (with the maximum size Minecraft can use) and a pack.svg here for each module.

True, but I'd like to point out the the current implementation of SVGs has some weird errors, causing lines that shouldn't be visible (e.g. Tower Structures, Ziprails). This is because the SVG has overlapping objects that are the same color, but when viewed on the (new) website, it doesn't correctly flatten the layers. If this can't be fixed, I think using the 512x512 images is the best option. 

Not using SVGs in the data packs will also allow other people to submit module icons without needing an SVG creating software. Though it will likely be converted to an SVG anyways in the future, so this isn't much of an issue.

BluePsychoRanger:
pack.svg has been added to all the modules now, this is ready to be used on the website

sparks0:
This looks all good to me, assuming we've done a JSON validation check on the pack.mcmetas? Idk if our auto-check does that?

sparks0 (approved):

misode (approved):

PR #609, opened 2021-06-04
Reduce range for dragon flap sound when end fishing

hozz8528:
The range right now (50) is huge. You normally stop hearing sounds at a distance of 15, so I reduced this to that. Right now it's kinda hard/annoying to stay away from other players that are fishing near you


BluePsychoRanger (approved):
Yeah, I put it at 50 originally because I wasn't sure how far the bobber could go. 15 is a good number to use

SpecialBuilder32 (approved):

misode:
My bad, I didn't know this PR existed so this change was already made is ebbc600f5b14383adbc5defc313ceea9523bd800.

PR #611, opened 2021-06-06
Add Undead Horses to Everstone

BluePsychoRanger:
- tho not normally obtainable in survival, baby zombie and skeleton horses do exist


SpecialBuilder32 (approved):
Looks like a nice simple change

PR #614, opened 2021-06-09
Chairs update 1.17

Dennis-0:
- Punching chair drops saddle
- Saddle of destroyed chair can't directly create new chair
- Moved scoreboard counter from main to advancement
- Chair killed by void because of visual cloud in 1.17
- Removed rotation correction


gm4_chairs/data/gm4_chairs/loot_tables/drop_saddle.json
```diff
@@ -6,9 +6,15 @@
       "entries": [
         {
           "type": "minecraft:item",
-          "name": "minecraft:saddle"
+          "name": "minecraft:saddle",
+          "functions": [
+            {
+              "function": "minecraft:set_nbt",
+              "tag": "{Tags:[\"gm4_chairs_ignore\"]}"
```
misode, line 13: You can't apply entity NBT like this, the `set_nbt` function sets item NBT inside the `tag:{}` compound.

gm4_chairs/data/gm4_chairs/advancements/hit.json
```diff
@@ -0,0 +1,22 @@
...
+    "punch": {
+      "trigger": "minecraft:player_hurt_entity",
+      "conditions": {
+        "entity": [
+          {
+            "condition": "minecraft:entity_properties",
+            "entity": "this",
+            "predicate": {
+              "type": "minecraft:pig",
+              "nbt": "{Tags:[\"gm4_chairs\"]}"
```
misode, line 12: This would be a good place to use the `gm4_chairs` team and skip needing an NBT check (even though it will have marginal benefits).

gm4_chairs/data/gm4_chairs/loot_tables/drop_saddle.json
Dennis-0: Oops, I meant to remove that part from when I was testing as I didn't need it anymore

SpecialBuilder32 (approved):
Tested and it all works as expected. The punching change is nice but I worry people might accidentally hit their chairs often? Not sure. 

I don't know why there are test failures on this PR as it has nothing to do with liquid tanks

misode (approved):

PR #615, opened 2021-06-09
Crossbow Cartridges update 1.17

Dennis-0:
New projectiles:
- Axolotl
- Tnt
- Redstone torch
- Soul torch

Checks only run if holding valid items in your offhand, instead every time when shooting an arrow with a crossbow.

Update block tag to make use of #gm4:full_collision.


SpecialBuilder32 (approved):
All these features and structure look good. I have not tested in-game though

Dennis-0:
This latest commit just changed the structure of the tripwire part, but it should work exactly the same

gm4_crossbow_cartridges/data/gm4_crossbow_cartridges/functions/projectile/redstone_torch/check.mcfunction
```diff
@@ -0,0 +1,19 @@
+# Check blocks, to place or drop redstone torch
+# @s = arrow with tag gm4_cb_rtorch, in a block
+# at @s
+# run from projectile/redstone_torch/loop
+
+# summon torch item if torch already exists at this block
+execute if block ~ ~ ~ #gm4_crossbow_cartridges:torch run loot spawn ~ ~ ~ loot gm4_crossbow_cartridges:redstone_torch
```
BluePsychoRanger, line 7: why do these use custom loot tables? You can just reference the minecraft block loot table for the torch types. Example below
```suggestion
execute if block ~ ~ ~ #gm4_crossbow_cartridges:torch run loot spawn ~ ~ ~ loot minecraft:blocks/redstone_torch
```

Dennis-0: Good question, didn't actually know about vanilla minecraft loot tables for some reason. Same problem for the saddle in Chairs then

BluePsychoRanger: Actually no, not the same for Chairs. This one is a block loot table, meaning when a player breaks a torch this loot table will drop (which is consistent with what would happen here). There's no block loot table for saddles cuz, well it's not a block. I'm taking care of that tho, so don't worry about chairs

Dennis-0: Yep just looked into it, I didn't make the connection that the torches are blocks so just repeated what I did for chairs, explains why I overlooked these minecraft loot tables

PR #616, opened 2021-06-23
Particles Pack: More particles

Dennis-0:
As requested by BartTheBart, some new particles that could add a lot to certain themes.
I understand that it might be a bit much to have double the amount of particles, but at least some of these would be great to have (like blossom, lava, enchant etc.)

Falling dust:
- Blossom (pink)
- Autumn (orange and brown)
- Spring (lime)
- Winter (white) (Replaces snow)
- Ash (black)

Other:
- Lava
- Fiesta (totem of undying)
- ~~Confetti (fountain shooting totem of undying particles)~~
- Ender (portal particles towards armour stand)
- Enchant (enchant particles towards armour stand)


Dennis-0:
> Would PR #617 replace the features introduced here, or would it be another additional option for types of particles?

That PR is mostly about the idea, but in case that PR is accepted, we should have a look which of the hardcoded particles are worth keeping. 
So it would be additional particles, not necessarily replacing them

SpecialBuilder32 (approved):
Tested and all these work as expected. I think the sizes and particle counts look pretty good too.

Bloo-dev (changes requested):
We looked at these changes in game and have some requests:

- remove the `confetti` particle
- remove `snow` or make `snow` and alias for winter

Bloo-dev (approved):

PR #617, opened 2021-06-23
Particles Pack: Allow all particles and particle cloud size

Dennis-0:
This is just about the idea, so I know parts of the code could be improved or removed, but the idea is that you can make any particle instead of just one from a predefined list.

In a written book:
- Page 1: `particle`
- Page 2: `<particle name>`
- Page 3 (optional): `small` or `big`

`<particle name>` can be any particle name in the game. For particles that require extra inputs, add it on the same page.
Examples for Page 2: `lava`, `dust 1 1 1 1`, `block diamond_block` , `item diamond`

Page 3 is to set the size of the particle cloud, where it defaults to a size of 1, `small` sets the size to 0 and `big` sets the size to 3. This size could be fully customizable too, but it shouldn't get much bigger than 3.

### DIFF
--- a/gm4_particles_pack/data/gm4_particles_pack/functions/apply_book.mcfunction
+++ b/gm4_particles_pack/data/gm4_particles_pack/functions/apply_book.mcfunction
@@ -3,7 +3,7 @@
 # run from gm4_better_armour_stands/store_book_pages
 
 # reset armor_stand and particle cloud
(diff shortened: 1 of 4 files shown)


Bloo-dev:
I think this gives a bit much control to the player, though we should keep this in mind for the future.

PR #618, opened 2021-06-24
Added undead ignore tag

EpyonProjects:


misode (commented):
I'd rather move the tag check to the `died` function, so we don't need the extra reset command in main

misode (approved):

PR #620, opened 2021-06-28
Added Glow Squid to Ink Spitting Squid

EpyonProjects:
* an additional advancement for glow squids
* changed two block tag checks into one water predicate

### DIFF
--- a/gm4_ink_spitting_squid/data/gm4/advancements/glow_ink_spitting_squid.json
+++ b/gm4_ink_spitting_squid/data/gm4/advancements/glow_ink_spitting_squid.json
@@ -1,0 +1,29 @@
+{
+  "display": {
+    "icon": {
+      "item": "minecraft:glow_ink_sac",
+      "nbt": "{CustomModelData:1}"
+    },
+    "title": {
+      "translate": "%1$s%3427655$s",
+      "with": [
+        "The Light, It Burns!",
+        {"translate": "advancement.gm4.glow_ink_spitting_squid.title"}
+      ]
+    },
+    "description": {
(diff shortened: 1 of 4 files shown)


gm4_ink_spitting_squid/data/gm4/advancements/glow_ink_spitting_squid.json
```diff
@@ -0,0 +1,29 @@
...
+    "description": {
+      "translate": "%1$s%3427655$s",
+      "with": [
+        "Glow due to an underwater glow squid",
+        {"translate": "advancement.gm4.glow_ink_spitting_squid.description"}
+      ],
+      "color": "gray"
+    }
+  },
+  "parent": "gm4:root",
```
SpecialBuilder32, line 23: I would recommend making this a child of the normal squid advancement, and perhaps making it a "goal" type frame as glow squids are slightly more rare

SpecialBuilder32 (changes requested):
Tested and it works as expected.

SpecialBuilder32 (approved):

gm4_ink_spitting_squid/data/gm4_ink_spitting_squid/predicates/in_water.json
```diff
@@ -0,0 +1,8 @@
+{
+  "condition": "minecraft:location_check",
+  "predicate": {
+    "fluid": {
+      "fluid": "minecraft:water"
```
BluePsychoRanger, line 5: this should've been `"tag": "minecraft:water"`. `"fluid"` makes it only work with source blocks

PR #621, opened 2021-06-28
Adds coarse and rooted dirt to BoO grass-able

EpyonProjects:


SpecialBuilder32 (approved):
Tested and works as expected. This is a good change I think; might help make Desire Lines easier to deal with. 

Sidenote: Did the Boots of Ostara clock ever get slowed down? I could've sworn that the placing of grass_blocks used to be much faster.

PR #623, opened 2021-07-05
Some Fruiting Trees Parity Fixes

BluePsychoRanger:
- Leaves are replaceable when a new tree grows into them
- When a tree grows, it should replace grass_block and mycelium with dirt

### DIFF
--- a/gm4_apple_trees/data/gm4_fruiting_trees/functions/tree/initialize.mcfunction
+++ b/gm4_apple_trees/data/gm4_fruiting_trees/functions/tree/initialize.mcfunction
@@ -12,6 +12,6 @@ scoreboard players set $max_layer gm4_tree_layer -1
 function #gm4_fruiting_trees:tree/initialize
 
 # start generation, unless expansion force quit generation (e.g. due to missing dy).
-execute unless score $cancel_generation gm4_tree_layer matches 1.. run function gm4_fruiting_trees:tree/analyze_seed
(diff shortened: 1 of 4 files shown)


gm4_apple_trees/data/gm4_fruiting_trees/functions/tree/initialize.mcfunction
```diff
@@ -12,6 +12,6 @@ scoreboard players set $max_layer gm4_tree_layer -1
 function #gm4_fruiting_trees:tree/initialize
 
 # start generation, unless expansion force quit generation (e.g. due to missing dy).
-execute unless score $cancel_generation gm4_tree_layer matches 1.. run function gm4_fruiting_trees:tree/analyze_seed
+execute unless score $cancel_generation gm4_tree_layer matches 1.. at @s run function gm4_fruiting_trees:tree/analyze_seed
```
misode, line 15: Doesn't this already run at the sapling marker?

misode (approved):

gm4_apple_trees/data/gm4_fruiting_trees/functions/tree/initialize.mcfunction
BluePsychoRanger: Yes, but if you move the marker during the initialize function (line 12)—for example for a 2x2 tree—then the reference gets reset to the old location when actually spawning in the tree. This would ensure that the location is set correctly after the initialize functions are run.

SpecialBuilder32 (approved):
All looks good :+1:

PR #624, opened 2021-07-05
Cozy campfires 1.17 update

misode:
- Uses a marker instead of an area effect cloud
- Cleaned up folder structure (sorry Bloo)
- Optimizations


SpecialBuilder32 (approved):
Tested and everything works as expected. The new code structure is also nice and straightforward. 

I thought however that the campfires weren't supposed to give absorption to players who took damage recently to prevent them as infinite-health sources in PVP scenarios. Was this feature removed?

misode:
> I thought however that the campfires weren't supposed to give absorption to players who took damage recently to prevent them as infinite-health sources in PVP scenarios. Was this feature removed?

@SpecialBuilder32 The module used to check for `HurtTime:0s`, which means the player cannot be hurt in the past 10 ticks. However the module only runs every 80 ticks. The decision to remove this check was also influenced by the high performance impact.

PR #626, opened 2021-07-07
Updated 3 new shamirs to match new load format

EpyonProjects:
As requested by 2cdc9a671522109127c7f20d4fc58360053e433a and 30f70bb509f92474d6e6e912c047601edcf56b3a


misode:
When you switch to optional tag entries, you can remove the empty metallurgy tag.

PR #634, opened 2021-07-13
Better Armour Stands: Turn correct way

Dennis-0:
`turn left` was turning clockwise and `turn right` was turning anticlockwise.
It would make sense if left and right is relative to the armour stand or from a top view, so this corrects the rotation to that point of view


SpecialBuilder32 (commented):
I don't really know how I feel about this change. It kinda feels weird when you're facing an armor stand you're configuring, and you tell it to turn left and it starts rotating to the players' right. I see the argument for both directions but I don't know which is more intuitive.

PR #636, opened 2021-07-13
Fulcio Raycast Fix

BluePsychoRanger:
- Fulcio uses a raycast instead of a hardcoded 3 blocks, so if there's a block in between the player and the 3rd block, it'll spawn the fulcio block in front of the block
- Fix the bug where you can drop an item and the predicate doesn't update
- converted AECs to markers


SpecialBuilder32 (approved):
Tested and Approved. Everything looks good :+1:

PR #638, opened 2021-07-14
Apple Trees Markers and Raycast Fix

BluePsychoRanger:
- area effect clouds replaced by markers
- raycasting checks the 6 adjacent blocks so if a player places it on a non-solid hitbox, it'll check surrounding blocks (thanks Epyon)
- apple tree saplings won't drop a sapling unless an oak sapling is actually dropped (usually only happens in creative)
- technically load score should go up since fruiting trees check for markers for the sapling instead of area effect clouds now, but there's no expansions or reliant modules yet?


BluePsychoRanger:
Don’t merge this yet; I need to make a change when I get back to my computer. We can’t use the dx=0 check for markers cuz they don’t have a hitbox

SpecialBuilder32 (approved):
Tested and I couldn't get the raycast to not work as expected so this is a pretty good change.

PR #640, opened 2021-07-15
Enderman Support Class: No effects in creative mode

Dennis-0:
Shulker and skeletons give effects (blindness and weakness) to nearby players, but players in creative mode should be excluded


SpecialBuilder32 (approved):
I agree this is a good change

Bloo-dev (approved):

PR #641, opened 2021-07-20
Xp Storage: Update sneaking check and fix visual bug

Dennis-0:
- Fixed bug where it displayed "Stored Experience:  Points" if no xp was stored yet, with `join` advancement that sets your score to 0 if needed
- Moved sneaking check to predicate instead of scoreboard
- Depositing all levels can be done instantly (without having to deposit 1 level first, so it would work when sneaking onto an ender chest without un-sneaking first)
- Updated all comments in right formatting

### DIFF
--- a/gm4_xp_storage/data/gm4_xp_storage/advancements/join.json
+++ b/gm4_xp_storage/data/gm4_xp_storage/advancements/join.json
@@ -1,0 +1,10 @@
+{
+    "criteria": {
+      "join": {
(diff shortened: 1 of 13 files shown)


misode (approved):
This has a feature change in that you no longer need to deposit one level in order to deposit all, a clock cycle later, but this is a change that makes sense to me.

SpecialBuilder32 (approved):
Tested and approved

PR #642, opened 2021-07-20
Beehive Inspectors: Fix visual bug

Dennis-0:
Fixed bug where the beehive lore would display "Honey: " without a number, if broken without silk touch, by setting this nbt if it's missing.

This also means that empty beehives will stack, mined with and without silk touch.


SpecialBuilder32 (approved):
Tested and I could not reproduce the bug you described. 

On a side note, is there a reason we decided not to add this lore to the bee_nest item?

Dennis-0:
> On a side note, is there a reason we decided not to add this lore to the bee_nest item?

I think a combination of a bee nest being a 'natural' so you wouldn't be able to clearly see this (as lore), and needing more checks

PR #644, opened 2021-07-26
Update Orbis for 1.17

BluePsychoRanger:
- [x] Make generation actually work: Summons the marker, then 1 clock cycle later, (attempt to) execute at it to generate the chunk
- [x] Use markers instead of area effect clouds
- [ ] Make chests seed random


BluePsychoRanger:
Oh should I increase the load version? Some modules (Dangerous Dungeons and Tower Structures) need to update their structures to use markers instead of area effect clouds for the population of chests and spawners. Apple Trees and Cooler Caves don't need an update tho, so what do I do with those?

misode:
> Actually does load version really matter, since 1.17 load basically resets all load versions?
@BluePsychoRanger Correct, we don't need to increase load version and we can make any breaking change we want.

gm4_orbis/data/gm4_orbis/functions/structure/chest_dungeon.mcfunction
```diff
@@ -6,3 +6,22 @@ execute store result score chest_height gm4_count run data get entity @s Pos[1]
...
 execute if score chest_height gm4_count matches 40.. run data merge block ~ ~ ~ {LootTable:"gm4_orbis:chests/dungeon_40"}
+
+# set the loot table seed
+data modify storage gm4_orbis:temp Pos set from entity @s Pos
+execute store result score $x gm4_orbis_config run data get storage gm4_orbis:temp Pos[0] 64
+execute store result score $y gm4_orbis_config run data get storage gm4_orbis:temp Pos[1] 64
+execute store result score $z gm4_orbis_config run data get storage gm4_orbis:temp Pos[2] 64
+data remove storage gm4_orbis:temp Pos
+
+scoreboard players operation $loot_table_seed gm4_orbis_config = #world_seed gm4_orbis_config
```
misode, line 17: Is there a reason why `world_seed` uses a `#` and `loot_table_seed` uses a `$`?

gm4_orbis/data/gm4_orbis/functions/structure/chest_dungeon.mcfunction
```diff
@@ -6,3 +6,22 @@ execute store result score chest_height gm4_count run data get entity @s Pos[1]
...
+scoreboard players operation $loot_table_seed gm4_orbis_config *= $x gm4_orbis_config
+scoreboard players operation $loot_table_seed gm4_orbis_config *= $y gm4_orbis_config
+scoreboard players operation $loot_table_seed gm4_orbis_config *= $z gm4_orbis_config
+
+execute store result block ~ ~ ~ LootTableSeed long 256 run scoreboard players get $loot_table_seed gm4_orbis_config
+
+scoreboard players reset $x gm4_orbis_config
+scoreboard players reset $y gm4_orbis_config
+scoreboard players reset $z gm4_orbis_config
+scoreboard players reset $loot_table_seed gm4_orbis_config
```
misode, line 27: I'm not a fan of all this resetting storage and scoreboards when it serves no purpose. It's less efficient for no gain.

gm4_orbis/data/gm4_orbis/functions/structure/spawn.mcfunction
```diff
@@ -6,8 +6,9 @@ function #gm4_orbis:structure/post_spawn
 scoreboard players add structure_count gm4_orbis_config 1
 
 kill @e[tag=gm4_orbis_structure_pos]
-summon area_effect_cloud ~ ~ ~ {Tags:[gm4_orbis_structure_pos]}
+summon marker ~ ~ ~ {CustomName:'"gm4_orbis_structure_pos"',Tags:[gm4_orbis_structure_pos]}
```
misode, line 9: No point in naming the entity when it only exist during this function

gm4_orbis/data/gm4_orbis/functions/structure/chest_tower.mcfunction
```diff
@@ -8,3 +8,22 @@ execute if score chest_height gm4_count matches 90..109 run data merge block ~ ~
...
+execute store result score $y gm4_orbis_config run data get storage gm4_orbis:temp Pos[1] 64
+execute store result score $z gm4_orbis_config run data get storage gm4_orbis:temp Pos[2] 64
+data remove storage gm4_orbis:temp Pos
+
+scoreboard players operation $loot_table_seed gm4_orbis_config = #world_seed gm4_orbis_config
+scoreboard players operation $loot_table_seed gm4_orbis_config *= $x gm4_orbis_config
+scoreboard players operation $loot_table_seed gm4_orbis_config *= $y gm4_orbis_config
+scoreboard players operation $loot_table_seed gm4_orbis_config *= $z gm4_orbis_config
+
+execute store result block ~ ~ ~ LootTableSeed long 256 run scoreboard players get $loot_table_seed gm4_orbis_config
```
misode, line 24: What is the math behind this? Why is each coordinate multiplied my 64, then added all together? And why is the resulting seed always a multiple of 256?

gm4_orbis/data/gm4_orbis/functions/structure/spawn.mcfunction
SpecialBuilder32: Naming all entities is good practice in case the entity stays around longer than it should. It makes it easier to debug when you can just /say @e and discover where the problematic entities are from.

It shouldn't be any slower since there's already nbt being set

gm4_orbis/data/gm4_orbis/functions/structure/chest_dungeon.mcfunction
BluePsychoRanger: `#` are used for constants, `$` are used for non-constant fake players

gm4_orbis/data/gm4_orbis/functions/structure/chest_dungeon.mcfunction
BluePsychoRanger: That's what I assumed, but I thought @Bloo-dev liked keeping debug stuff clean

gm4_orbis/data/gm4_orbis/functions/structure/chest_tower.mcfunction
BluePsychoRanger: Good point, I did it cuz I was thinking that they'd need to be multiplied by something anyways for the decimals, but those are always aligned, and for the 256 I wanted nearby chests to get seeds that were farther from each other, but I guess it doesn't matter as long as they're all different

misode:
Since `gm4_orbis_config` is a scoreboard that you might want on the sidebar (for various stats or orbis pre-gen), it should be kept clean. Either use another objective like `gm4_count` or use `#x` so it's hidden.

Bloo-dev (changes requested):
There are some changes that need to be made to the chest seed calculation.

The current argumentation is symmetric under permutations of the coordinates, i.e. chests at the 3 tuples (1,2,3) and (1,3,2) would get the same seed assigned.
The situation is even worse if one of the components is 0. In that case all the affected chests are degenerate and their seed collapses to 0.

You should probably offset the coordinate components by a factor of 10, e.g. 1*x+10*y+100*z.

misode (changes requested):
Settings the chest loot seed is an interesting idea, but right now makes little sense and I feel is implemented badly. There is already non-seed-deterministic randomness in the structure variant selection which should be addressed first.

I would suggest to split these changes and temporarily undo the chest loot seed setting so we can merge the required changes for 1.17, then we can take a look at making structures more deterministic.

BluePsychoRanger:
> Settings the chest loot seed is an interesting idea, but right now makes little sense and I feel is implemented badly. There is already non-seed-deterministic randomness in the structure variant selection which should be addressed first.
> 
> I would suggest to split these changes and temporarily undo the chest loot seed setting so we can merge the required changes for 1.17, then we can take a look at making structures more deterministic.

Alright, I'm not entirely sure how to undo the commits though

misode (approved):

Bloo-dev (approved):

PR #645, opened 2021-07-27
Prefix CustomModelData

BluePsychoRanger:
Adds prefix "342" to every CustomModelData, followed by a 4-digit identifier.

Using a prefix will allow easier compatibility with other data packs, so resource packs can be merged without requiring a change in the data packs. 342 was chosen because it's the first 3 digits of our language translations. Note that the resource pack will need to be updated alongside this.

I've claimed the prefix 342 for Gamemode4 on this website: https://mcdatapack.vercel.app


misode (approved):
This is already a very impactful PR, the longer we wait the more conflicts we will get. Good to merge as soon as there is a PR on GM4_Resources for this. For the next few versions we should include both the prefixed and non-prefixed model overrides in the resource pack, so existing items can keep their custom texture. A script to generate model json files would be helpful regardless.

BluePsychoRanger:
I will remake this PR from scratch, as it has many conflicts

PR #646, opened 2021-07-28
Added Glow Item Frame support for 1.17 HoloTags

EpyonProjects:
Allows you to create Holographic Tags using Glow Item Frames

- separate entity checks are better than entity tags, with 2 entities.


misode (approved):

misode:
The implementation is fine, but since using glow item frames doesn't give any additional benefits to holograms, I don't think this is necessary.

PR #647, opened 2021-07-29
Update Display Frames to 1.17

EpyonProjects:
* Fixed detection so it now consistently creates Display Frames
* Changed from aec's to marker entities

### DIFF
--- a/gm4_display_frames/data/gm4_display_frames/functions/thrown_potion_tracking/spawn_tracker.mcfunction
+++ b/gm4_display_frames/data/gm4_display_frames/functions/thrown_potion_tracking/spawn_tracker.mcfunction
@@ -2,7 +2,7 @@
 # @s = Splash/Lingering Potion of Invisibility
 # run from tick
 
-summon minecraft:marker ^ ^ ^.03 {Tags:["gm4_df_potion_tracker"]}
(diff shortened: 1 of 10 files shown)


gm4_display_frames/data/gm4_display_frames/functions/thrown_potion_tracking/spawn_tracker.mcfunction
```diff
@@ -1,8 +1,8 @@
-# tracks thrown splash potions of invisibility with a line of aec's
+# tracks thrown splash potions of invisibility with a line of markers
 # @s = Splash/Lingering Potion of Invisibility
 # run from tick
 
-summon minecraft:area_effect_cloud ^ ^ ^.03 {Duration:2,Tags:["gm4_df_potion_tracker"]}
+summon minecraft:marker ^ ^ ^.03 {Tags:["gm4_df_potion_tracker"]}
```
SpecialBuilder32, line 5: This marker should have a CustomName for debug purposes.

SpecialBuilder32 (changes requested):
Tested; I was worried this might leave behind markers in strange situations, but I could not get it to break even with some quick-and-dirty chunk loading trickery.

Though I was the display-frames-god for a while, this does appear to be more robust and reliable in the detection of invisibility potions.

PR #648, opened 2021-07-29
Added piglin-chair reverting

EpyonProjects:
If a player is near (like 200 blocks) a chair when it converts to a piglin, it converts back 50t later


SpecialBuilder32 (approved):
This is a clever and lightweight solution to an obscure problem XD. Though I doubt this will really ever occur in game, since it doesn't add any continually running clock or processing, I don't see why **not** to include this.

Bloo-dev (approved):
Tested and code-checked. Ready to be merged.

PR #650, opened 2021-07-29
Enderman Support Class: Lowered Distance

EpyonProjects:
* Lowered distance from 25 to 16, to match sound distance.
* Should also help with lag.


misode:
Why would it needs to match the sound distance?

EpyonProjects:
No reason, it just seems like a needlessly large detection range

PR #652, opened 2021-07-29
Better Fire: Arrows nerfed

EpyonProjects:
* Non-Infinity Flame arrows despawn upon landing and have a chance to drop a flint.
* Reduces amount of damage because previously the flame arrow would constantly set fire, and continue to do so until they despawned.


gm4_better_fire/data/gm4_better_fire/loot_tables/flint.json
```diff
@@ -0,0 +1,55 @@
...
+            {
+              "function": "minecraft:set_lore",
+              "entity": "this",
+              "lore": [
+                {
+                  "translate": "%1$s%3427655$s",
+                  "with": [
+                    "Charred arrowhead",
+                    {
+                      "translate": "item.gm4.better_fire.flint"
```
misode, line 36: Use `item.gm4.better_fire.flint.lore` here.

gm4_better_fire/data/gm4_better_fire/functions/on_fire.mcfunction
```diff
@@ -1,5 +1,8 @@
+# @s, flame arrow or creeper on fire
+# runs from main
+
 #fire arrows place fire
-execute at @s[type=arrow] run setblock ~ ~ ~ fire keep
+execute as @s[type=arrow,tag=!gm4_cb_arrow,nbt={inGround:1b}] if data entity @s Fire at @s run function gm4_better_fire:arrow
```
misode, line 5: I can't find any reference to this `gm4_cb_arrow` tag.

misode (changes requested):
Since we don't usually play with this module it's hard to judge these changes, but they feel right.

misode:
Oh forgot to include: why summon a flint item when it always gets destroyed by the fire immediately?

gm4_better_fire/data/gm4_better_fire/functions/on_fire.mcfunction
EpyonProjects: #615 adds a common tag to all arrow-based projectiles used in Crossbow Cartridges.  Unknown if torches shot using Crossbow Cartridges should activate Better Fire.

EpyonProjects:
It _can_ spawn the flint, it's just extremely unlikely.  @Bloo-dev suggested to add the chance to even spawn the item.  My thinking was that the arrow on fire would burn the feather and stick.

gm4_better_fire/data/gm4_better_fire/loot_tables/flint.json
```diff
@@ -0,0 +1,55 @@
...
+                  {
+                    "translate": "item.gm4.better_fire.flint"
+                  }
+                ],
+                "italic": false
+              }
+            },
+            {
+              "function": "minecraft:set_nbt",
+              "tag": "{CustomModelData:1}"
```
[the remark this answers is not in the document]
EpyonProjects: We need to update the list :/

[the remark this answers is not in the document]
misode: Are you using the correct list? The google doc is updated.

gm4_better_fire/data/gm4_better_fire/functions/on_fire.mcfunction
misode: Instead of checking for the crossbow cartridges tag, better fire should add a general ignore tag which CC should set.

SpecialBuilder32 (approved):
Tested and approved. 

I'm not convinced this does too much to "nerf" flame arrows; I did some testing on a few forests (rip trees) and this did not decrease fire spread too much. However I still think this is a good change

gm4_better_fire/data/gm4_better_fire/loot_tables/flint.json
```diff
@@ -0,0 +1,55 @@
+{
```
BluePsychoRanger, line 1: please put this under the namespace `/loot_tables/entities/flame_arrow` or something similar. Don't forget to update the functions that call this loot table

PR #654, opened 2021-08-01
Smelteries 1.17

BluePsychoRanger:
- add copper
- add raw ores
- utilize `"items": []` properly


misode (approved):

SpecialBuilder32 (approved):
Tested and Approved. 

There has been some discussion about changing how smelteries operates in the future, but we will cross that bridge later.

PR #657, opened 2021-08-03
Beehive Inspectors: Main check to advancement trigger

Dennis-0:
Advancement `bee_nest_destroyed` triggers when a beehive is broken, so the main function can be removed.

### DIFF
--- a/gm4_beehive_inspector/data/gm4_beehive_inspector/functions/break_beehive.mcfunction
+++ b/gm4_beehive_inspector/data/gm4_beehive_inspector/functions/break_beehive.mcfunction
@@ -1,7 +1,0 @@
-# @s = player breaking beehive
-# at @s
-# run from advancement break
-
-advancement revoke @s only gm4_beehive_inspector:break
-
-execute as @e[type=item,nbt={Item:{id:"minecraft:beehive"},Age:0s},sort=nearest,limit=1] run function gm4_beehive_inspector:inspect
--- a/gm4_beehive_inspector/data/gm4_beehive_inspector/functions/break_beehive.mcfunction
+++ b/gm4_beehive_inspector/data/gm4_beehive_inspector/functions/break_beehive.mcfunction
@@ -1,0 +1,7 @@
+# @s = player breaking beehive
+# at @s
+# run from advancement break
+
+advancement revoke @s only gm4_beehive_inspector:break
+
+execute as @e[type=item,nbt={Item:{id:"minecraft:beehive"},Age:0s},sort=nearest,limit=1] run function gm4_beehive_inspector:inspect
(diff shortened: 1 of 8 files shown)


gm4_beehive_inspector/data/gm4_beehive_inspector/functions/inspect.mcfunction
```diff
@@ -1,10 +1,9 @@
-# Inspect beehive item
-# @s = beehive item
+# @s = beehive item, Age:0
 # at @s
-# run from main
+# run from break
 
 # store bee count in scoreboard
-execute store result score #bees gm4_beehive_insp run data get entity @s Item.tag.BlockEntityTag.Bees
+execute store result score $bees gm4_beehive_insp run data get entity @s Item.tag.BlockEntityTag.Bees
```
misode, line 6: This uses the objective that you removed

gm4_beehive_inspector/data/gm4_beehive_inspector/functions/inspect.mcfunction
```diff
@@ -1,10 +1,9 @@
-# Inspect beehive item
-# @s = beehive item
+# @s = beehive item, Age:0
 # at @s
-# run from main
+# run from break
```
BluePsychoRanger, line 3:
```suggestion
# run from break_beehive
```

gm4_beehive_inspector/data/gm4_beehive_inspector/loot_tables/lore.json
```diff
@@ -5,12 +5,12 @@
       "entries": [
         {
           "type": "minecraft:item",
-          "name": "minecraft:stick",
+          "name": "minecraft:honeycomb",
```
misode, line 8: 🤔

gm4_beehive_inspector/data/gm4_beehive_inspector/functions/break_beehive.mcfunction
```diff
@@ -0,0 +1,7 @@
+# @s = player breaking beehive
+# at @s
+# run from advancement break
+
+advancement revoke @s only gm4_beehive_inspector:break
+
+execute as @e[type=item,nbt={Item:{id:"minecraft:beehive"},Age:0s},sort=nearest,limit=1] run function gm4_beehive_inspector:inspect
```
misode, line 7: This should probably have a distance check

misode (commented):
I'm in favor of also adding this to bee nests to make it less confusing. The original reason to exclude those was mainly for performance reasons.

Dennis-0:
Removed the check for beehive to allow bee_nest too, and changed the sorting to `furthest`, as items thrown by the player at the same tick, would be selected instead as they would be closer to the player most of the time. If needed these thrown items can be ignored completely, but the extra check seems unnecessary for now as it can only really happen on purpose.

misode (approved):

SpecialBuilder32 (approved):

PR #659, opened 2021-08-06
Audere to use Item Modifiers

BluePsychoRanger:
for audere (and moneo, but that change will happen in #625), the shulker box trick was needed to replace the item in the player's hand, but now we have item modifiers to do that seamlessly


gm4_audere_shamir/data/gm4_audere_shamir/item_modifiers/netherite_update.json
```diff
@@ -0,0 +1,4 @@
+{
```
misode, line 1: Please rename this file to match the mcfunction name, it is infuriating me

PR #661, opened 2021-08-10
Heart Canisters Recipe Change

Bloo-dev:
_DISCLAIMER: This is simply a suggestion. Criticism is welcome._ 

This PR alters the way Heart Canisters, both tier 1 and tier 2, are crafted. It does not change how heart canisters work.

### But what for?
Heart Canisters are obviously very over-powered, and changing the recipe won't change that; instead this PR aims to make the recipes more memorizable and to make the step from tier one to tier two harder.
This is achieved by requiring more than one tier one to craft a tier two. The PR also alters the recipes for both to use items that are related to health-modifying buffs in game.

I created this PR as this has been talked about multiple times, but no one actually implemented this so far. This PR might help people to visualize these changes better.

Fig. 1
_Fig. 1: The new heart canister tier 1 recipe._
Fig. 2
_Fig. 2: The new heart canister tier 2 recipe._


BluePsychoRanger:
Glistering Melon Slices seem a little too easy. It takes 53.333 gold ingots to craft a full set of heart canisters. I would suggest golden apples be used in the recipe instead (or in addition) for the gold requirement increase and since it provides absorption, which is the closest thing to health boost offered by the vanilla game. Also wither skeleton skulls do not make much sense. The "lore" of wither skeleton skulls is that it's used to hold together the souls from the soul sand. A nether star would make a bit more sense as it holds power that can be translated into potion effects. Perhaps end crystal for the T1 recipe and nether star for the T2 recipe.


Heart Canister T1 Recipe Proposal:
```
           O - Obsidian
OIO        I - Iron Block
MEM        M - Glistering Melon
OAO        E - End Crystal
           A - Golden Apple
```

Heart Canister T2 Recipe Proposal:
```
OIO        O - Obsidian
HNH        I - Iron Block
OHO        H - Heart Can T1
           N - Nether Star
```

EpyonProjects:
I like Beeps' T1 but Bloo's T2.

misode:
@BluePsychoRanger Why not 3 golden apples instead of the glistering melons?

BluePsychoRanger:
@misode I just felt like Bloo wanted those glistering melons in there, and I think it's reasonable enough to include them

Bloo-dev:
I've implemented the golden apples and the end stone. I've also swapped the wither skull to a nether star, so you now have a choice between getting a beacon or an extra heart as a reward after killing a wither.
This has the side effect of making tier 2 heart canisters much more expensive -- but that's a good thing in my opinion. 3 rows of hearts are quite unnecessary and should be more of a novelty rather than the norm. Images attached below.

heart_canister_tier1
_Fig. 1: Heart Canister Tier 1 recipe._

heart_canister_tier2
_Fig. 2: Heart Canister Tier 2 recipe._

PR #651, opened 2021-07-29
Sunken Treasures 1.17

BluePsychoRanger:
- modified loot tables
- Fixes Squid and Guardian spawning
- make loot biome dependent
    - in oceans, rates will be very high, in rivers rates will be medium, and anywhere else rates will be low
    - in warm oceans coral can drop from sand
    - in lukewarm or warm oceans tropical fish can spawn from sand
- added gravel to sunken treasure, which has a different loot tables
- added shipwreck treasure and supply chests

New Loot Tables: https://docs.google.com/spreadsheets/d/15FemIXUjZOoXFcVOMcpTCLLPzSWevwn7fJ9mP58zRsQ/edit

PR #635, opened 2021-07-13
Midnight Menaces 1.17

BluePsychoRanger:
- move randomizers to predicates
- goblins and illusioners will simply replace new mobs, similar to how icy strays and sandy husks work; goblins or illusioners (depending on the night) will replace 2% of creepers, skeletons, and zombies
- loot tables standardized to specify that each loot table is an entity loot table

PR #630, opened 2021-07-13
SCUBA Gear 1.17 Rewrite

BluePsychoRanger:
- add headers for all functions
- utilize lib_player_heads for SCUBA Helmet
- use predicate to check for fluid: #water instead of relying on hardcoded waterloggable block tag
- water check happens at player's eye level instead of just ~ ~1 ~
- added (creative-only) loot tables for all items

PR #629, opened 2021-07-13
End Fishing 1.17 Rewrite

BluePsychoRanger:
- enderpuff slow falling is more reliable
- if flying with elytra, enderpuff slow falling is inverted (sneak to activate, instead of sneak to deactivate slow falling)
- fishing rod durability is properly calculated and applied using item modifiers
- end city treasure lore/CMD is done via loot tables instead of functions
- code optimizations
- uses markers instead of area effect clouds

PR #663, opened 2021-08-11
Make All Loot Tables Consistent

BluePsychoRanger:
- Instead of random loot tables in the `loot_table` namespace (e.g. `gm4_custom_crafters:destroy_custom_crafter`), each loot table will be put into a subfolder: items, blocks, entities, chests, gameplay, technical.

  - items are custom items added by modules
  - blocks are block loot tables for when a custom block is destroyed
  - entities are entity death loot tables for custom entities added by modules
  - chests are chest loot tables for lootable chests added by modules
  - gameplay are loot tables related to specific gameplay added by modules (e.g. end fishing or sunken treasure)
  - technical are loot tables that are used solely as helpers for functions, usually these items are used for data modifications then killed immediately

- Adds loot tables for every item added by each module

The following modules have not been modified because they should be (or have already been) modified by an existing PR:
- [x] Crossbow Cartridges (#615)
- [x] Better Fire (#652)
- [x] End Fishing (#629)
- [ ] All Metallurgy Modules (#625)
- [x] Midnight Menaces (#635)
- [ ] Potion Liquids (#610)
- [x] SCUBA Gear (#630)
- [x] Sunken Treasure (#651)
- [ ] Zauber Cauldrons (#627)


BluePsychoRanger:
Note that this PR should likely be merged before the 1.17 release because otherwise redirects would have to be made for loot tables that reference other loot tables (e.g. Block Compressors calls the Custom Crafters loot table, which is under a different namespace after this PR). This won't be a problem if we do it before the full 1.17 release because players need to update every module to 1.17 due to `load` changing, so we can do breaking updates all at once. This _could_ be merged later, but it would require either leaving redirects behind or increasing the load score for certain modules.

misode:
I'm not a fan of these changes. I see lots of trouble for no benefit.
- harder to give yourself custom items with commands
- we need to maintain legacy fallbacks for existing mobs and chests that are already in the world which reference old loot tables, for at least a major version
- potentially lots of merge conflicts
- history and git blame is lost

BluePsychoRanger:
> * harder to give yourself custom items with commands

It's not harder, since you'd still do the process of typing out the namespace then just using tab; I'd argue it's easier since you can see what type of loot table you're getting

> * we need to maintain legacy fallbacks for existing mobs and chests that are already in the world which reference old loot 
tables, for at least a major version

The only file that this pertains to is the Shulker from Lightning Rods, which 1) might not even release for 1.17, and 2) already has a 90% chance to drop nothing, so it's unlikely to be very noticeable since you either kill it completely or keep it and thus never kill it

> * potentially lots of merge conflicts

As stated in the original message I've left out modules that have open PRs

> * history and git blame is lost

git blame (and git lens) works as if this PR didn't happen; it treats it as the exact same file. It is true that history is lost though.

Bloo-dev:
I'll stick with what has been said above and close this PR as these changes can be introduced on a module-by-module basis during the regular update cycle, if a module requires this structure.

PR #664, opened 2021-08-11
Better Armour Stands: Tracking Update

Dennis-0:
#### Main feature changes:
- Book data given to armour stand by right clicking it with a book and quill, instead of throwing it at the armour stand
  - Armour stands placed in the world have arms enabled by default
  - `equip` takes the item from the player's offhand, and places this item into the slot the player is looking at (works on head, mainhand, and offhand). This can now also be used when the armour stand already has an item in this head/offhand slot, throwing out this previous item
  - `copy` modifies the armour stand items in the player's offhand 
    - Now also copies `NoGravity`/`Invisible`
    - Lore displays copied features (if different from default)
  - `paste` takes data from armour stand items in the player's offhand
- Changed `arms`, `base`, `size`, `visible`, `gravity` to toggles, with particles and sounds (instead of separate codes for e.g. `base` and `no base`)
- Changed the way poses are set, by tracking the player:
  - `pose` instead of `arm`/`leg`/`head`/`body`:  This detects what part of the armour stand you're looking at.
  - Sneak while right clicking armour stand to enable "alternative" method to drag parts
  - Sneak to align its rotation value to multiples of 15 degrees
  - (Optionally) `mirror`: This mirrors selected arm/leg to other arm/leg
  - Left click to "invert" the pose
- Changed `move` and `rotate` to `pose`, and right clicking the base plate of the armour stand:
  - Move: The armour stand will float in front of the player until it is right clicked again
    - Sneak to align its position to 1/16th of a block
  - Rotate (Sneak to activate): Drag the rotation of the armour stand
    - Sneak to align it to multiples of 15 degrees
 - Added `flip`: This will flip/mirror the full pose of the armour stand

#### Less important:
- Rotation values calcutated more accurately (because scores only saved integer numbers)
- Changed `data merge .. {..}` to `data modify .. set value ..`
- Changed advancement for giving arms to a general advancement for customizing the armour stand
- Improved comments
- [ ] Name for "Customize an armour stand" advancement


Bloo-dev:
Note: @Dennis-0  is still working on this.

Bloo-dev (commented):
This is not a full code review but I tested this in game and it worked well.
Yet there are some minor things you should tweak:
- The text displayed on the title bar should use translate to allow for translations

PR #665, opened 2021-08-15
Guidebook

BluePsychoRanger:
This module will add a dynamic guidebook for Gamemode 4. Each module will set their pages of the guidebook and will decide how each page is unlocked. Some modules have initial unlocked pages to help the players figure out the module, and most have unlockable pages which are granted and stored via advancement triggers.

To obtain a guidebook, players can loot village chests (45% chance for a village chest to contain one) or find zombie villagers (20% chance for a zombie villager to spawn with the guidebook in their hand).

If you want to test it and just read through every page, you can use the following command:
`/advancement grant @s through gm4_guidebook:root`

List of Page Unlock Events:
https://docs.google.com/document/d/1OLG1yFGyZX2wR-h6lyHhSgdjRKO2slLqUD5iRiEi8o0/edit

Below is a technical spreadsheet used to generate the pages.
https://docs.google.com/spreadsheets/d/1ToBBaGTEinkEVff60-zvP6Zbf8-9mJULhjQWriuiU0c/edit?usp=sharing


MichaelMiner137:
Shouldn't the author be the player and not Audrey? It doesn't make sense that she is the author when the player is the one that has to research and unlock the different pages.

gm4_pig_tractors/data/gm4_pig_tractors/advancements/guidebook/page_0.json
```diff
@@ -0,0 +1,136 @@
...
+              "score": "load.status"
+            },
+            "range": 1
+          },
+          {
+            "condition": "minecraft:entity_properties",
+            "entity": "this",
+            "predicate": {
+              "vehicle": {
+                "type": "minecraft:pig"
```
Dennis-0, line 65: This also triggers when sitting on a chair, so I think this should either ban the pigs with tag `gm4_chairs` specifically, or just ignore pigs with `NoAI`

PR #667, opened 2021-08-17
Forceload 1.17

BluePsychoRanger:
Forceload should now work with 1.17. Upgrading worlds from 1.16 should work as well.

- instead of checking for the entity (because you can't do that reliably anymore) it checks for the birch sign
- uses markers instead of area effect clouds
- uses x=0 instead of distance=0..
- force initializes overworld marker


Bloo-dev:
How do you wait for the chunk being loaded?

BluePsychoRanger:
> How do you wait for the chunk being loaded?

When you add a chunk to be force loaded, it _is_ forceloaded, but entities are not. So you can load a chunk and place blocks, check blocks, and even move entities there in the same tick. But you cannot check for an entity in the same tick. This is why I switched to just checking for the birch sign to determine if the chunk is loaded.

lib_forceload/data/gm4_forceload-1.0/functions/check_dimension.mcfunction
```diff
@@ -4,4 +4,6 @@
 advancement revoke @s only gm4_forceload-1.0:dimension_marking
 
 # spawns a dimension marker if one doesn't already exist in the dimension
-execute unless entity @e[type=minecraft:area_effect_cloud,tag=gm4_dimension,distance=0..] run function gm4_forceload-1.0:mark_dimension
+execute as @e[type=area_effect_cloud,tag=gm4_dimension] at @s run function gm4_forceload-1.0:update_marker
+execute store success score $initialized gm4_dimension if block 29999998 1 7133 birch_wall_sign
+execute if score $initialized gm4_dimension matches 0 run function gm4_forceload-1.0:mark_dimension
```
misode, line 9: Why can this not be done in 1 command?
```mcfunction
execute unless block 29999998 1 7133 birch_wall_sign run function gm4_forceload-1.0:mark_dimension
```

lib_forceload/data/gm4_forceload-1.0/functions/load.mcfunction
```diff
@@ -1,6 +1,6 @@
 scoreboard objectives add gm4_dimension dummy
 
-execute in minecraft:overworld unless block 29999998 1 7133 birch_wall_sign run function gm4_forceload-1.0:init_chunk
+execute store success score $initialized gm4_dimension if block 29999998 1 7133 birch_wall_sign
+execute if score $initialized gm4_dimension matches 0 run summon marker ~ 0 ~ {CustomName:'"minecraft:overworld"',Tags:["gm4_dimension","gm4_new_dimension"]}
+execute if score $initialized gm4_dimension matches 0 as @e[type=marker,tag=gm4_new_dimension,limit=1] run function gm4_forceload-1.0:init_ow_chunk
 function gm4_forceload-1.0:load_wait
-
-execute as @a at @s run function gm4_forceload-1.0:check_dimension
```
misode, line 6: What's the reason that the overworld needs a custom treatment rather than how it was done before?

lib_forceload/data/gm4_forceload-1.0/functions/mark_dimension.mcfunction
```diff
@@ -4,12 +4,11 @@
 function gm4_forceload-1.0:init_chunk
 
 # spawn dimension marker
-summon minecraft:area_effect_cloud 29999999 0 7135 {Duration:2147483647,Tags:["gm4_dimension","gm4_new_dimension"]}
+summon minecraft:marker ~ 0 ~ {Tags:["gm4_dimension","gm4_new_dimension"]}
```
misode, line 7: This function runs at worlds spawn or the player's position from the advancement, why are the absolute coordinates removed?

lib_forceload/data/gm4_forceload-1.0/functions/update_marker.mcfunction
```diff
@@ -0,0 +1,9 @@
+# @s = area effect cloud dimension marker
```
misode, line 1: The convention from other modules is to call this function `upgrade_marker`

misode (changes requested):

lib_forceload/data/gm4_forceload-1.0/functions/check_dimension.mcfunction
BluePsychoRanger: They changed how `unless block` works if the position isn't loaded, it will not run any commands after that. This is why I have to check the success of the `if block`

lib_forceload/data/gm4_forceload-1.0/functions/load.mcfunction
BluePsychoRanger: Since we're now checking the block instead of the marker, setting the overworld sign will cause the overworld marker to never generate.

lib_forceload/data/gm4_forceload-1.0/functions/mark_dimension.mcfunction
BluePsychoRanger: It needs to be spawned in a loaded position in order to perform `data modify` commands on it. It can't be found if it's spawned in a newly force loaded chunk.

lib_forceload/data/gm4_forceload-1.0/functions/update_marker.mcfunction
BluePsychoRanger: Fixed it in the latest commit

misode (approved):

PR #668, opened 2021-08-19
Standard crafting 1.17

EpyonProjects:
- Adds Amethyst and Dripstone block to item conversion
- Declares `gm4_custom_crafters:temp/crafter` storage


Bloo-dev (commented):
I don't really like the dripstone recipe cause it feels wrong. But so does turning four tiny pointy dripstone pieces into a solid cubic meter of dripstone, so I guess it's how vanilla wanted it to be.

PR #673, opened 2021-08-24
Reveal invisible armour stands with "glow"

Dennis-0:
Instead of revealing invisible armour stands when holding a book and quill, the book now requires a page with "glow" to avoid revealing their locations when it's not needed


Dennis-0:
This might not be the most optimal solution, but it might be good to not always reveal all invisible armour stands when holding a book, or by "ignoring" certain armour stands

PR #675, opened 2021-08-24
Offhand equip fix

EpyonProjects:
- Organizes predicates a bit
  - and adds offhand check for armorstand and empty


gm4_better_armour_stands/data/gm4_better_armour_stands/functions/apply_book.mcfunction
```diff
@@ -24,13 +24,13 @@ execute if data storage gm4_better_armour_stands:temp {pages:["gravity"]} run fu
 execute if data storage gm4_better_armour_stands:temp {pages:["turn"]} run function gm4_better_armour_stands:toggle/turn
 
 # Equip item into specified slot.
-# execute if data storage gm4_better_armour_stands:temp {pages:["equip"]} run function gm4_better_armour_stands:equip/select
+execute if data storage gm4_better_armour_stands:temp {pages:["equip"]} run function gm4_better_armour_stands:equip/select
```
misode, line 27: Should this be the place where we check if the offhand is empty?

Bloo-dev (approved):

gm4_better_armour_stands/data/gm4_better_armour_stands/functions/apply_book.mcfunction
EpyonProjects: I originally thought that was the case, but Dennis said that using "equip" with nothing in your offhand, then it should still pop out the item from the targeted armor stand.

Dennis-0: Depends on how we want to implement it. I wanted "equip" to act like interacting with the armour stand in vanilla, where it would swap items. So if you don't hold an item it would take the item from the armour stand

PR #677, opened 2021-08-25
Add lib_brewing

BluePsychoRanger:
This library allows modules to implement splash and lingering brewing stand recipes via loot tables. Each brewing stand has a marker entity to track when a potion is brewing; brewing stands will have to be broken and replaced for this to work in a pre-existing world.

- Included is a README.md which explains how to use the library

- Potion Liquids and Zauber Cauldrons have been implemented with this: Floating Potions can be brewed into splash and lingering variants, and brewing stands will explode when Zauber Potions are brewed into their splash and lingering variants.


lib_brewing/USAGE.md
```diff
@@ -0,0 +1,162 @@
...
+
+
+Additional functionality is available:
+- in the two created function files (<brewing_stand/splash.mcfunction> and <brewing_stand/lingering.mcfunction>) additonal commands can be run before the `/loot` command, in order to change the outcome of the loot tables
+
+- the function tag </data/gm4_brewing/tags/functions/finish_brew.json> runs AFTER default interactions of creating splash and lingering potions
+    - to test for items in the brewing stand, use the entity data from the marker:
+        - `execute if entity @s[nbt={data:{BLOCK_DATA}}]`
+        - e.g. this is the code that checks if a splash potion should be created: 
+            - `execute if entity @s[nbt={data:{Items:[{Slot:3b,id:"minecraft:gunpowder"}]}}]`
```
Bloo-dev, line 162: Did something go wrong with the markdown in this file? The formatting looks off

Bloo-dev (changes requested):
NBT stored on the marker entity should follow the Gamemode 4 formatting standard for NBT, this means:
- NB tags should be nested within a tag named after the module id
- NB tags should be in `snake_case`

In practice this means that you should change `Insert` to `gm4_brewing.insert`. We should make this change now as making it in post will lead to backwards compatibility issues.

lib_brewing/data/gm4_brewing-1.0/functions/prep_finish_brew.mcfunction
```diff
@@ -0,0 +1,12 @@
+# @s = brewing stand that finished brewing
+# run from check_brew
+
+# mark as finished
+tag @s remove gm4_brewing_tick
+tag @s add gm4_brewing_finished
+
+# store contents of brewing stand
+data modify entity @s data.Items set from block ~ ~ ~ Items
```
Bloo-dev, line 9: This should be changed to `data.gm4_lib_brewing.<key>`. I think a more descriptive name such as `data.gm4_lib_brewing.previous_items` might make for a cleaner look.

lib_brewing/data/gm4_brewing-1.0/functions/check_brew.mcfunction
```diff
@@ -0,0 +1,6 @@
+# @s = none
+# run from process
+
+execute as @e[type=marker,tag=gm4_brewing_tick] at @s if block ~ ~ ~ brewing_stand{BrewTime:1s} run function gm4_brewing-1.0:prep_finish_brew
+
+execute if entity @e[type=marker,tag=gm4_brewing_tick,limit=1] run schedule function gm4_brewing-1.0:check_brew 1t
```
Bloo-dev, line 6: We could aviod this selector by slightly modifying the function structure and by introducing another fake player.

By setting a fake player to a fixed value `A` at the start of this function and by setting that fake player to another fixed value `B` within `gm4_brewing-1.0:prep_finish_brew`, you could deduct that such a marker still exists. This would obviously require moving the block check into `gm4_brewing-1.0:prep_finish_brew` or a helper function which acts as a middle-man between this function and `gm4_brewing-1.0:prep_finish_brew`.

I have a suspicion that this decrease the average runtime in worlds with many entities.

Bloo-dev (changes requested):
Please note that I did not check `gm4_potion_liquids` in this review, as that is @SpecialBuilder32 's domain.
This looks good apart from two minor things I noticed.

lib_brewing/data/gm4_brewing-1.0/functions/check_brew.mcfunction
BluePsychoRanger: I've made a change, I don't understand the moving the block check part tho.

Bloo-dev: Not moving the block check will alter the logic, here are two diagrams for you.

Behavior **before** 8ccb2d3:
| marker exists? | brewing_stand{BrewTime:1s}? | Outcome       |
|----------------|-----------------------------|---------------|
| yes            | yes                         | Abort loop    |
| yes            | no                          | Continue loop |
| no             | yes                         | Abort loop    |
| no             | no                          | Abort loop    |

Behavior **after** 8ccb2d3:
| marker exists? | brewing_stand{BrewTime:1s}? | Outcome       |
|----------------|-----------------------------|---------------|
| yes            | yes                         | Abort loop    |
| yes            | no                          | Continue loop |
| no             | yes                         | Continue loop |
| no             | no                          | Continue loop |

BluePsychoRanger: But the score is only ever set to 0 if it exists. This would indeed continue the loop if the marker gets unloaded within the 16 ticks, but will also allow it to catch it when it does get reloaded. TBH I think this or the previous version is fine; for the previous version, the loop only ever runs if there is a brewing stand within 16 ticks of finishing the brew

Bloo-dev: I see, that's alright. Especially as the loop would be escaped if another brewing stand finishes brewing, so infinite loops would be unlikely.
@misode what do you think?

Bloo-dev (approved):

PR #680, opened 2021-09-14
Garden Variety Draft

MichaelMiner137:
All of the primary mechanics are done and I would like some help optimizing it before I start adding all the trees.

Additional Info:
Info: https://docs.google.com/document/d/1eB4xUxHo9Fq_IxzTToc9v36dYLg85LacTQ0wSIOpeF4/edit?usp=sharing
Storage: https://docs.google.com/spreadsheets/d/1ieVZzgK131HhOpTyIl70N_oD_lZQjcunC7XVewLUdEY/edit?usp=sharing
Trades: https://docs.google.com/spreadsheets/d/15BnbOuuu4VDdram5aXjNPHibA13xdoJCSPzjO9yYVsU/edit?usp=sharing

Current Features:
- 1 New Tree (King Palm)
- Analyzer (used to find what mutations a sapling has)
- Arborist Villager (sells a variety of saplings and other foliage items)
- Custom Saplings (copied from fruiting_trees and modified to fit my needs)
- Mutations (can randomly appear when a tree is grown, effects tree generation)

Missing Features:
- Crossbreeding (allows you to breed two trees to create another)
- Mutation Spreading (allows you to spread mutations between trees)
- 2x2 and 3x3 sapling support


Bloo-dev:
Is this meant to be a draft PR? If so, you can convert it to a draft via the web interface.

misode:
Why is GitHub drunk and added all these commits, despite me explicitly merging?

gm4_garden_variety/data/gm4_garden_variety_trees/functions/load.mcfunction
```diff
@@ -0,0 +1,6 @@
+execute if score gm4 load matches 1 if score gm4_garden_variety load matches 1 run scoreboard players set gm4_garden_variety_trees load 1
+execute unless score gm4 load matches 1 run data modify storage gm4:log queue append value {type:"missing",module:"Garden Variety Trees",require:"Gamemode 4"}
+execute unless score gm4_garden_variety load matches 1 run data modify storage gm4:log queue append value {type:"missing",module:"Garden Variety Trees",require:"Garden Variety"}
```
misode, line 3: Why is this module split in two, rather than a subfolder in the main namespace?

gm4_garden_variety/pack.mcmeta
```diff
@@ -0,0 +1,36 @@
...
+    "video_link": "",
+    "wiki_link": "https://wiki.gm4.co/wiki/Garden_Variety",
+    "required_modules": [
+        "custom_crafters",
+        "lib_lore"
+    ],
+    "recommended_modules": [
+        "orbis",
+        "podzol_rooting_soil",
+        "fruiting_trees"
```
misode, line 26: Fruiting trees is not a module, rather it should say it recommends apple trees. Why is apply trees not a required module? Does it implement its own custom tree and sapling logic? If so, that seems like a waste and makes things harder to maintain. Making fruiting trees a library may make sense.

gm4_garden_variety/data/gm4_garden_variety/functions/integration/fruiting_trees/update_leaves.mcfunction
```diff
@@ -0,0 +1,7 @@
+# checks whether @s is a fruit leaf, and if so allows further evaluation of the age
+# @s = @e[type=armor_stand,tag=gm4_fruiting_leaf]
+# at @s align xyz
+# run from #gm4_fruiting_trees:tree/leaf/fruiting/age
```
misode, line 4: From what I can tell this function isn't called from anywhere. Were you planning to integrate this with fruiting trees?

misode (commented):
I very much like the original concept of adding custom trees while also giving bees more uses. But it's made oh so complicated, almost matching metallurgy and zauber cauldrons. It includes a custom villager, an analyzer block, some stuff with soil conversion. All that just to get some custom trees seems a bit overkill imo, especially maintenance wise.

I didn't manage to test this in-game since I couldn't figure out how to obtain a palm tree, an arborist villager, or one of these mutated trees. So let me know if I'm misunderstanding some of the features of this module.

If I'd have to "fix" this module, I would try to simplify things. What elements are required to make this work? Does it really need an analyzer, or is there another more intuitive way for players to obtain the information that the analyzer would give them? Is the arborist villager required, or would a wandering trader also work? Are mutations necessary and do they add value to the module?

gm4_garden_variety/data/gm4_garden_variety_trees/functions/load.mcfunction
MichaelMiner137: I mirrored how fruiting_trees does it as apple_trees is under its own namespace. It also makes it much easier to understand how an expansion works since you can just copy the _trees namespace.

gm4_garden_variety/pack.mcmeta
MichaelMiner137: garden_variety does not require fruiting_tree mechanics however if it is enabled it does add additional functionality in the form of some trees having the chance of spawning with fruit leaves.

When I first made the module, fruiting_trees was required. However, for a variety of reasons I decided to remove this requirement.
- I have my own system for dropping saplings from the leaves that is more efficient. fruiting_trees uses armor stands for each leaf that has fruit, however, I use a single leaf for each tree that uses a marker.
- I don't remember the specifics so maybe I am misremembering. I remember running into problems when it came to merging the mutation system into sapling placement and breaking. Because of this I decided to just make my own system for handling saplings.
- I would love it if fruiting_trees was its own library because I do not like the idea of forcing the player to use apple_trees because fruiting_trees funcationality is required.

Bloo-dev:
Closing this as there has been no movement on this since March. This needs to be overhauled anyways, as world gen has changed a lot in the latest Minecraft version.

PR #683, opened 2021-09-23
Add Animi

Bloo-dev:
### Features
Animi is a Shamir for Metallurgy that makes items "soulbound". Items with Animi behave differently upon the player's death:
- Instead of being dropped, they are respawned with the player.
- Animi can be applied to all tools and armor, and other modules may register their own items by use of the function tag `#gm4_animi_shamir:mark_item_validity`.

When a player dies carrying Animi items, a soul particle is played for each Animi item (technically stack) in the player's inventory.

### Technical Details
This implementation uses `storage` to cache the player's items between death and respawning. However, this means that player NBT has to be read to obtain the player's `UUID` to create an index list in storage. These reads are only executed if a player dies with an Animi item.
At the time of advancement execution the player's inventory is already dropped and no longer available in NBT form. This has two consequences:
   **1.** The items can not be read from player NBT and must therefore be targeted as entities.
   **2.** Targeting the items as entities could potentially be unsafe, however, I've never noticed an item escaping undetected.

On the topic of **2.**: An undetected item would not result in a duplication glitch, as undetected items never make it into the item cache in storage. This means an undetected Animi item will simply drop like normal instead of being respawned with the player.

### Limitations
Animi does not generally preserve inventory sorting, as the items are no longer available in the player's NBT upon advancement reward execution. This implementation can not be adjusted to preserve sorting as that information is lost upon the player's death (when the items are dropped in entity form).
However, upon death players naturally drop items in ordef of inventory slots. This results in a fixed pickup order due to `arbitrary` item selection. Hence items will somewhat respect inventory order but will not reassemble the pre-death layout completely.

### DIFF
--- a/gm4_animi_shamir/data/gm4_animi_shamir/advancements/death.json
+++ b/gm4_animi_shamir/data/gm4_animi_shamir/advancements/death.json
@@ -1,0 +1,33 @@
+{
+  "criteria": {
+    "death": {
+      "trigger": "minecraft:entity_hurt_player",
+      "conditions": {
+        "player": [
+          {
+            "condition": "minecraft:entity_scores",
+            "entity": "this",
+            "scores": {
+              "gm4_animi_deaths": 1
+            }
+          },
+          {
+            "condition": "minecraft:value_check",
+            "value": {
+              "type": "minecraft:score",
+              "target": {
+                "type": "minecraft:fixed",
+                "name": "gm4_animi_shamir"
+              },
+              "score": "load.status"
+            },
+            "range": 1
+          }
+        ]
+      }
+    }
+  },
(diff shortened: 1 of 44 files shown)


Bloo-dev:
I'm looking into optimizing the 20Hz, as the current setup is not desirable.

Bloo-dev:
Further testing showed that this is actually multiplayer safe.
Players dying in the same tick still die one after another, triggering the death rewards in sequence, or as shown in a timeline:

- Player A dies
- Player A drops items
- Player A reward functions
- Player B dies
- Player B drops items
- Player B reward functions

Therefore the increased item detection radius does not cause serious issues.

PR #690, opened 2021-10-25
Chairs: General update for smaller issues

Dennis-0:
- Fixed player teleporting to void if saddle removed: The chair was teleported to the void (to avoid kill particles), which also teleported the player sitting on it and killed them. This change dismounts the player right before killing the chair.
- Saddle item is returned with original data by storing data in the pig/chair (in `HandItem[0]`)
- Fixed pig appearing for 1 tick when spawned by spawning it in the void and teleporting it back up
- Improved lightning fixes: Delay of 50 ticks is removed, and new pig/chair is spawned with the same position/rotation, instead of hardcoded summons for the 4 directions, to keep its rotation
- Void teleporting to -10000 instead of -1000, because the minimum world height is -2048
- Updated comments for the correct format


gm4_chairs/data/gm4_chairs/functions/lightning/tag.mcfunction
```diff
@@ -1,6 +1,8 @@
-# runs at player near lightning strike
+# Target NoAI zombified_piglin after lightning strike near player
 # @s = player
+# at @s
 # runs from advancement, gm4_chairs:lightning
 
 advancement revoke @s only gm4_chairs:lightning
-schedule function gm4_chairs:lightning/delay 50t
+
+execute as @e[type=zombified_piglin,nbt={NoAI:1b}] unless data entity @s Tags at @s run function gm4_chairs:lightning/revive_chair
```
Bloo-dev, line 8: This no longer verifies that the `zombie_piglin` is actually within a stair block, which is a necessary requirement for being a chair. You should probably include a block check just to be sure.

gm4_chairs/data/gm4_chairs/functions/create_chair.mcfunction
```diff
@@ -1,11 +1,22 @@
...
+# spawn chair
+summon minecraft:pig ~ ~-10000.39 ~ {Tags:["gm4_chairs"],Team:"gm4_chairs",NoAI:1b,Saddle:1b,NoGravity:1b,Silent:1b,DeathTime:19s,InLove:2147483647,Attributes:[{Name:"generic.movement_speed",Base:0.0}],ActiveEffects:[{Id:14,Amplifier:0,Duration:2147483647,ShowParticles:0b},{Id:11b,Amplifier:10b,Duration:2147483647,ShowParticles:0b}],DeathLootTable:"minecraft:empty"}
+
+# set chair to orientation of stairs
+execute if block ~ ~ ~ #minecraft:stairs[facing=north] positioned ~ ~-10000 ~ as @e[type=minecraft:pig,tag=gm4_chairs,distance=..0.4,limit=1] at @s run tp @s ~ ~10000 ~.05 0 0
+execute if block ~ ~ ~ #minecraft:stairs[facing=south] positioned ~ ~-10000 ~ as @e[type=minecraft:pig,tag=gm4_chairs,distance=..0.4,limit=1] at @s run tp @s ~ ~10000 ~-.05 180 0
+execute if block ~ ~ ~ #minecraft:stairs[facing=east] positioned ~ ~-10000 ~ as @e[type=minecraft:pig,tag=gm4_chairs,distance=..0.4,limit=1] at @s run tp @s ~-.05 ~10000 ~ 90 0
+execute if block ~ ~ ~ #minecraft:stairs[facing=west] positioned ~ ~-10000 ~ as @e[type=minecraft:pig,tag=gm4_chairs,distance=..0.4,limit=1] at @s run tp @s ~.05 ~10000 ~ -90 0
```
Bloo-dev, line 13: What are the advantages of this over the previous implementation? This is generally slower than the previous implantation.

gm4_chairs/data/gm4_chairs/functions/kill.mcfunction
```diff
@@ -1,9 +1,14 @@
...
+# @s = pig (chair) to be removed
+# at @s positioned ~ ~.5 ~ / ^ ^.9 ^.2
+# run from main / hit
 
 # spawn saddle item
 loot spawn ~ ~ ~ loot gm4_chairs:drop_saddle
 tag @e[type=minecraft:item,distance=..1,limit=1,sort=nearest] add gm4_chairs_ignore
 
+# dismount if sitting on chair
+execute align xyz positioned ~.5 ~ ~.5 as @a[predicate=gm4_chairs:sitting_in_chair,distance=..0.5] at @s align y run tp @s ~ ~1 ~
```
Bloo-dev, line 11:
```suggestion
execute align xyz positioned ~.5 ~ ~.5 as @a[predicate=gm4_chairs:sitting_in_chair,distance=..0.5] at @s align y run tp @s ~ ~ ~
```
Should be safer as `~ ~1 ~`. The latter could actually be abused or teleport the player into an invalid location.

gm4_chairs/data/gm4_chairs/functions/main.mcfunction
```diff
@@ -1,7 +1,7 @@
-#saddle item on a stair block -> spawn pig
-execute as @e[type=minecraft:item,nbt={Item:{id:"minecraft:saddle"}},tag=!gm4_chairs_ignore] at @s if block ~ ~ ~ #minecraft:stairs[half=bottom] positioned ~ ~-.1 ~ align xyz unless entity @e[type=minecraft:pig,tag=gm4_chairs,dx=0,dz=0,dy=0] at @s run function gm4_chairs:create_chair
+# detect saddle item on a stairs and spawn chair
+execute as @e[type=minecraft:item,nbt={Item:{id:"minecraft:saddle"}},tag=!gm4_chairs_ignore] at @s if block ~ ~ ~ #minecraft:stairs[half=bottom] align xyz positioned ~.5 ~ ~.5 unless entity @e[type=minecraft:pig,tag=gm4_chairs,distance=..0.4] run function gm4_chairs:create_chair
```
Bloo-dev, line 2: The previous implementation of this line made sure there is no chair already within the same block as this chair.
Your new implementation checks that there is no chair in a 0.4 block radius around the center of the bottom face of this block. Is this intended? See the attached image for a visualization.
Selection
_Fig. 1) Old implementation in blue, new implementation in red._

gm4_chairs/data/gm4_chairs/functions/sit_counter.mcfunction
```diff
@@ -1,5 +1,7 @@
-#run from advancement sitting_in_chair
+# Increase score for time sitting in a chair
 # @s = player sitting on the chair
+# at @s
+# run from advancement sitting_in_chair
 
 advancement revoke @s only gm4_chairs:sitting_in_chair
```
Bloo-dev, line 7: Using an advancement for this is actually quite bad for performance. Now that you have a predicate anyways you could just check the predicate on a _very_ slow clock.

Bloo-dev (commented):
I did not test these changes in game yet, but here's a full code review. I noted a couple of things that should be discussed before this is merged.

gm4_chairs/data/gm4_chairs/functions/lightning/tag.mcfunction
Dennis-0: It's possible for the stair block to be broken before the chair is returned, leaving a NoAI piglin. I'm not sure yet if the fire from the lightning can break a wooden stair block within that time, but it would definitely be possible to "abuse" it and create NoAI piglins that won't disappear

gm4_chairs/data/gm4_chairs/functions/create_chair.mcfunction
Dennis-0: The previous implementation summoned the pigs at the stair block, but that showed the pig for 1 tick (before invisibility is applied), so now it's spawned far below and teleported up instantly. So now I need 4 summon commands for direction checks and a single "teleport to stair block" command, so might as well put the summon command with all chair info in that single command instead of copying it 4 times.

gm4_chairs/data/gm4_chairs/functions/kill.mcfunction
Dennis-0: In what way could this be abused? This teleports the player exactly inside the stair block, so if there's a block below the stair block, it would be exactly on top of that block without falling through the ground

gm4_chairs/data/gm4_chairs/functions/sit_counter.mcfunction
Dennis-0: Would you suggest a separate function that's always running on a slow clock? (like every 10-30 seconds)

I was thinking of having this clock only run while a player is sitting on a chair, started by a player sitting on it (1.18 advancement trigger), but disadvantages of this is that relogging would stop this clock and not restart when logging back in

Dennis-0:
Added a slow clock (10s, could be slower if needed) for the time the player sits on a chair. 
Scoreboard value now increases by 10 each cycle, so the timer is still measured in seconds, otherwise players who sat in a chair in a previous version would have their time multiplied by 10 after this update.

gm4_chairs/data/gm4_chairs/functions/lightning/tag.mcfunction
Bloo-dev: That is true, but converting all `NoAI:1b` pigman into a chair can also create a lot of issues. Let's say you're setting all mobs in your world to `NoAI:1b` for debug reasons; now you'd have a bunch of chairs in your Pigman farm, that would end up killing themselves.
Therefore, I'd suggest revising this.

Bloo-dev:
I tested this in game and it works as expected. This will be merged before 2020/11/24, once the `NoAI` issue has been addressed.

gm4_chairs/data/gm4_chairs/functions/lightning/tag.mcfunction
Dennis-0: I guess that's a good reason. Added the stairs check!

PR #691, opened 2021-10-25
Vecto Shamir: Small improvement to vehicle check

Dennis-0:
In theory it should still work exactly the same as before, just some changes to the code

- Changed id to player's UUID value
- Improved the check for removing the vehicle, by switching around some selectors
- Comments to standard format


Bloo-dev (approved):
Note for future reviewers: Most of these changes are changes to header comments.

I've not tested this in game yet, but the code looks good.

PR #693, opened 2021-10-26
Apple Trees: Pick apples while holding any item

Dennis-0:
Currently apples can only be taken from apple trees while not holding any item, which can make this process quite annoying if you need to select a different slot for each apple.

This change allows you to take apples from apple trees while holding any item (except for apples).

#### Technical details:
- After picking the apple, it is moved to the "no apple" stage, instead of staying in the "ripe apple" stage. It still takes the same amount of time to start the next cycle, so it's only an internal change to make checks a bit more convenient.


Bloo-dev:
Tested this in game and it works as expected.

PR #696, opened 2021-11-18
Add lib_potion_tracking

BluePsychoRanger:
Library for tracking thrown potions

- tracks custom potions by giving it a scoreboard id then summoning a marker entity with a matching id
- the marker is repeatedly teleported to the matching potion
- when the marker no longer has a matching potion (i.e. the potion hit something), it will run a function tag

- updates Display Frames to use this library

****IMPORTANT****
- updates forceload library to have a repeating command block running a function tag
    - this is used to catch potions that are dispensed directly into a block
- requires forceload library; modules using lib_potion_tracking must include both forceload and potion_tracking in pack.mcmeta


gm4_display_frames/data/gm4_display_frames/predicates/invisibility_potion_marker.json
```diff
@@ -0,0 +1,26 @@
+{
+  "condition": "minecraft:alternative",
+  "terms": [
+    {
+      "condition": "minecraft:entity_properties",
+      "entity": "this",
+      "predicate": {
+        "nbt": "{data:{gm4_potion:{Item:{tag:{Potion:'minecraft:invisibility'}}}}}"
```
misode, line 8: This predicate doesn't help to optimize these checks. The NBT checks still need to happen and the entity data will be serialized 3 times. It will be faster to check storage or entity data with commands.

gm4_display_frames/data/gm4_display_frames/functions/potion_tracking/tag_potion.mcfunction
```diff
@@ -0,0 +1,4 @@
+# @s = any potion entity
+# run from #gm4_potion_tracking:tag_potion
+
+execute if predicate gm4_display_frames:invisibility_potion run tag @s add gm4_potion
```
misode, line 4: This is also a place where the library could copy the data to storage first, and then this function could have 3 commands checking that storage instead of an inefficient predicate.

gm4_display_frames/data/gm4_display_frames/functions/potion_tracking/verify_landed_potion.mcfunction
```diff
@@ -0,0 +1,4 @@
+# @s = any potion tracker marker that landed
+# run from #gm4_potion_tracking:potion_landed
+
+execute if predicate gm4_display_frames:invisibility_potion_marker run function gm4_display_frames:potion_tracking/splash_item_frames
```
misode, line 4: Same comment as in `tag_potion`. Checking storage will be more efficient.

misode (changes requested):
Looks good in general, but some weird choices regarding optimizing NBT checks. 

A predicate that checks the `entity_properties` for `nbt` is just as bad as a command doing the same check, if not worse. And having 3 of those checks is 3x worse.

gm4_display_frames/data/gm4_display_frames/functions/potion_tracking/tag_potion.mcfunction
BluePsychoRanger: Sorry, I didn't know that predicates were worst. The potion tracking library already puts the entire potion entity data into storage; I just thought the 3x checks would be better in an `alternative` predicate check, thus used the entity data instead of storage data. I'll make the changes to use the storage data.

gm4_display_frames/data/gm4_display_frames/functions/potion_tracking/tag_potion.mcfunction
```diff
@@ -1,4 +1,10 @@
 # @s = any potion entity
 # run from #gm4_potion_tracking:tag_potion
 
-execute if predicate gm4_display_frames:invisibility_potion run tag @s add gm4_potion
+data modify storage gm4_display_frames:temp Item.tag set from entity @s Item.tag
```
misode, line 4: Why does the library not copy to storage here, but it does for `potion_landed`?

lib_potion_tracking/data/gm4_potion_tracking-1.0/functions/tick.mcfunction
```diff
@@ -4,7 +4,8 @@
 # run from gm4_potion_tracking-1.0:resolve_tick
 
 # function tag so modules can tag their custom potions (should add the "gm4_potion" tag)
-execute as @e[type=potion,tag=!gm4_potion] run function #gm4_potion_tracking:tag_potion
+execute as @e[type=potion,tag=!gm4_potion_tracking_checked] run function #gm4_potion_tracking:tag_potion
+tag @e[type=potion] add gm4_potion_tracking_checked
```
misode, line 8: Instead of this second command, the library could add its own function in `#tag_potion` to add the tag (it doesn't matter in which order this happens).

misode (changes requested):

lib_forceload/data/gm4_forceload-1.0/functions/init_chunk.mcfunction
```diff
@@ -3,7 +3,8 @@
 forceload remove 29999999 7134
 forceload add 29999999 7134
 
-fill 29999998 0 7135 29999997 2 7131 minecraft:bedrock
+fill 29999998 0 7135 29999997 2 7130 minecraft:bedrock
 setblock 29999998 1 7134 minecraft:yellow_shulker_box
 setblock 29999998 1 7133 minecraft:birch_wall_sign[facing=north]
 setblock 29999998 1 7132 minecraft:jukebox{RecordItem:{id:"minecraft:stick",Count:1,tag:{}}}
+setblock 29999998 1 7131 minecraft:repeating_command_block{auto:1b,Command:"function #gm4_forceload:command_block_tick"}
```
misode, line 10: This places the command block in every dimension. However I don't think the current potion tracker `tick` function handles this correctly. Either filter the potions in the current dimension with `x=0`, or only place the repeating command block in the overworld.

Bloo-dev (changes requested):

misode (approved):

Bloo-dev (approved):

PR #710, opened 2021-11-29
Update Note Block Interface

Dennis-0:
No feature changes, mostly just moved the tune detection to an advancement trigger for `item_used_on_block`. This also triggers when placing something on the side of the note block, so the score check is still there, except it won't run each tick anymore.

Comments are also updated to the formatting standard.


Dennis-0:
I specifically told it to commit into dev/1.18 but still went into master...

PR #711, opened 2021-11-29
Note Block Interface Update

Dennis-0:
No feature changes, mostly just moved the tune detection to an advancement trigger for item_used_on_block. This also triggers when placing something on the side of the note block, so the score check is still there, except it won't run each tick anymore.

Comments are also updated to the formatting standard.


gm4_note_block_interface/data/gm4_note_block_interface/functions/interface/create.mcfunction
```diff
@@ -1,38 +1,41 @@
...
+execute if block ~ ~ ~ note_block[note=19] run summon area_effect_cloud ~ ~ ~ {Duration:20,Tags:["gm4_nbi_display"],CustomNameVisible:1,CustomName:'{"text":"C♯"}'}
+execute if block ~ ~ ~ note_block[note=20] run summon area_effect_cloud ~ ~ ~ {Duration:20,Tags:["gm4_nbi_display"],CustomNameVisible:1,CustomName:'{"text":"D"}'}
+execute if block ~ ~ ~ note_block[note=21] run summon area_effect_cloud ~ ~ ~ {Duration:20,Tags:["gm4_nbi_display"],CustomNameVisible:1,CustomName:'{"text":"D♯"}'}
+execute if block ~ ~ ~ note_block[note=22] run summon area_effect_cloud ~ ~ ~ {Duration:20,Tags:["gm4_nbi_display"],CustomNameVisible:1,CustomName:'{"text":"E"}'}
+execute if block ~ ~ ~ note_block[note=23] run summon area_effect_cloud ~ ~ ~ {Duration:20,Tags:["gm4_nbi_display"],CustomNameVisible:1,CustomName:'{"text":"F"}'}
+execute if block ~ ~ ~ note_block[note=24] run summon area_effect_cloud ~ ~ ~ {Duration:20,Tags:["gm4_nbi_display"],CustomNameVisible:1,CustomName:'{"text":"F♯"}'}
+
+# updates note block with block under
 execute unless block ~ ~-1 ~ #gm4_note_block_interface:skulls run clone ~ ~-1 ~ ~ ~-1 ~ ~ ~-1 ~ replace force
```
Dennis-0, line 39: I wasn't completely sure if this line is still needed. When I removed it, it still worked fine, so was thinking about removing it.

EpyonProjects: Are those notes possible?

Dennis-0: yes there's 25 notes, but I mean the clone command

BluePsychoRanger:
If you add a condition check for `is_sneaking=false`, it will only detect when you actually play the note block (applies to both hitting and tuning the note block). This is how it's done for the guidebook (`gm4_guidebook/advancements/note_block_interface/page_0.json`)

gm4_note_block_interface/data/gm4_note_block_interface/functions/interface/create.mcfunction
BluePsychoRanger: This is used to update the note block when backwards tuning. Not sure if it's still needed, cuz it might've been to update it visually for when players have F3 up.

BluePsychoRanger: It appears this is not necessary, since it works the same if a skull is under the note block. This command is also responsible for the headless pistons thing

Dennis-0:
> If you add a condition check for `is_sneaking=false`, it will only detect when you actually play the note block (applies to both hitting and tuning the note block, which may or may not be what you want). This is how it's done for the guidebook (`gm4_guidebook/advancements/note_block_interface/page_0.json`)

Do you mean separate advancement triggers for increase/decrease note? Otherwise it wouldn't detect when sneaking

BluePsychoRanger:
> Do you mean separate advancement triggers for increase/decrease note? Otherwise it wouldn't detect when sneaking

Oh true, nvm then

SpecialBuilder32 (changes requested):
Tested and it all looks good. 

The removal of the clone command should also remove the skulls block tag, since it doesn't appear to be used anywhere else.

misode (approved):
Tested and works fine. LGTM

Bloo-dev (approved):
I did a full code review and could not find any issues that would prevent this from being merged.

PR #714, opened 2021-12-02
Mob Curing 1.18 (renamed Mob Conversion)

EpyonProjects:
- Potion Cleric updated to offer a random assortment from all potions, requires lib_trades
  - All vanilla effects (including Luck, if Zauber is installed)
  - Lightning in a Bottle, Potion Liquids, and Zauber Cauldrons if installed
- Added Pillager <- Evoker <- Illusioner curing, if LiaB is installed
- Added Piglin <- Zombified Piglin, if a Weakness potion is used
  
- General refactoring
  - Folders are nice
- Updated Guidebook pages


Bloo-dev:
Luck should probably only be included if Zauber Cauldrons is installed, as luck has very little uses without.

Bloo-dev:
Suggested edit:
```diff
diff --git a/gm4_mob_curing/data/gm4_mob_curing/functions/cure/consume_item.mcfunction b/gm4_mob_curing/data/gm4_mob_curing/functions/cure/consume_item.mcfunction
index 5febf786..564c6648 100644
--- a/gm4_mob_curing/data/gm4_mob_curing/functions/cure/consume_item.mcfunction
+++ b/gm4_mob_curing/data/gm4_mob_curing/functions/cure/consume_item.mcfunction
@@ -2,14 +2,14 @@
 # run from items in convert files
 
 # store count in scoreboard
-execute store result score @s gm4_mob_cu_data run data get entity @s Item.Count
+execute store result score @s gm4_mob_curing_data run data get entity @s Item.Count
 # remove 1
-scoreboard players remove @s gm4_mob_cu_data 1
+scoreboard players remove @s gm4_mob_curing_data 1
 # store back into nbt
-execute store result entity @s Item.Count byte 1 run scoreboard players get @s gm4_mob_cu_data
+execute store result entity @s Item.Count byte 1 run scoreboard players get @s gm4_mob_curing_data
 
 # Spigot fix
-kill @s[scores={gm4_mob_cu_data=0}]
+kill @s[scores={gm4_mob_curing_data=0}]
 
 # remove the tag
 tag @s remove gm4_gapple
diff --git a/gm4_mob_curing/data/gm4_mob_curing/functions/init.mcfunction b/gm4_mob_curing/data/gm4_mob_curing/functions/init.mcfunction
index 359797ae..622cc1b3 100644
--- a/gm4_mob_curing/data/gm4_mob_curing/functions/init.mcfunction
+++ b/gm4_mob_curing/data/gm4_mob_curing/functions/init.mcfunction
@@ -1,4 +1,4 @@
-scoreboard objectives add gm4_mob_cu_data dummy
+scoreboard objectives add gm4_mob_curing_data dummy
 
 execute unless score mob_curing gm4_modules matches 1 run data modify storage gm4:log queue append value {type:"install",module:"Mob Curing"}
 scoreboard players set mob_curing gm4_modules 1
diff --git a/gm4_mob_curing/data/gm4_mob_curing/functions/potion_cleric/discount_curer.mcfunction b/gm4_mob_curing/data/gm4_mob_curing/functions/potion_cleric/discount_curer.mcfunction
index 6d38c435..648b116e 100644
--- a/gm4_mob_curing/data/gm4_mob_curing/functions/potion_cleric/discount_curer.mcfunction
+++ b/gm4_mob_curing/data/gm4_mob_curing/functions/potion_cleric/discount_curer.mcfunction
@@ -6,8 +6,8 @@ data merge storage gm4_mob_curing:temp {Gossips:[{Type:"minor_positive",Value:20
 data modify storage gm4_mob_curing:temp Gossips[0].Target set from entity @e[type=minecraft:item,limit=1,distance=..3,tag=gm4_gapple] Thrower
 data modify entity @s Gossips set from storage gm4:temp Gossips
 
-execute store result score $curer gm4_mob_cu_data run data get storage gm4_mob_curing:temp Gossips[0].Target[3] 0.001
-execute as @a[distance=..10] store result score @s gm4_mob_cu_data run data get entity @s UUID[3] 0.001
-execute as @a[distance=..10] if score @s gm4_mob_cu_data = $curer gm4_mob_cu_data run advancement grant @s only gm4:mob_curing
+execute store result score $curer gm4_mob_curing_data run data get storage gm4_mob_curing:temp Gossips[0].Target[3] 0.001
+execute as @a[distance=..10] store result score @s gm4_mob_curing_data run data get entity @s UUID[3] 0.001
+execute as @a[distance=..10] if score @s gm4_mob_curing_data = $curer gm4_mob_curing_data run advancement grant @s only gm4:mob_curing
 
 data remove storage gm4_mob_curing:temp {}

```

gm4_mob_curing/data/gm4_mob_curing/functions/init.mcfunction
```diff
@@ -0,0 +1,11 @@
+scoreboard objectives add gm4_mob_cu_data dummy
```
Bloo-dev, line 1:
```suggestion
scoreboard objectives add gm4_mob_curing_data dummy
```
As of Minecraft 1.18 there is no longer a length limit for objective names. Note that this change will have to be made in multiple places if you accept it.

gm4_mob_curing/data/gm4_mob_curing/functions/main.mcfunction
```diff
@@ -0,0 +1,9 @@
+#execute on red mushrooms to check for cows around them
+execute at @e[type=item,nbt={Item:{id:"minecraft:red_mushroom"}},limit=1,sort=random] if entity @e[type=item,distance=..1.5,limit=1,nbt={Item:{id:"minecraft:brown_mushroom"}}] as @e[type=cow,distance=..1.5,limit=1,nbt={ActiveEffects:[{Id:2b}]}] unless data entity @s NoAI at @s run function gm4_mob_curing:cure/mooshroom
+execute at @e[type=item,nbt={Item:{id:"minecraft:golden_carrot"}},limit=1,sort=random] as @e[type=zombified_piglin,distance=..1.5,limit=1] unless data entity @s NoAI at @s run function gm4_mob_curing:cure/zombified_check
+execute at @e[type=item,nbt={Item:{id:"minecraft:golden_carrot"}},limit=1,sort=random] as @e[type=zoglin,distance=..1.5,limit=1,nbt={ActiveEffects:[{Id:1b}]}] unless data entity @s NoAI at @s run function gm4_mob_curing:cure/zoglin
+execute at @e[type=item,nbt={Item:{id:"minecraft:golden_apple"}},limit=1,sort=random] as @e[type=witch,distance=..1.5,limit=1,nbt={ActiveEffects:[{Id:18b}]}] unless data entity @s NoAI at @s run function gm4_mob_curing:cure/witch
```
Bloo-dev, line 5: You're searching through all items entities three times here. A single `execute at @e[type=item...] run function...` in combination with `@s` checks inside said function would definitely be beneficial. This would also remove the double lookup for `@e[type=item,nbt={Item:{id:"minecraft:golden_carrot"}},limit=1,sort=random]`.

Furthermore, `limit=1` makes it so that only one cow can be converted per clock pulse (globally!); This seems unintentional.

gm4_mob_curing/data/gm4_mob_curing/functions/main.mcfunction
```diff
@@ -0,0 +1,9 @@
+#execute on red mushrooms to check for cows around them
```
Bloo-dev, line 1: Outdated comment.

gm4_mob_curing/data/gm4_mob_curing/functions/cure/mooshroom.mcfunction
```diff
@@ -1,8 +1,9 @@
...
+# run from main
+
 summon mooshroom
 particle explosion ~ ~.5 ~ 1 .3 1 0 1
 playsound minecraft:entity.zombie_villager.converted neutral @a[distance=..8]
-execute as @e[type=item,limit=1,nbt={Item:{id:"minecraft:red_mushroom"}},distance=..3,sort=nearest] run function gm4_mob_conversion:consume_item
-execute as @e[type=item,limit=1,nbt={Item:{id:"minecraft:brown_mushroom"}},distance=..3,sort=nearest] run function gm4_mob_conversion:consume_item
+execute as @e[type=item,limit=1,nbt={Item:{id:"minecraft:red_mushroom"}},distance=..3,sort=nearest] run function gm4_mob_curing:cure/consume_item
+execute as @e[type=item,limit=1,nbt={Item:{id:"minecraft:brown_mushroom"}},distance=..3,sort=nearest] run function gm4_mob_curing:cure/consume_item
 data merge entity @s {DeathTime:19s,Health:0f,DeathLootTable:"minecraft:empty"}
```
Bloo-dev, line 9: Are you sure this still works? I remember something about this being broken in 1.18 -- but I could be wrong.

_This comment also applies to the other curing flavors._

gm4_mob_curing/data/gm4_mob_curing/functions/cure/mooshroom.mcfunction
```diff
@@ -1,8 +1,9 @@
-#@s = cows with slowness needing conversion at @s
-#run from main
+# @s = cows with slowness needing conversion at @s
+# run from main
+
 summon mooshroom
```
Bloo-dev, line 4: This freshly summoned mob will face south by default, which will look slightly disappointing visually. Try to teleport your new mob to the old one to prevent this.

Also, I'm not sure how vanilla's curing handles NBT, but do we want to transfer things like `CustomName`?

_This comment also applies to the other curing flavors._

gm4_mob_curing/data/gm4_mob_curing/functions/potion_cleric/discount_curer.mcfunction
```diff
@@ -0,0 +1,13 @@
...
+# run from initialise_potion_cleric
+
+# Builds the gossips array on storage to avoid a bug with modifying gossips
+data merge storage gm4_mob_curing:temp {Gossips:[{Type:"minor_positive",Value:20}]}
+data modify storage gm4_mob_curing:temp Gossips[0].Target set from entity @e[type=minecraft:item,limit=1,distance=..3,tag=gm4_gapple] Thrower
+data modify entity @s Gossips set from storage gm4:temp Gossips
+
+execute store result score $curer gm4_mob_cu_data run data get storage gm4_mob_curing:temp Gossips[0].Target[3] 0.001
+execute as @a[distance=..10] store result score @s gm4_mob_cu_data run data get entity @s UUID[3] 0.001
+execute as @a[distance=..10] if score @s gm4_mob_cu_data = $curer gm4_mob_cu_data run advancement grant @s only gm4:mob_curing
```
Bloo-dev, line 11: This seems hacky. A function call to avoid the duplicate `@a` wouldn't give much of an improvement here, but why limit it to 10 blocks range in the first place? Shouldn't the curer also get that advancement if they dropped the item from a height of more than 10 blocks?

gm4_mob_curing/data/gm4_mob_curing/functions/cure/liab/check.mcfunction
```diff
@@ -0,0 +1,5 @@
+# Checks Lightning in a Bottle-specific entity reverting
+# run from main
+
+execute at @e[type=item,nbt={Item:{id:"minecraft:totem_of_undying"}},limit=1,sort=random] as @e[type=illusioner,distance=..1.5,limit=1,nbt={ActiveEffects:[{Id:1b}]}] unless data entity @s NoAI at @s run function gm4_mob_curing:cure/liab/illusioner
+execute at @e[type=item,nbt={Item:{id:"minecraft:crossbow"}},limit=1,sort=random] as @e[type=evoker,distance=..1.5,limit=1,nbt={ActiveEffects:[{Id:2b,Amplifier:3b}]}] unless data entity @s NoAI at @s run function gm4_mob_curing:cure/liab/evoker
```
Bloo-dev, line 5: These could also be moved into the aforementioned single function file that uses `@s` to analyze the item further.

Bloo-dev (changes requested):
I did a full code review of this. Most of it looks ok, there is a performance issue in `main`, though.

gm4_mob_curing/data/gm4_mob_curing/functions/potion_cleric/discount_curer.mcfunction
EpyonProjects: How would you find and grant the advancement to the thrower?  This fake curing is not defined under the typical Zombie Doctor curing advancement, and you cannot select an entity based on a score value.  You also cannot advancement grant <score value>, so even though I have the UUID of the player, it doesn't seem like I can access them any other way.  The distance check is for lag reduction.

gm4_mob_curing/data/gm4_mob_curing/functions/main.mcfunction
EpyonProjects: The cow limiter is only per `@e[item`, no?

gm4_mob_curing/data/gm4_mob_curing/functions/cure/mooshroom.mcfunction
EpyonProjects: This is regarding line 8, the data merge entity.   This still functions to kill, however no longer functions to ignore the death animation and cloud particle effect.  This will be changed to a off-screen teleport and kill @s

gm4_mob_curing/data/gm4_mob_curing/functions/cure/mooshroom.mcfunction
EpyonProjects: Comparing against lightning and pig->piglin, it summons a new entity.  This copies that functionality, though I'll take the suggestion of teleporting in-place and copying all data, to match the way Shroomites does it.

gm4_mob_curing/data/gm4_mob_curing/functions/main.mcfunction
Bloo-dev: This is `main.mcfunction`. Each of these lines will only act as a single item per clock pulse.

gm4_mob_curing/data/gm4_mob_curing/functions/potion_cleric/discount_curer.mcfunction
Bloo-dev: The item has a NB tag which stores the UUID of the thrower.

gm4_mob_curing/data/gm4_mob_curing/functions/cure/liab/illusioner.mcfunction
```diff
@@ -0,0 +1,16 @@
+# Reverts the Illusioner to an Evoker
```
sparks0, line 1: Are Illusioners available in survival? I thought they were unused except maybe by mysterious midnights? Why would players want to convert an illusioner into an evoker if they could find one, for the expensive cost of a totem? Same goes for why would players want to convert an evoker into a pillager?

sparks0:
How will the site/git react to a module being renamed? I think it'll count as a new module, so I suppose we'd remove the 1.18 version of mob conversion? It means things like in-game update messages for the module won't tell you to upgrade to the new one and they'll have separate pages on the site without a download history for older versions.

Bloo-dev:
> How will the site/git react to a module being renamed? I think it'll count as a new module, [...]

Fairly sure you are correct about this. I think @misode and me discussed what happens to renames when we did the website re-write. Iirc it might be possible to hard-code a special case for this, though we might want to consider adding a field to the `pack.mcmeta` that holds old names of the module as this might happen to more modules as features change.

I'd suggest something like
```
"predecessors":[
                          "old_module_id",
                          "very_old_module_id"
]
```
Not sure how easily this can be included in the new beet-based setup, though.

misode:
We have the problem right now with `lightning rods` -> `lightning in a bottle`, to keep it simple, they're treated as separate modules on the site.

@Bloo-dev's suggestion of storing module ID renames in the `pack.mcmeta` sounds good, and would only require a small (probably one line) change in the beet build script. The changes on the site will be more thorough and require some thought. For example when on the 1.18 `mob-curing` page it should change the URL to `mob-conversion`, and might need to re-fetch some data.

PR #715, opened 2021-12-06
Remove flowering azalea blocks from block tags

Dennis-0:
Replaced `#minecraft:flowers` with `#minecraft:small_flowers` and `#minecraft:tall_flowers`, because `flowering_azalea` and `flowering_azalea_leaves` are part of the `#minecraft:flowers` block tag, which was included in `#gm4:no_collision` and `#gm4:foliage`.


misode (approved):
Looks good! I don't think this needs further review

PR #716, opened 2021-12-10
Orbis 1.18 part 2

misode:
More improvements to orbis, specifically the structures.

Use loot tables to decide a random structure (instead of checking for bedrock patterns). This has the unfortunate side effect of losing seed determinism. But I don't think that was a very important feature and it made it a lot easier to add a bunch of structure variants.

Update the height dependent chest loot to match the 1.18 terrain.


misode:
This version is roughly the same as the one we used on the public server. Merging as it's unlikely to get further testing and the released version has some problems

PR #717, opened 2021-12-17
Guidebook - Initial Lag Fix

BluePsychoRanger:
This PR reduces the massive lag spike from happening upon reloading with the guidebook module installed


Bloo-dev:
Irrelevant for the public repository. Guide Books are being reworked entirely.

PR #719, opened 2021-12-20
Ink Spitting Squid Disabled by Conduit Power

BluePsychoRanger:
Squids will no longer give affect players with the conduit power effect.

This is to improve the experience of underwater building, as many module encourage deep sea builds, but Ink Spitting Squid makes it very annoying to do so.


misode:
Seems reasonable. We can put this on server testing.

Bloo-dev:
Seems like a good change. To be put into server testing.

SpecialBuilder32 (approved):
Tested in single-player and could find no issues. I think this is a good change.

PR #725, opened 2021-12-29
Better Armour Stands: Lock, Undo, General improvements

Dennis-0:
### Feature changes
- Locking the armour stand:
  - Add `lock`: Disable interactions with the armour stand (removed hitbox)
  - Add `unlock`: Re-enable interactions with the armour stand. 
  Holding a book with `unlock` temporarily enables interactions nearby locked armour stands, to make it possible to right click it with this book to unlock it. These "temporarily unlocked" armour stands will glow to show which can be unlocked.
- Add undo ability: Stop holding the book and quill to revert its pose/position/rotation. This is also triggered when the player gets too far away the armour stand. The previous data is stored in the marker in front of the player when the player starts editing the armour stand. 
- `move` now "grabs" the armour stand smoothly, without hardcoded distance checks
- Improved interactions with armour stand without arms:
  - Select the correct armour stand more accurately, and less required steps in the ray cast
  - If arms are re-enabled, other codes in the book are now also checked
- Switched behaviour of `default`:
  - Not sneaking will fully reset the armour stand, and run the "default" functions in expansion packs
  - Sneaking will reset the pose of the part you're looking at
- Removed randomness to head/body pose when placing a copied armour stand (if stacksize > 1) 
- Barrier particle when no valid code is given is only shown to player using the book
- Updated guidebook pages: Each code shows information on hover, and clicking the code will copy it to your clipboard.
 
### Internal changes
- Clean up folder structure
- Removed ray cast by selecting an armour stand holding non-empty book and quill, and by adding a tag to the book if given with `equip`. This way it should be impossible for multiple armour stands to hold a non-tagged book and quill (unless spawned in with commands)
- Removed "use book while tracking" advancement trigger and tracking tag checks for using the book:
  - Armour stands can't be given items anymore while it's being edited (DisabledSlots)
  - Player can't use the book on another armour stand while editing another armour stand (advancement trigger for this is only revoked after player is done editing)
- Improved book handling for players in creative mode (no longer deletes the book if the armour stand was holding a book and quill already)
- Moved checks in main to separate loops that are only run while the checks are needed
- Move "broken armour stand" detection from main to tick, that's only active while an armour stand is being active
- Player id set from `UUID[0]` (for consistency with other datapacks)
- Changed villager (in interact detection for armour stands without arms) to wandering trader
- Entities killed in the void are teleported 10000 blocks down instead of 1000 (for minimum world height in a custom world)
- Replaced animal entities for "looking_at" detection with AEC (disappears instantly)
- Replaced `@p[tag=gm4_bas_active]` with `@a[tag=gm4_bas_active,limit=1]`

## Particles Pack
- Add "default" function for resetting the armour stand
- Instead of resetting the particle score and particle clouds and readding it after the checks, use a "success" score to check if code was successfully applied
- Edit guidebook pages, added "click to copy" on hover over each code

## Poses Pack
- Player selectors now use the same arguments as similar selectors in BAS
- Edit guidebook pages, added "click to copy" on hover over each code


Bloo-dev (commented):
Why did you change your `@p` to `@a`? This does not affect functionality or performance (unless you want to deal with dead players) but just makes your code larger and more complicated.

Dennis-0:
> Why did you change your `@p` to `@a`? This does not affect functionality or performance (unless you want to deal with dead players) but just makes your code larger and more complicated.

I was told `@a[limit=1]` is the same as `@p` in this case, without the distance calculations, and since there's only 1 player with this tag, no distances are needed

Bloo-dev:
Update to my previous comment:
This does make _a difference_, but I don't know if it's what you intended. You now select entities by `oldest first` instead of `closest first`.

Dennis-0:
> Update to my previous comment: This does make _a difference_, but I don't know if it's what you intended. You now select entities by `oldest first` instead of `closest first`.

Since only 1 player can exist at any time with this tag, `oldest first`, `closest first`, or another sorter, is all the same

Dennis-0:
I kept finding new things to add/fix to this "small update", but I should be done now

misode (changes requested):
I tried to break it and found two bugs so far, unsure if they were introduced in this PR though.

1. Logging out while moving an armor stand (in singleplayer) leaves the wandering trader and marker behind, with a particle effect. This may affect more than `move`
2. Walking away while moving an armor stand so the original position is unloaded and then undoing teleports the armorstand away correctly, but its slots will be disabled.

Dennis-0:
> I tried to break it and found two bugs so far, unsure if they were introduced in this PR though.
> 
> 1. Logging out while moving an armor stand (in singleplayer) leaves the wandering trader and marker behind, with a particle effect. This may affect more than `move`
> 2. Walking away while moving an armor stand so the original position is unloaded and then undoing teleports the armorstand away correctly, but its slots will be disabled.

Good catch, that's indeed because this PR, because the "unnecessary" checks in main that resets tags etc. were removed. (as a backup in case it still failed, it would still be fixed once the BAS functions are run again)

1. Didn't expect this to happen, but I guess the "tick" function from editing runs before the player and armour stand are loaded? (Would fix itself after giving the book to another armour stand)
What particle effect was this? I didn't see that.

2. Makes sense, since the position is reset before the armour stand tracking is cleared (Would also fix itself when BAS is used again)

I'll have a look at fixing these bugs soon

misode (approved):
Tested again and the second issue is fixed, however I could still reproduce the first issue (logging out while editing). It does indeed fix itself after editing another armor stands. This is acceptable imo since it's not exploitable and it reduces main checks.

Bloo-dev (approved):
I've had a quick look over the code and couldn't find any blocking issues. Clear to merge.

PR #729, opened 2022-01-02
Prevents Desire Lines from breaking moss carpet

EpyonProjects:
This PR completely prevents Desire Lines from breaking moss carpet by adding an explicit exception for the block.


misode (approved):
Makes sense since it has a bounding box. The other two cases are `azalea` and `flowering_azalea`, but I'm not sure if those are even possible to happen with desire lines.

SpecialBuilder32 (approved):
Explicitly excluding moss carpet is a better solution that editing the entire block tag, I like it.

PR #731, opened 2022-01-02
Make Ender Bolt ignore pets

Bloo-dev:
Ender Bolt can be pretty dangerous to use, especially when you own a dog or a horse. This PR makes your pets immune to your Ender Bolt by implementing an ID system. 

This also has the nice side effect of Ender Bolt no longer being able to infect you once you deselect your sword, as each instance of infection now carries a unique ID of the source and ignores entities with a matching `UUID` or `Owner` nbt. Servers that are coop-focused (e.g. Gamemode 4's Obsolescence Map) can toggle `$enable_pvp` in `gm4_ender_bolt_foreign_id` to `0` (default `1`) to stop Ender Bolt from infecting other players or other players' pets.


SpecialBuilder32 (commented):
Tested, and it works as it says on the tin. 

One thing I did find, though without being able to track down the source of the problem, is that ender-bolt tridents seem to no longer spread infection.

PR #733, opened 2022-01-03
Reduce the time required for sneak charge pneumas

Levertion:
Without this change, it takes:
- 2 seconds for level 1
- 4 seconds for level 2
- 8 seconds for level 3
- 16 seconds for level 4

Compare e.g. to gliding, which gives dolphins grace 2 all the time, this makes e.g. rushing significantly less useful for travel, as a quarter of the time would be spent sneaking (it might turn out that level 2 is most efficient for travel, but it's non-trivial to calculate that)

After this change:
- 1 second for level 1
- 2 seconds for level 2
- 4 seconds for level 3
- 6 seconds for level 4

Note that I haven't touched draining, since that is based on its own seperate numbers, which is 3 ticks for level 1, 6 ticks for level 2, 9 ticks for level 3 and 14 ticks for level 4; presumably because no-one is going to use draining anyway, so it doesn't matter how 'OP' it is.

### DIFF
--- a/gm4_orb_of_ankou/data/gm4_orb_of_ankou/functions/pneumas/bounding.mcfunction
+++ b/gm4_orb_of_ankou/data/gm4_orb_of_ankou/functions/pneumas/bounding.mcfunction
@@ -1,7 +1,7 @@
 # @s = player with bounding pneuma who stopped sneaking
 # run from pneumas/sneak/stopped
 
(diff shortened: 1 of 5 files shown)


Bloo-dev:
This is part of the Orb of Ankou 1.19 update now.

PR #740, opened 2022-01-10
add glow squid below y30

TheThanathor:
* Squid spawns below y30 are replaced with glow squid


Bloo-dev:
This should probably only work on gravel, as sand wouldn't generate within the Glowsquid's natural habitat.

Bloo-dev:
> This should probably only work on gravel, as sand wouldn't generate within the Glowsquid's natural habitat.

There's been some discussion about this on Discord, and people have convinced me that Glow Squid from sand is OK to have.

Bloo-dev (approved):

PR #741, opened 2022-01-13
Improve Everstone Aging

BluePsychoRanger:
- mobs now age at the same rate regardless of how many mobs are loaded in the area
- buffed mob aging: mobs age quicker since the specified times are on average, not at minimum (e.g. guardian -> elder guardian is 2 hrs on average instead of 2hrs minimum)
- when a mob is converted via aging, ALL data is transferred over except tags
---
- guardians will no longer age if there is an elder guardian within a 32 block radius
---
- smithed compatibility
- refactored code for optimization and readability
- added function tags for age locking and aging (expansion compatibility)


BluePsychoRanger:
Note: this was not tested in game

misode (approved):
After testing this in-game and leaving it open for about 3 hours I can confirm mobs are aging correctly. However (it might just be randomness), but I had to wait 3 hours until one guardian aged while the wiki mentions 20 minutes. Then I looked at the code and saw it was changed to approximately 2 hours.

Due to how time consuming and random this module is, the best way to test it will be to put it on the server.

image

Bloo-dev:
This has been running on the public server for about 2.5 months now and no issues have been reported. I'll mark this as tested.

PR #744, opened 2022-01-20
Lib machines

BluePsychoRanger:
Machine Blocks Update

- [x] Custom Crafters bundled in lib_machines (now a library)
    - [x] Custom Crafters has a new format (load.status = 2)
    - [x] Custom Crafter recipe outputs are handled via loot table
- [x] All machines are craftable
    - [x] All machines have new (simpler) crafting recipes
    - [x] Crafting recipe now gives an item instead of modifying the block
- [x] Machine blocks can be placed down from an item and be picked up by mining the block
- [x] Machine blocks now have a different CMD depending on rotation, so they can be textured differently (1 CMD for facing up, 1 CMD for facing down, and 1 CMD for facing horizontally; the armor stand gets properly rotated based on the rotation of the block)
- [x] Machine blocks have a special container name that supports a custom GUI via resource pack (complex translation and font magic stuff)
- [x] Every machine block entity has an “gm4_entity_version” score for future updates
---
- [x] Enchantment Extractors now take an item inside it instead of dropping on top of it
- [x] Compressors and Tinkering Compressors now have a “pressure plate” on top which gets depressed when any entity is on top of it
- [x] Compressors and Tinkering Compressors have a single facing direction (down) when placed
- [x] Tinkering Compressors are created by smooshing a compressor item and tinker shamir item (similar to normal metallurgy smooshing); unsmooshing works similarly, using obsidian and the tinkering compressor
- [x] Smelteries are now the Furnace instead of the hopper
    - [x] The items are doubled directly in the output slot
    - [x] The multi block now requires a cauldron instead of a hopper (existing multi block structures will update with an empty cauldron)
    - [x] Old (hopper) smelteries without a multi block structure will convert to a furnace smeltery
    - [x] Each smeltery smelt will have a 1.1% chance to fill the cauldron with lava
- [x] Liquid Minecarts and Ender Hopper Minecarts are crafted in a custom crafter instead of in a hopper minecart
- [x] Tunnel Bores uses lib_machines to detect destruction instead of separate clocking
- [x] Soul Glass and Trapped Signs use lib_machines to detect destruction and give back the actual item instead of items used to craft them
---
- [x] Relocators changed to be able to pick up any container block, saving NBT
- [x] Relocators properly works with the new machines
---
- [x] Advancements for machine blocks (also Liquid Minecarts and Tunnel Bores) updated to inventory changed for obtaining the item
- [ ] Guidebook pages updated to reflect these changes


BluePsychoRanger:
Warning: This is untested, so don't install this on any world you actually care about

Bloo-dev:
I played around with this in-game and tried to break it (I did not do a code analysis!), but couldn't.
However, at the point of writing this comment this PR is still a draft/being worked on, so don't take this as final approval.

Bloo-dev:
We've installed this on a mirror of the current public server and the upgrade paths seem to work. However, we have noticed that Zauber Liquids is not yet fully supported by relocators -- i.e. wormholes lose their data. However, this can be addressed in another PR once this has been server tested.

BluePsychoRanger:
> So in attempting to do some testing on this module, I ran into some strange behavior with installing it that I'd like to bring up here. I don't think the load stuff is configured properly, but I'm not sure about that so I can't fix it myself.
> 
> What I attempted to do: Debug the liquid-tanks upgrade path (determined in voicecall to cause problems). First I created a new world and installed the liquid tanks from the website, made a few tanks and contraptions, and then installed the `lib_machines`, base and gm4_liquid_tanks from this PR, and deleted the old gm4_liquid_tanks.zip from the website.
> 
> This fails to load, and prints to chat that LT is disabled because its missing CC. Since CC is included in `lib_machines` this confused me. Tracking the error deeper reveals that I needed `lib_forceload` as well, but no message logging said error made it to chat.
> 
> I do not know how LanternLoad should be used in the case with the enumerate loading methods, but I suspect that an entry should have been made in the error log within both enumerate functions if their dependency (forceload for machines, and machines for CC respectively, though the second would never show up) to notify of this error.

You have to remember that the error log is a GM4 thing, NOT a LanternLoad thing. LL just prevents things from loading if their dependencies do not properly load. Since lib_machines and lib_custom_crafters are libraries, we do not have any code in them to deal with the error logging. However this will never be an issue assuming the data packs are packaged by beet or when using Github, the user (AKA you) installs all GM4 libraries as a precaution (or just check the `pack.mcmeta` for which libraries are required).

gm4_block_compressors/data/gm4_block_compressors/functions/machine/create.mcfunction
```diff
@@ -0,0 +1,18 @@
...
...u00a0Amount","font":"gm4:offscreen"},{"translate":"gui.gm4.block_compressor","font":"gm4:container_gui","color":"white"},{"text":"Compression\\u00a0Amount","font":"gm4:half_scale"},{"text":"Compression\\u00a0Amount","font":"gm4:inverted"},{"text":"Compression\\u00a0Amount","font":"gm4:inverted_spacing"},{"text":"Compression Amount","font":"minecraft:default","color":"#373737"}]]},{"translate":"%1$s%3427656$s","with":[{"translate":"container.gm4.block_compressor","font":"minecraft:default","color":"#373737"},[{"translate":"container.gm4.block_compressor","font":"gm4:half_scale"},{"translate":"container.gm4.block_compressor","font":"gm4:inverted"},{"translate":"container.gm4.block_compressor","font":"gm4:inverted_spacing"},{"translate":"container.gm4.block_compressor","font":"gm4:offscreen"},{"translate":"gui.gm4.block_compressor","font":"gm4:container_gui","color":"white"},{"translate":"container.gm4.block_compressor","font":"gm4:half_scale"},{"translate":"container.gm4.block_compressor","font":"gm4:inverted"},{"translate":"container.gm4.block_compressor","font":"gm4:inverted_spacing"},{"translate":"container.gm4.block_compressor","font":"minecraft:default","color":"#373737"}]]}]}'}
```
SpecialBuilder32, line 7: Beeps, would you mind explaining briefly how the custom-font stuff here works? From the laymans perspective it seems like theres a lot of redundant data here.

BluePsychoRanger: The redundant data is there in order to ensure the name and GUI remains centered. Basically the way the custom font stuff is working in this instance is that it's printing the name with half the normal width, then prints it in reverse then prints the name. That all happens with an offset so the player doesn't see it. Then it prints the string that actually shows the GUI texture. Then it once again prints the name with half width and then in reverse. Then it prints the actual name of the container that the players see. Then it does that all again with the custom namespace which means the translation shows up instead of the English set name. 

TL;DR the reason it prints the name a bunch of times is to keep it centered. If you look at one of the hopper names (e.g. ender hoppers) you will see that it is much simpler.

misode (approved):

Bloo-dev (approved):

PR #745, opened 2022-01-31
Beet Credits Management

Bloo-dev:
### Overview
Allows Beet to automatically check credits contained in `{moduleId}/pack.mcmeta` against a list of centrally stored credits in `contributors.json` and updates them upon module building. This will result in credits being unified across the site, so people no longer have to track down credits for each module, however, the old-fashioned pack.mcmeta-only way of doing credits will stay functional.

### Technical Details
`contributors.json` stores an array of json objects -- so called contributors -- which does not match the credit format in `pack.mcmeta`; However, this new format is much more versatile and fits our use case better. 
~~Therefore this also contains an updater which will update credits from the old format if necessary. This updating does only occur if the credits actually change, so this won't cause module updates just for a format change.
This also means that the website will have to be able to deal with both types of credits. Let me know if we should force-push a credit format update.~~ 
**UPDATE:** We're using this opportunity to update the format. Misode used a script to update all mcmetas.

Credits within `pack.mcmeta` are replaced with credits from `contributors.json` if:
1. An entry with a matching name exists in `contributors.json`
2. The list of links from `contributors.json` does NOT match the list of links from `pack.mcmeta`


misode:
> however, the old-fashioned pack.mcmeta-only way of doing credits will stay functional.

Got rid of this requirement. It seems like a good opportunity to clean things up. Like mentioned in a commit message, I used a one-time use [script](https://gist.github.com/misode/0f5bd4fa624f8210a51ee44e17e55b8c) to transform the pack.mcmeta files. This will make it easier running this on old version branches.

I also took the opportunity to add `recommends`, `wiki_link` and `video_link` to the meta file so we no longer need to fetch pack.mcmeta on the site.

Bloo-dev:
I've re-implemented the "Array - Object" based representation of credits to enforce a slightly more strict JSON style. Internally this format is still converted to a `dict` which is keyed by name, to ensure `O(n)`.

misode (approved):

PR #746, opened 2022-02-01
Fix spawner minecarts not accepting vanilla spawners without SpawnPotentials

misode:


misode:
This is the same fix that was applied on the server and it's been working correctly there.

PR #747, opened 2022-02-03
[GitHub Actions] Upload module artifact for pull requests

misode:
This PR makes it so pull requests build the modules and upload them as artifacts, which can be downloaded from the GitHub website.

As an example, I created a [pull request](https://github.com/misode/GM4_Beet/pull/2) and by navigating to checks, you can find the [artifacts](https://github.com/misode/GM4_Beet/actions/runs/1786177809). Only the modified modules are zipped and included in the artifact.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -47,7 +47,7 @@ jobs:
 
       - name: Build all modules
         if: github.event_name != 'pull_request'
-        run: beet build --set 'meta.patch=true'
+        run: beet --set 'meta.patch=true' build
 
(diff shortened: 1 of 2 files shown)


RitikShah:
@SpecialBuilder32 It looks like this just produces zips for PRs so that folks can easily test them (option 1)!

misode:
@SpecialBuilder32 Yes the former. Beet is not super relevant here, it's just a QOL change to more easily share zipped modules of a PR. I fully agree that beet knowledge or running python should not be required for local development.

PR #752, opened 2022-02-18
Rope Ladders Module

MichaelMiner137:
Allows you to use a "rope ladder" on a a chain of ladders causing the ladder chain to be extended down by 1 block.
Rope ladders can only be used when you are aiming either at the top of a ladder chain or at the bottom of a ladder chain.

Currently there is no way to craft rope ladders as this module will require lib_machines


gm4_rope_ladders/data/gm4_rope_ladders/functions/ladder_placement/interact_rcd.mcfunction
```diff
@@ -0,0 +1,13 @@
...
+# run from advancement gm4_rope_ladders:rcd_ladder
+
+advancement revoke @s only gm4_rope_ladders:rcd_ladder
+
+scoreboard players set $ladder_placed gm4_rol_data 0
+
+function gm4_rope_ladders:ladder_placement/raycast/init
+
+execute if score $ladder_placed gm4_rol_data matches 1 run playsound minecraft:block.ladder.place ambient @a[distance=..15] ~ ~ ~
+execute if score $ladder_placed gm4_rol_data matches 0 run title @s actionbar {"text":"Rope ladder can only be 64 blocks long","color":"red"}
```
Bloo-dev, line 13: I don't think this actionbar text fits the vanilla style. Use a sound or particle to communicate this, instead of the actionbar. That will also save you from having to provide translations.

gm4_rope_ladders/data/gm4_rope_ladders/functions/ladder_detection/spawn_rcd.mcfunction
```diff
@@ -0,0 +1,7 @@
+# spawns "right click detection" wandering trader
+# @s = player holding rope ladder
+# position of ladder
+# run from gm4_rope_ladders:ladder_detection/found_ladder
+
+summon wandering_trader ~ ~100 ~ {NoGravity:1b,Silent:1b,Invulnerable:1b,NoAI:1b,Tags:["gm4_rol_rcd","gm4_rol_rcd_ladder"],ActiveEffects:[{Id:14b,Amplifier:0b,Duration:999999,ShowParticles:0b}],Offers:{}}
+execute positioned ~ ~100 ~ run tp @e[type=wandering_trader,tag=gm4_rol_rcd_ladder,dx=0] ~ ~-100 ~
```
Bloo-dev, line 7: Instead of relying on spatial selection of the `wandering_trader`, you could harden your algorithm by introducing a `new_something` tag, and remove the tag after initialization.

gm4_rope_ladders/data/gm4_rope_ladders/tags/blocks/valid_replace.json
```diff
@@ -0,0 +1,6 @@
+{
+  "values": [
+    "minecraft:air",
+    "minecraft:cave_air"
+  ]
```
Bloo-dev, line 5: You should use `#gm4:air` here.

gm4_rope_ladders/data/gm4_rope_ladders/tags/blocks/ladder_scan.json
```diff
@@ -0,0 +1,7 @@
+{
+  "values": [
+    "minecraft:air",
+    "minecraft:cave_air",
```
Bloo-dev, line 4: You should use `#gm4:air` here.

gm4_rope_ladders/data/gm4_rope_ladders/tags/blocks/air.json
```diff
@@ -0,0 +1,6 @@
+{
+  "values": [
+    "minecraft:air",
+    "minecraft:cave_air"
+  ]
+}
```
Bloo-dev, line 6: Why does this file exist? This seems like a leftover and would be redundant as `#gm4:air` handles this for you.

Bloo-dev (changes requested):
This is not a comprehensive review, but should act as a good starting point. I've found a couple of issues that should be resolved:
- You're including outdated versions of our logo
- The use of `$ray` for both the vertical ladder stepper as well as the right click detection ray makes things slightly confusing, I would suggest splitting that up into two fake players
- There's a performance issue with the way you're checking for players holding rope ladders

Bloo-dev (approved):

gm4_rope_ladders/pack.mcmeta
```diff
@@ -0,0 +1,38 @@
...
+        ],
+        "Updated By": [
+            "USERNAME",
+            "USERNAME"
+        ],
+        "Icon Design": [
+            "USERNAME"
+        ]
+    }
+}
```
Bloo-dev, line 38: Credit yourself :) (and fill out the template fields).

gm4_rope_ladders/data/gm4_rope_ladders/functions/init.mcfunction
```diff
@@ -0,0 +1,11 @@
+scoreboard objectives add gm4_rol_data dummy
+scoreboard objectives add gm4_rol_break_ladder minecraft.mined:minecraft.ladder
```
Bloo-dev, line 2: Using `rol` (an abbreviation) isn't necessarily bad, but just to make sure you're aware of it:
In Minecraft 1.18 scoreboard objectives don't have an explicit length limit anymore, so feel free to use `gm4_rope_ladders_data`.

Bloo-dev:
I just noticed that this consumed ladders from creative mode players -- should be an easy fix, though!

sparks0:
Tested and reviewed the code. Pretty nifty module! It might be worth allowing ladders to replace flowers, grasses and snow layers, since they're sometimes in the way on the ground and in the case of double-tall variants might stop the ladder being reachable again.

sparks0 (approved):

PR #755, opened 2022-02-23
Bat Grenades - Add Bat in a Bottle

Dennis-0:
### Bat in a Bottle
- Hit a bat while holding a glass bottle to capture it
- Throw it on the ground to release the bat

#### To do: (Will be looked at if this feature update is accepted)
- [ ] Decide what should happen when the bottle hits the ground: Release, instantly explode, or either with certain conditions?
- [ ] Check if all important data is transferred from the bat to the Bat in the Bottle - Currently only `CustomName` and `Tags`
- [ ] Add Guidebook page for capturing/releasing a bat - Discover when obtaining a glass bottle after unlocking the main info
- [ ] Minor fixes
  - [ ] When using the Bat in a Bottle on water (to create a water bottle): Release or explode bat
  - [ ] Prevent custom glass bottles to be used in capturing a bat: Add nbt check
  - [ ] For bottles in stack size >1: Release bat for each item or ignore
- [ ] Set CustomModelData
- [ ] Custom texture for the RP

### DIFF
--- a/gm4_bat_grenades/data/gm4_bat_grenades/advancements/capture.json
+++ b/gm4_bat_grenades/data/gm4_bat_grenades/advancements/capture.json
@@ -1,0 +1,32 @@
+{
+  "criteria": {
+    "requirement": {
+      "trigger": "minecraft:player_hurt_entity",
+      "conditions": {
+        "entity": {
+          "type": "minecraft:bat"
+        },
+        "player": {
+          "equipment": {
+            "mainhand": {
+              "items": [
+                "minecraft:glass_bottle"
+              ]
+            }
+          }
+        },
+        "damage": {
+          "type": {
+            "is_explosion": false,
+            "is_fire": false,
+            "is_magic": false,
+            "is_projectile": false
+          }
+        }
+      }
+    }
+  },
+  "rewards": {
+    "function": "gm4_bat_grenades:bottle/hit"
+  }
+}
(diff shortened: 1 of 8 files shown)


Bloo-dev:
I'm not to sure about this from a feature POV. This adds a lot of weight (relative to the current size of Bat Grenades), which drastically increases the required maintenance for a module that was previously really easy to maintain. However, the feature it adds doesn't seem like it would get a lot of use.

Dennis-0:
> I'm not to sure about this from a feature POV. This adds a lot of weight (relative to the current size of Bat Grenades), which drastically increases the required maintenance for a module that was previously really easy to maintain. However, the feature it adds doesn't seem like it would get a lot of use, or not enough to justify the increase in module size.

If this update is considered too much weight for this small module, I could reduce the function to just one, and make it explode on impact with the ground (to make it actual grenades).
I just think this could be a fun feature to make this module more than just a mild annoyance, although I also get it if we'd just want to keep Bat Grenades as it is. 

I mostly just wanted to get this requested feature into a PR, so it doesn't stay as an idea that's never accepted/denied.

sparks0:
Not keen on this mainly because why can you bottle bats and no other mobs? Fish, silverfish, endermites and chickens are all about the same size. But then the purpose of the module becomes more of an entity mover module like a "creature cage" and we'd have to support every mob and ask if it's balanced to let people just pick up and store mobs in their inventory.

Bloo-dev:
I'm closing this as the idea was not able to obtain the required admin approvals.

Issue #759, opened 2022-03-06
Gemini bred axolotls despawn when there aren't players around


hozz8528:
**Describe the bug**
title

**To Reproduce**
1. Breed axolotls with gemini
2. Walk away 128 blocks
3. Come back and the gemini 'twin' is gone

**Expected behavior**
Gemini 'twin' not gone

PR #760, opened 2022-03-07
Adds PersistenceRequired to all Gemini babies

EpyonProjects:
Should address #759, thanks @hozz8528 for catching this.


Bloo-dev:
There's no need to add this tag to all gemini babies, as most mobs do not require `PersistenceRequired`.
The only mobs that are breedable and need this tag should be ocelots, cats and axolotls. Trader llamas only require it if `DespawnDelay` is set, but I'm unsure what this defaults to as we do not explicitly set it.

Bloo-dev (changes requested):
Please address the issues raised in https://github.com/Gamemode4Dev/GM4_Datapacks/pull/760#issuecomment-1062020506.

Bloo-dev (approved):

PR #762, opened 2022-03-22
[Experiment] Use submodules for libraries

misode:


Bloo-dev:
This has been discussed in a staff meeting and is considered to be harder to maintain than having the libraries within the main repo.
The library folders will then hold their own license files and documentation internally.

PR #764, opened 2022-03-27
Adding Cavotio shamir

kyrkis:
The cavotio shamir is a metallurgy add on which adds the thorium band called Cavotio

Cavotio can be put on any pickaxe and grant's the ability to instamine close to anything. The blocks mined do not drop any items though, even including containers. 

The idea of this shamir is that it ties in to thorium's destructive property by being able to destroy blocks very easily with the side effect of also destroying the items. This makes this shamir very useful for excavating, for an underground base for example but close to useless for resource gathering, thus balancing it out.


Bloo-dev:
This was brought up for discussion in a staff-internal meeting. The instant mine in combination with voiding items is an interesting concept and a great match for a shamir, however, voiding items isn't enough to balance this.
A durability penalty is out of the question as that would make this shamir too similar to Audere.

The following options were suggested to balance this properly:
1. Breaking blocks with the shamir has a chance of spawning flowing lava blades (low lava level)
2. To use the shamir you have to have a certain potion effect only obtainable via beacons

Either option seems good, however, 2. will be harder to telegraph to the player. A combination of both could also work.
Note that 1. essentially forces the player to use fire resistance anyways, which could be an interesting twist.

Bloo-dev:
This will be reworked and submitted as another PR from another fork. Closing this PR.

PR #765, opened 2022-04-11
Add Dripleaf Filters

sparks0:
This module adds a new passive functionality to dripleaves. If an item normally lands on a dripleaf, the dripleaf collapses and the item falls through. With this module, the dripleaf will look at the items in a hopper placed in front of it, and if any of the items match the item on the leaf's ID, the item slides off the leaf into the hopper.

Screenshot_3

The purpose of this module is to allow the sorting of non-stackable items without the immense cost of allays, especially on multiplayer servers. The leaves can also don't need to be constantly linked to a note block, which might make for much simpler setups.

This module uses no marker entities inside the leaves which means you can use as many leaves as you want without impacting performance. It runs on the normal 16tick clock.


gm4_dripleaf_filters/data/gm4_dripleaf_filters/functions/check_hopper_filter.mcfunction
```diff
@@ -0,0 +1,35 @@
...
+
+# store the item's ID in storage
+data modify storage gm4_dripleaf_filters:cache itemCompare set from entity @s Item.id
+# store the hopper's items in storage
+data modify storage gm4_dripleaf_filters:cache hopperCompare set from block ^ ^-1 ^1 Items
+
+# loop through the slots to see if any IDs match
+# slot 0
+execute store success storage gm4_dripleaf_filters:cache itemsAreDifferent byte 1 run data modify storage gm4_dripleaf_filters:cache hopperCompare[0].id set from storage gm4_dripleaf_filters:cache itemCompare
+execute if data storage gm4_dripleaf_filters:cache {itemsAreDifferent:0b} run tag @s add gm4_dripleaf_filters_match
```
BluePsychoRanger, line 12: Many of these checks are using tags on the entity, but this can be moved completely to scores of fake players, which is more efficient. For an example, look at how the checks are done in custom crafters recipe_validity_check (lib_machines version). This function checks if the counts match, but the same can be done for the id. https://github.com/BluePsychoRanger/GM4_Datapacks/blob/lib-machines/lib_machines/data/gm4_custom_crafters-2.0/functions/recipe_validity_check.mcfunction

BluePsychoRanger (changes requested):
There are some optimizations that can be made in this module. Most notably is not using data storage as a temporary check, and instead using scores with fake players. Another optimization is not using the entity's tags to verify things and instead using (once again) scores with fake players. Scores with fake players are the most optimized way to do temporary checks, since it doesn't have to parse any files (data storage) or find entities (entity tags).

gm4_dripleaf_filters/data/gm4_dripleaf_filters/functions/check_hopper_filter.mcfunction
sparks0: I have now moved the two tags and the byte storage value to custom players with at $ prefix. Thanks for the feedback!

gm4_dripleaf_filters/data/gm4_dripleaf_filters/functions/check_hopper_filter.mcfunction
```diff
@@ -0,0 +1,38 @@
...
+# slot 4
+execute unless score $filters_match gm4_dripleaf_filters matches 1 if data storage gm4_dripleaf_filters:cache {hopperCompare:[]} run scoreboard players set $hopper_checked gm4_dripleaf_filters 1
+execute unless score $filters_match gm4_dripleaf_filters matches 1 unless score $hopper_checked gm4_dripleaf_filters matches 1 store success score $items_are_different gm4_dripleaf_filters run data modify storage gm4_dripleaf_filters:cache hopperCompare[0].id set from storage gm4_dripleaf_filters:cache itemCompare
+execute unless score $filters_match gm4_dripleaf_filters matches 1 unless score $hopper_checked gm4_dripleaf_filters matches 1 if score $items_are_different gm4_dripleaf_filters matches 0 run scoreboard players set $filters_match gm4_dripleaf_filters 1
+execute unless score $filters_match gm4_dripleaf_filters matches 1 unless score $hopper_checked gm4_dripleaf_filters matches 1 run data remove storage gm4_dripleaf_filters:cache hopperCompare[0]
+
+execute if score $filters_match gm4_dripleaf_filters matches 1 run function gm4_dripleaf_filters:deposit_item
```
Bloo-dev, line 38: This could be done recursively, saving on one command per slot and making this a lot shorter. For this you have to create one function `match_item`:
```
execute store success score $items_match gm4_dripleaf_filters run data modify storage gm4_dripleaf_filters:cache hopperCompare[0].id set from storage gm4_dripleaf_filters:cache itemCompare
execute if score $items_match gm4_dripleaf_filters matches 0 run data remove storage gm4_dripleaf_filters:cache hopperCompare[0]
execute if score $items_match gm4_dripleaf_filters matches 0 if data storage gm4_dripleaf_filters:cache hopperCompare[0] run function gm4_dripleaf_filters:match_item
```

That is called from `check_hopper_filter`:
```suggestion
# loop through the slots to see if any IDs match
function gm4_dripleaf_filters:match_item
execute if score $items_match gm4_dripleaf_filters matches 1 run function gm4_dripleaf_filters:deposit_item
```

Note that I have not tested this in-game.

gm4_dripleaf_filters/pack.mcmeta
```diff
@@ -0,0 +1,28 @@
...
+    "site_description": "Dripleaves will flick items that match a hopper's contents into the hopper for you!",
+    "wiki_link": "https://wiki.gm4.co/wiki/Dripleaf_Filters",
+    "required_modules": [],
+    "recommended_modules": ["book_binders"],
+    "credits": {
+        "Creator": [
+            "Sparks"
+        ],
+        "Icon Design": [
+            "No Icon Yet"
```
Bloo-dev, line 25: Leave this empty.

Bloo-dev (changes requested):
The general structure of this look good, I have left you some comments on making this code shorter.
You're using windows line endings, please make sure to commit unix line endings.
Make sure your header comments are in line with [CONTRIBUTING.md](https://github.com/Gamemode4Dev/GM4_Datapacks/blob/master/CONTRIBUTING.md).

sparks0:
I've just realized that potions all have the same ID. Should that item specifically check for nbt data too? I suppose you could potentially use liquid tanks to sort potions, but then you lose things like duration.

Bloo-dev:
> I've just realized that potions all have the same ID. Should that item specifically check for nbt data too? I suppose you could potentially use liquid tanks to sort potions, but then you lose things like duration.

There are a lot of unstackable items that store important data in NBT, all three potion types, axolotl and tropical fish buckets, shulker boxes and enchanted books just to name a few.
Personally, I'd either stick to the fuzzy id matching or do exact matching, instead of a hybrid of the two.

sparks0:
> > I've just realized that potions all have the same ID. Should that item specifically check for nbt data too? I suppose you could potentially use liquid tanks to sort potions, but then you lose things like duration.
> 
> There are a lot of unstackable items that store important data in NBT, all three potion types, axolotl and tropical fish buckets, shulker boxes, and enchanted books just to name a few. Personally, I'd either stick to the fuzzy id matching or do exact matching, instead of a hybrid of the two.

Probably a good idea. Keep it simple. I'll mull over some ideas for maybe a separate module that offers stricter sorting, maybe as a by-product of something else.

BluePsychoRanger (approved):
Looks good! Code is much more optimized now

Bloo-dev (approved):

PR #766, opened 2022-04-11
Added diamond toggle to disassemblers

sparks0:
As discussed in the 4/4/22 staff meeting, this update adds a (by default OFF) toggle to disassemblers that allows players to choose to let the custom block disassemble diamond tools alongside the normal selection.

Use `/scoreboard players set disassemble_diamonds gm4_disassembler 1` to enable diamond disassembly.


gm4_disassemblers/data/gm4_disassemblers/functions/check_disassemble_recipe.mcfunction
```diff
@@ -32,6 +32,16 @@ execute if data storage gm4_disassemblers:temp/item Item{id:"minecraft:leather_c
...
+execute if score disassemble_diamonds gm4_disassembler matches 1 if data storage gm4_disassemblers:temp/item Item{id:"minecraft:diamond_leggings"} run function gm4_disassemblers:items/diamond_leggings
+execute if score disassemble_diamonds gm4_disassembler matches 1 if data storage gm4_disassemblers:temp/item Item{id:"minecraft:diamond_boots"} run function gm4_disassemblers:items/diamond_boots
+execute if score disassemble_diamonds gm4_disassembler matches 1 if data storage gm4_disassemblers:temp/item Item{id:"minecraft:diamond_chestplate"} run function gm4_disassemblers:items/diamond_chestplate
+execute if score disassemble_diamonds gm4_disassembler matches 1 if data storage gm4_disassemblers:temp/item Item{id:"minecraft:diamond_pickaxe"} run function gm4_disassemblers:items/diamond_pickaxe
+execute if score disassemble_diamonds gm4_disassembler matches 1 if data storage gm4_disassemblers:temp/item Item{id:"minecraft:diamond_shovel"} run function gm4_disassemblers:items/diamond_shovel
+execute if score disassemble_diamonds gm4_disassembler matches 1 if data storage gm4_disassemblers:temp/item Item{id:"minecraft:diamond_hoe"} run function gm4_disassemblers:items/diamond_hoe
```
BluePsychoRanger, line 43: These all check the same score; perhaps consider a separate function. The duplicate score checks aren't super bad though and is likely about the same as calling a function.

gm4_disassemblers/data/gm4_disassemblers/functions/init.mcfunction
```diff
@@ -1,6 +1,7 @@
 scoreboard objectives add gm4_disassembler dummy
 scoreboard players set #100 gm4_disassembler 100
 scoreboard players set #2 gm4_disassembler 2
+scoreboard players add disassemble_diamonds gm4_disassembler 0
```
BluePsychoRanger, line 4: I'd prefer a `gm4_disassemblers_config` scoreboard objective so that it's clear these are values that _should_ be modified by users. It'd also be nice to have a consistent standard for config scoreboards for future data packs (currently the only other data packs with a config is orbis, which uses `gm4_orbis_config` and the guidebook, which uses `gm4_guide_config`)

BluePsychoRanger (changes requested):
Looks good functionally, might wanna consider a separate config scoreboard objective to indicate modifiable settings.

gm4_disassemblers/data/gm4_disassemblers/functions/init.mcfunction
sparks0: I actually completely agree but I seem to remember from the meeting that someone (@Bloo-dev ? ) was against this. I think he wanted the impact of this setting to be as minimal as possible. Perhaps he can remind us why he wanted it this way?

Bloo-dev: That is true, yes. I don't think an extra scoreboard is justified, even though the impact of one is minimal.

sparks0:
Fixed line endings

PR #769, opened 2022-05-10
Fix machine placement at extreme altitudes

BluePsychoRanger:
At the top and bottom of the world, the clone command to count the number of player heads didn't work, so the function read 0 heads to replace. This PR fixes that.

Minor load status has been bumped to 1, as libraries need a minor bump every update.


misode:
In the future it would be helpful to bump the version in a separate commit, for easier reviewing

misode (approved):
Seems to work fine in-game!

PR #770, opened 2022-05-12
Smithed compat

BluePsychoRanger:
Updates all data packs to be compatible with smithed conventions. This PR also adds an entity_version score to all custom entities to facilitate in future upgrading of pre-existing entities.

---
- [x] All custom entities have smithed.entity
- [x] Non-movable entities have smithed.strict
- [x] Custom blocks markers have smithed.block
--
- [x] @e selectors have tag=!smithed.entity
- [x] effect commands check tag=!smithed.strict
- [x] Commands that move entities check tag=!smithed.strict
- [x] Block checks check for a smithed.block entity before modifying the block
--
- [x] Vanilla loot tables use weld 
--

- [x] All custom entities have a entity version score


gm4_apple_trees/data/gm4_fruiting_trees/functions/main.mcfunction
```diff
@@ -1,4 +1,8 @@
 # age fruiting leaves
 execute as @e[type=armor_stand,tag=gm4_fruiting_leaf] run function gm4_fruiting_trees:tree/leaf/fruiting/age
 
+# entity versioning
+execute as @e[type=armor_stand,tag=gm4_fruiting_leaf,tag=!smithed.entity] run function gm4_fruiting_trees:upgrade_path/smithed_compat
+execute as @e[type=marker,tag=gm4_fruiting_sapling,tag=!smithed.entity] run function gm4_fruiting_trees:upgrade_path/smithed_compat
```
misode, line 6: I'm not convinced these upgrade paths are worth it. And if they're done, what's the reason for using `gm4_entity_version` if you're not testing for that here in `main`? If you can check whether or not a version needs to be upgraded without this version score, I don't think the score is necessary.

lib_trades/data/gm4_trades-1.1/functions/resolve_load.mcfunction
```diff
@@ -0,0 +1,4 @@
+execute if score gm4_trades load.status matches 1 if score gm4_trades_minor load.status matches 1 run function gm4_trades-1.1:load
+
+execute unless score gm4_trades load.status matches 1 run schedule clear gm4_trades-1.1:main
+execute if score gm4_trades load.status matches 1 unless score gm4_trades_minor load.status matches 1 run schedule clear gm4_trades-1.1:main
```
misode, line 4: Why is this commit part of this PR? Why does the version need to be bumped?

misode (changes requested):
Adding `smithed.entity` and `smithed.strict` seems fine. But I find the Weld implementation and the upgrade paths a bit questionable.

lib_forceload/data/minecraft/loot_tables/blocks/yellow_shulker_box.json
```diff
@@ -2,68 +2,109 @@
   "type": "minecraft:block",
   "pools": [
     {
-      "rolls": 1,
```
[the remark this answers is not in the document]
BluePsychoRanger: It's modified in a way that keeps functionality the same, so even if this loot table overwrites the old one, all data packs reliant on it will not show any effect.

gm4_apple_trees/data/gm4_fruiting_trees/functions/main.mcfunction
BluePsychoRanger: `gm4_entity_version` is for future use (if we need to change things, we currently need to check tags and then use a new tag which isn't sustainable). I could check for the score here, but that'd take an extra execute subcommand, which would be slightly less efficient, so I used a tag check wherever possible.

lib_trades/data/gm4_trades-1.1/functions/resolve_load.mcfunction
BluePsychoRanger: lib_trades was changed to not change wandering traders with the `smithed.entity` tag. Since a change was made, the version needs to be bumped.

BluePsychoRanger:
> @BluePsychoRanger Can you just take a quick browse through this and let me know if it is still valid for 1.19?

Yes, this PR has just been updated for 1.19

BluePsychoRanger:
This should not be merged until #785 is merged. This PR changes the `lib_trades` version, which will conflict with the versioning PR

PR #773, opened 2022-05-24
Hide Nametags of Custom Mobs

BluePsychoRanger:
Hides names of custom mobs by putting them on a team with `nametagVisibility never`. Default team is called `gm4_hide_name`, but other teams can be added to suit specific needs (i.e. additional options). This causes the death message to display the custom mob name while preventing the name tag from showing up when looking at the mob.

Also adds a `§` to the end of every custom mob name (some were missing) in order to allow for retexturing of mobs by Optifine/CIT.

Adds custom name to the Goblin added by Midnight Menaces.


Bloo-dev (approved):
Code reviewed this, could not find any typos.

PR #775, opened 2022-05-30
Add READMEs to some modules

BluePsychoRanger:
Adds READMEs, Licenses, and examples in each library module

TODO:
- [ ] teleport example for lib_forceload
- [x] lib_machines examples
- [ ] put READMEs and examples in base modules (fruiting trees, relocators, orbis, etc.)


misode:
Examples shouldn't be in the `data` folder, as that would mean they get included in the modules. Imo a better place would be `lib_forceload/examples/example_name/`.

BluePsychoRanger:
Okay, I was debating how to put the examples cuz in the separate repos we just had them as separate packs. So in `examples` would it be a full data pack with `pack.mcmeta` and the `data` folder?

misode:
> So in `examples` would it be a full data pack with `pack.mcmeta` and the `data` folder?

Yes that would be good. Unless we want to use [Lectern](https://github.com/mcbeet/lectern) for our examples, but I guess that would be weird with the rest of our modules not really using beet.

Bloo-dev:
We're closing this PR as it is very outdated and it seems better to start with a fresh PR based on the current master branch and set a smaller scope per PR.

PR #778, opened 2022-06-16
Fix tunnel bore drop for 1.19

BluePsychoRanger:
Item drop has changed in 1.19


misode:
Is this the only module that could be affected by this change?

BluePsychoRanger:
Good catch misode. Liquid minecarts are affected due to them being command block minecarts which don’t drop a command block

BluePsychoRanger:
I've tested this in game and all the mine carts drop their correct items

PR #781, opened 2022-06-25
Fix type_specific player predicate

BluePsychoRanger:
The auto-updater didn't catch these


misode:
Looks good, no testing required!

PR #782, opened 2022-06-29
Auto crafting

BluePsychoRanger:
The auto crafter is a multi-block structure utilizing a new block. It can perform any custom crafter recipe, automatically accepting any CC recipe expansions. Additionally, installing this module will add all vanilla crafting recipes to the GM4 Custom Crafter.

To craft the auto crafter base block the following recipe should be placed in a custom crafter:
Screen Shot 2022-06-28 at 9 02 44 PM

The multi block is as follows (the top and bottom blocks are barrels):
Screen Shot 2022-06-28 at 9 04 38 PM

The auto crafter interface is used to set the shape of the recipe (any item can be used to set the shape). Items are inputted through the barrel at the top. The auto crafter will take items out of the barrel and save them in a buffer (virtual inventory). Once the recipe is filled (based on the shape previously set), the auto crafter will stop taking out items. Once that happens it's ready to craft. To complete the craft, power the piston. The resulting item will appear in the output barrel.

If the recipe fails or if there are more items in the buffer than the recipe is set for (e.g. fill the shape, input items, then modify the shape), then the extra items get placed in the output barrel.

If the crafting operation doesn't overflow the output barrel, a sound and particle is used to convey that the crafter did something. There are different sounds and particles based on if the craft itself succeeded or failed.

If the crafting operation would overflow the output barrel, a sound and particle is used to convey that nothing happened. When this happens, all items stay in the auto crafter buffer (virtual inventory). This means after clearing the output, the piston can be powered again to try again.

--
TODO:
- [x] Generator for all vanilla recipes
- [x] Auto Crafter base block (using lib_machines)
- [x] Auto Crafter functionality
- [x] Drop items when crafting fails
- [x] Overflow detection
- [x] Mirroring


Bloo-dev:
We did some in-game testing and came to the following conclusions:
- recipes with empty rows or columns should be top-left aligned virtually when the shape is read into storage
- recipes that can be mirror should have their mirror version stored as a separate recipe (only a hand full of these exist, namely stairs, axes, leashes and the observer)
- (optional) the resource pack still needs textures for the GUI and the controller block

BluePsychoRanger:
> * recipes that can be mirror should have their mirror version stored as a separate recipe (only a hand full of these exist, namely stairs, axes, leashes and the observer)

I've updated this PR (and the generator) to include the mirrored recipes, along with force-aligning to the top-left (spyglass now is cartable as expected with the top-left rule). 

> * recipes with empty rows or columns should be top-left aligned virtually when the shape is read into storage

This has been decided to be moved to a separate Custom Crafters PR, since it's a change to how the CC processes recipes and other modules will need to be updated to support the new stricter recipe shape rules (e.g. ender hoppers needs to remove its duplicate recipe checks for each row). The auto crafter vanilla recipes already follow this stricter recipe shape rule, so once the CC PR is created and merged, players will be able to craft the vanilla recipes without top-left aligning.

PR #783, opened 2022-07-04
Crossbow Cartridges: Add Tadpoles

Dennis-0:
Add Tadpole in a Bucket, with the same functionality as the other fishes in bucket


Bloo-dev (approved):
Code looks good.

PR #785, opened 2022-08-18
Fix Versioning of Libraries

BluePsychoRanger:
We did our versioning wrong because we never verify if the minor version matches the major version before changing the minor version.

Without this PR, when we bump the major version, the minor version can remain the same as the old minor version. E.g. base 1.2 would convert to base 2.2 instead of base 2.0 since minor version 2 is greater than minor version 0.

Once this PR is merged, the versioning will work properly and the minor version will always match its major version. Note that this is a non-breaking change and as long as this gets merged it will fix the mistakes of the old versioning


SpecialBuilder32 (approved):
Glad we found and fixed this before it caused any trouble :P

PR #787, opened 2022-08-23
CC Shifting

BluePsychoRanger:
This PR allows custom crafting recipes to be inputted in any part of the grid. Before the recipe checks are performed, the items are virtually shifted to the top-left in storage. This is a breaking change and all modules have been updated to follow the new recipe format which assumes top-left alignment. 

This PR also updates a few recipes:
- ender hopper minecarts, liquid minecarts, and tunnel bores are now shapeless
- the horse armor recipe works when mirrored

Additionally, this PR adds an output caching system which stores the contents of the dropper after a recipe succeeds. Once this happens, the crafter will not attempt to check recipes until the contents of the dropper changes. This is to prevent recursive recipes from happening, especially with shapeless recipes.


BluePsychoRanger:
I've tested this and it works. Another tester is appreciated

Bloo-dev:
Tested this in-game, the PR does not yet introduce this behavior for auto-crafters.

Bloo-dev (commented):
I did a code review of the implementation of the recipe shifting, and it seems good (couldn't find anything to complain about). Please not that I did not code-review all recipe changes.

During the review I noticed that thus PR deletes the files `recipe_check.json` and `recipe_check.mcfunction` from teleportation anchors, is this on prupose?

BluePsychoRanger:
> During the review I noticed that thus PR deletes the files `recipe_check.json` and `recipe_check.mcfunction` from teleportation anchors, is this on purpose?

They were extra files from a refactor; figured I'd remove it in a single PR

PR #788, opened 2022-08-30
First Run of Smithed Releases

BluePsychoRanger:
This PR prepares **Chairs**, **Note Block Interface**, and **Ziprails** to be released on Smithed.

No modifications are made to the module code, only some READMEs and images were added. The READMEs are added for each module as a selling point to be read on the Smithed launcher, similar to how PMC posts are made (in fact, the info was ripped from our previous PMC posts).

Modules now have an `images` folder which the README references for images


SpecialBuilder32 (approved):
Looks good. We'll probably tweak positioning or aspect ratio of images once we've published a test-pack on smithed.

PR #796, opened 2022-09-30
Add Cavaro Shamir

Bloo-dev:
This implements the Cavaro Shamir as outlined by @kyrkis.

### Metallurgy Basics
- Thorium Band
- CMD 140
- For pickaxes & shovels

### What does it do?
Cavaro Tools void any blocks mined, however, Haste II is automatically granted to the player when using a Cavaro.
Additionally, breaking a block with Cavaro has a 1/16 chance to spawn a lava blade (**not** a lava source block).

Cavaro only grants haste after one block was broken, making accidental griefing less likely.


Bloo-dev:
This Shamir was discussed in an internal meeting, the main issue being the haste level of choice. This Shamir's use-cases depend on the haste level chosen:
- Haste 0-1 are obtainable via vanilla beacons
- Haste 0-6 doesn't instant-mine Deepslate
- Haste 4-127 breaks hand animations (no animation at all)
- Haste 128-255 instant-mines dirt and breaks block breaking particles and sounds

Block breaking particles and sounds may be fixed by commands, whereas the hand breaking animation can not be fixed by a command; Therefore this Shamir could go two ways:
1. Provides a haste level lower than 4 (0 to 3): The Shamir would essentially act as a portable beacon for carving out large areas, but is of little to no help on Deepslate levels, preventing diamond mining using this Shamir, but also preventing carving out large areas on Deepslate level.
2. Provides a haste level above 127 (128 to 255): The Shamir can be used to instant-mine Stones, Ores, Deepslate, Dirt, Sand, Gravel, etc. . This makes the Cavaro shovels redundant but allows using this Shamir to carve out large areas at deepslate level. Skilled players will be able to abuse this to mine diamonds by switching to another pickaxe once they encounter diamonds.

During the meeting, people argued for and against both implementations and no consensus could be found. The main arguments were:
- Cavaro is not useful if it simply acts as a portable beacon
- Cavaro is too overpowered if it allows for Deepslate instant-mining, as it can then be used to vastly speed up diamond mining
- Cavaro on shovels is useless, because shovels instant-mine most things anyways (with efficiency)
- Making Deepslate instant-mineable removes a lot of creativity from the game by making more creative Deepslate removal tactics redundant (e.g. Tunnel Bores, TNT Quarries)

Feedback is welcome. This Shamir will not be added to the repo until further discussion.

PR #786, opened 2022-08-22
Calling Bell: Adds the ability to summon a wandering trader

EpyonProjects:
- By right clicking on a bell with an emerald
- Consumes 1 emerald
- 2 sec delay between summons (lag failsafe)

PR #803, opened 2022-10-10
Update Mob Curing to require lib_trades-1.2+

EpyonProjects:
To be merged after #786, which updates lib_trades to 1.2

Mob Curing is broken with 1.1, and does not currently check minor version


BluePsychoRanger:
Can confirm. The reason it's broken isn't only the lack of version checking, but it also calls a function tag which calls a non-existent function in gm4_trades-1.1. I.e. it currently doesn't work even if the correct version of the library is loaded

Bloo-dev:
> Can confirm. The reason it's broken isn't only the lack of version checking, but it also calls a function tag which calls a non-existent function in gm4_trades-1.1. I.e. it currently doesn't work even if the correct version of the library is loaded

Does this PR fix this issue then?

EpyonProjects:
> > Can confirm. The reason it's broken isn't only the lack of version checking, but it also calls a function tag which calls a non-existent function in gm4_trades-1.1. I.e. it currently doesn't work even if the correct version of the library is loaded
> 
> Does this PR fix this issue then?

Should now that I've moved it from the Calling Bell submission

PR #804, opened 2022-10-11
Update Forterra to 1.19

EpyonProjects:
- Increased height range for post 1.18 height increase
- Fortune now modifies returns
- Works for Deepslate, and increased rates
- Decreased rates for Stone


Bloo-dev:
I am somewhat confused about the double mention of some ores in the stone loot table. We should discuss this in person.

Bloo-dev:
I was sent a spreadsheet that explains the layout of the loot table. This is still a lot of data to check, and typos may be likely. A playtest would be best to determine whether this is good to go.

Bloo-dev (approved):
The JSON is valid and agrees (judging from pointwise checks) with the vanilla terrain gen.

PR #813, opened 2022-10-25
Infinitas Shamir: Update to Spiraculum

EpyonProjects:
Infinitas, a play on "Infinity", is used to keep the fluid state of the bucket consistent.

Previously, Spiraculum was placed on an empty bucket and would "vacuum" nearby entities towards the player.  Infinitas keeps that functionality, and also introduces a "bottomless bucket" function that will drain liquids.

Applying Infinitas on a Water Bucket will allow the player to place water without losing it.  You can even pick up and place fish, like a regular Water Bucket.

A Milk Bucket can be repeatedly consumed, and you can place infinite Powder Snow.

Lava Buckets can be accessed by toggling `$lava_infinitas` in `gm4_ml_data`, allowing you to place an infinite amount of Lava buckets as well.  If the config is not toggled on, then creation of them is not possible either.


Bloo-dev:
What happens when you place a bucket with infinitas in a liquid tank or a lava bucket with infinitas in a furnace?

EpyonProjects:
> What happens when you place a bucket with infinitas in a liquid tank or a lava bucket with infinitas in a furnace?

It breaks.  We can't control tools used in machinery

EpyonProjects:
Just noticed I didn't mention it, you cannot currently use a stack of buckets with Infinitas.  I have no clue how to detect that.

EpyonProjects:
> Just noticed I didn't mention it, you cannot currently use a stack of empty Infinitas buckets. I have no clue how to detect that.

@Bloo-dev, following your confirmation from #799, I am still looking for a solution to unstack empty buckets with Infinitas.   Would it be reasonable to add an unstackable tag upon smooshing Infinitas?

Bloo-dev:
> > [...] I have no clue how to detect that.

Does using a bucket not trigger an item used scoreboard?

> [...] Would it be reasonable to add an unstackable tag upon smooshing Infinitas?

I suppose that would be a better option then having it break when stacked; We should invesitage ways to detect the use of stack buckets first.

Bloo-dev (changes requested):
- Smooshing of old Spiraculum bands does not work, and unsmooshing an old Spiraculum bucket returns Spiraculum instead of Infinitas.
- In my world, Infinitas on an empty bucket does not seem to work, instead Infinitas buckets act like normal buckets when picking up fluids, turning into water buckets and losing their NBT in the process (and thereby Shamir).

Placing water using an Infinitas Water Bucket (obtained via smooshing) works as expected.
It would be nice if there was a way to empty an Infinitas bucket; maybe shift right clicking could do that?

**This is not a code review, I simply tested the datapack in-game.**

EpyonProjects:
When unsmooshing the Spiraculum bucket, did you allow the item to change?  Placing and breaking doesn't update the head?

EpyonProjects:
Could you list the steps you took to install and create the empty Infinitas bucket?  I am not seeing your issue with empty Infinitas buckets, with new or upgraded worlds.

EpyonProjects:
> When unsmooshing the Spiraculum bucket, did you allow the item to change? Placing and breaking doesn't update the head?
> 
> Edit: The smooshing problem has been discussed and will be fixed in the following commit. I will look into shift right clicking to empty the bucket. Are you expecting to be able to pick up liquids with an empty Infinitas bucket using shift as well?

I have committed the fix for being able to smoosh Spiraculum, so it applies Infinitas onto the bucket.

If you still want the feature, I can add it so shift right clicking disables the Infinitas feature, allowing you to empty/pick up a liquid from/into the bucket

Bloo-dev:
I have updated this branch to include beet.

ToffeeMax:
This is now on the Public Server for testing

ToffeeMax (changes requested):
**Tested and confirm the following works:**

- Liquid Removal keeps bucket NBT
- Infinite Water, Powder Snow,Lava placement
- Infinite Milk Consumption
- Collection and replacement of tropical fish keeps NBT
- Putting of shamir on bucket works
- Infinite Lava Toggle works

**Needs to be fixed/missing functionality:**

- All buckets loose texture from resource pack. Minor change, but could be nice

**Needs investigating:**

- When installing the datapack, Infinitas didn't seem to run at all. A manual "init" "main" was needed.

ToffeeMax:
Bug:
- When bucketing up a tropical fish, it does not save the type of tropical fish and resets its type
- This does not mirror normal vanilla behavior (if that is what is being emulated)

EpyonProjects:
> **Needs investigating:**
> 
>     * When installing the datapack, Infinitas didn't seem to run at all. A manual "init" "main" was needed.

What did you see when you installed the pack?

PR #815, opened 2022-10-26
Gemini Shamir Update: Adds Allays and Frogs, fixes Turtles

EpyonProjects:
Both Frogs will lay Frogspawn block

Turtles did not work at all and spawned a baby Turtle instead of laying Turtle Eggs.  Now both Turtles will lay Turtle Eggs.


Bloo-dev (approved):
This looks good to me. Works in game too.

PR #819, opened 2022-11-08
Forming Press 1.19: Add Mangrove Log to Charcoal and Fix Dark Oak multiplier

EpyonProjects:
Added Mangrove Log to Charcoal Block recipe

Also, fixed the last log to charcoal recipe not storing success,  which would skip applying multiplier.  This fixes Dark Oak not applying multiplier and always outputting 2 Charcoal Blocks


Bloo-dev:
As a note for future updates: Ever since `gm4_auto_crafters` were introduced, custom crafting supports item tags, therefore all of these could be combined into a single item tag check.

However, this is not necessarily a blocking issue for this PR; it's better to have mangrove wood implemented for now.

PR #822, opened 2022-11-12
Auto Crafter QoL Changes

BluePsychoRanger:
- Planks can now be used instead of stairs for the auto crafter multiblock structure
- Auto Crafters run on a clock equal to dropper speed


SpecialBuilder32 (approved):
I was told in voice-chat to approve this PR as it was just tested exhaustively during its creation :P

PR #823, opened 2022-11-14
Worldgen Apple Trees and lib_trees

BluePsychoRanger:
## Refactor
Apple Trees and Fruiting Trees have been separated into separate modules. Fruiting Trees have been moved to `lib_trees` and no longer provide support for specific fruit aging. Apple Trees now uses worldgen and jigsaws to generate trees, so orbis and the seed generation/analyzer are no longer required.

## Changes
There are more apple tree variants. The frequency of apple trees have been raised and the average number of apples in an apple tree has been increased. Old apple tree saplings will not stack with new apple tree saplings, but both will be able to grow apple trees.

## Additions
Golden apple trees will generate somewhat rarely in the world and will produce golden apples (1 every 4 hours per leaf). Leaves will drop normal apple tree saplings. If the original gold ore below the tree remains, golden roots will grow, replacing stone and dirt. Golden roots will produce 1-3 gold ore per hour depending on how much it grows before being mined (because it branches at certain depths of the root growth).


gm4_apple_trees/pack.mcmeta
```diff
@@ -15,14 +15,12 @@
     "libraries": [
         "lib_trades"
     ],
```
Bloo-dev, line 17: This is missing `lib_trees`.

gm4_apple_trees/data/gm4_apple_trees/functions/leaf/apple/hand_interact.mcfunction
```diff
@@ -1,9 +1,8 @@
...
 # @s = player taking apple from tree with empty hand
-# at @s
+# located at @s
 # run from advancement apple_hand_empty
 
 advancement revoke @s only gm4_apple_trees:apple_hand_empty
 
 playsound minecraft:entity.item_frame.remove_item block @a[distance=..8] ~ ~ ~ 0.3 0.6
-
-execute positioned ^ ^ ^2.5 as @e[type=minecraft:armor_stand,tag=gm4_ripe_apple,distance=..3,sort=nearest,limit=1,predicate=!gm4_apple_trees:apple_holding] at @s align xyz run function gm4_apple_trees:tree/leaf/fruiting/set_stage/0
```
Dennis-0, line 9: Keep this line. It advances to the next stage, with the remaining time added.

gm4_apple_trees/data/gm4_apple_trees/functions/leaf/apple/pick.mcfunction
```diff
@@ -1,9 +1,10 @@
...
 # at player picking the apple
-# run from apple/interact
+# run from gm4_apple_trees:leaf/apple/interact
 
 # return item to player
 item replace entity @a[tag=gm4_apple_take,limit=1] weapon.mainhand from entity @s weapon.mainhand gm4_apple_trees:count
 
 # update armour stand
-execute at @s align xyz run function gm4_apple_trees:tree/leaf/fruiting/set_stage/0
+execute at @s align xyz positioned ~0.5 ~1.5 ~0.5 positioned ^-.25 ^ ^.1 run function gm4_apple_trees:leaf/apple/drop
```
Dennis-0, line 10: Revert this line. It advances to the next stage, with the remaining time added.
`execute at @s align xyz run function gm4_apple_trees:leaf/set_stage/0`

gm4_apple_trees/data/gm4_apple_trees/item_modifiers/count.json
```diff
@@ -8,5 +8,5 @@
     },
     "score": "gm4_apple_data"
   },
-  "add": true
+  "add": false
```
Dennis-0, line 11: This should be `true`. It takes the item count in the player's selected item (should be 1 less than before the interaction), then this count is added to the count on the armor stand's main hand (1), which is then moved to the player's selected slot.

Bloo-dev:
Further testing has shown that the method used to place apple trees (and apples) in this is deterministic based on the chunk coordinates. I.e. a chunk will always generate the same apple tree. This should be fixed before this can be merged.

Bloo-dev:
Possible solution:
Instead of placing a jigsaw to decide whether to place an apple / a leaf or not, we place a command block which executes `execute if predicate <random chance> run function place_apple/leaf`.
This will fix the leav and apple distribution, however, tree-shape selection could simply use a similar process.

misode:
You can keep the generator script as a reference for when you or someone else wants to rewrite it in the beet ecosystem

SpecialBuilder32 (approved):
I've tested this with the new versioning and all the features work as expected. 

Would like one more person to test the PR artifacts before we merge this :+1:

PR #826, opened 2022-11-29
Fix a few recipes in AutoCrafters, remove item deletion glitch

jpeterik12:
AutoCrafters Currently is broken for some recipes, such as TNT, and can delete items in the recipe shape interface if scoreboards interact strangely. 

The TNT issue was caused by an edge case when a recipe uses an inline item list. This was fixed by editing the generator.
The item deletion glitch was caused by some leftover code copied from Custom Crafters, used for when doing multiple crafts at one time. When installing the module on an older world, this could use a scoreboard value of 0 when multiplying the item counts, deleting all the items in the interface. The module shouldn't modify interface items at all, so I removed the unneeded code.
`auto_crafter/craft/apply_multiplier` and `finish_crafting`

This PR those two issues.

Some edge cases that do not exist yet, but may in the future:
Any vanilla recipe that uses buckets/bottles in either an inline list or a tag won't have the item placed back into it
Any shapeless recipe that uses the same inline list multiple times will not be combined (might allow crafting with only 1 of said item)


Bloo-dev:
Please provide more technical information about how this fixes the aformetioned issue.

Bloo-dev:
I did not test all recipes, but the code look good and the rework (and removal of the multi-craft) of the crafting method makes sense. Works in game.

PR #828, opened 2022-12-09
Beet 2: Electric Boogaloo

misode:
## Changes
- All `pack.mcmeta` files in the base, modules and libraries have been replaced with `beet.yaml`.
- Libraries can now be depended on transitively, and we can depend on libraries outside this repository, like smithed.
- Deleted the template pack, because it was outdated and caused constant problems
- The project now uses `poetry` to handle python dependencies. Data pack developers are now required to use this to build modules. More information can be found in [`CONTRIBUTING.md`](https://github.com/misode/GM4_Datapacks/blob/beet-2/CONTRIBUTING.md#contributing-to-gamemode-4).
- The `gm4_modules` patch version will now only be bumped when a change is made to the actual module, not to any libraries or base. This is only used for the automatic update messages so probably not a huge deal.
- Switched some module generation code to beet (all scripts except guidebook and auto crafter)
- Logic to publish new versions to Modrinth. This requires an op-in for each module and only handles creating new versions. Projects still have to be created manually for each module.

## Todo
- [x] Improve and update `CONTRIBUTING.md`
- [x] Document the python code with docstrings
- [x] Move all generating scripts to the beet pipeline (ignoring the guidebook)

## Future work
- Restructure the repository, for example with `modules` and `libs` folders.
- Start using [smithed](https://wiki.smithed.dev/) libraries where appropriate


CONTRIBUTING.md
```diff
@@ -90,5 +112,5 @@ Make sure to credit all relevant people in `pack.mcmeta`. Most modules credit th
```
RitikShah, line 112: If `pack.mcmeta` doesn't exist anymore, how do credits with for each pack? Maybe some parts of the CONTRIBUTING.md are outdated now?

convert.py
```diff
@@ -0,0 +1,40 @@
...
+    required:{' []' if len(required) == 0 else ''.join([f'{NL}      - gm4_{r}' for r in required])}
+    recommended:{' []' if len(recommended) == 0 else ''.join([f'{NL}      - gm4_{r}' for r in recommended])}
+    notes:{' []' if note is None else f'{NL}      - {note}'}
+    video: {'null' if len(video) == 0 else video}
+    wiki: {'null' if len(wiki) == 0 else wiki}
+    credits:{''.join([f'{NL}      {c}:{"".join([f"{NL}        - {p}" for p in credits[c]])}' for c in credits])}
+"""
+
+	with open(f"{id.removesuffix('pack.mcmeta')}/beet.yaml", "w") as f:
+		f.write(config)
```
RitikShah, line 40: Was this a script to turn the old `pack.mcmeta`s into `beet.yaml`s? Not sure if you want to keep this in the root level of the repo.

gm4_mysterious_midnights/beet.yaml
```diff
@@ -0,0 +1,29 @@
...
+      load: '../lib_forceload'
+
+meta:
+  gm4:
+    description: Full moons gain a element of surprise with Myserious Midnights! This module adds random events during full moons. You might encounter falling stars, or Skeletons wielding poisonous arrows! So you better listen for that howl...
+    required: []
+    recommended:
+      - gm4_midnight_menaces
+    notes: []
+    video: null
```
RitikShah, line 20: I had no idea that `null` existed in yaml. Why not just omit the field?

RitikShah (approved):
I left a couple of minor comments, but overall, looks like some major improvements. I think you should add some docstrings to your custom plugins explaining in more detail, perhaps on what data it read from `meta.gm4` or something.

You could also consider a `beet gm4 init` or something to replace the template pack in the future.

With downloading python, there can be a lot of pitfalls especially with how varied installation is across computers. Some people might already have python installed, some might have the Windows Store edition. I'd maybe clarify that the project is using Python 3.10. Fortuntely, `poetry` will at least manage the dependencies of the project separately.

CONTRIBUTING.md
misode: You're right I still need to update these sections

convert.py
misode: Deleted in the next commit 👍

gm4/plugins/create_manifest.py
```diff
@@ -0,0 +1,74 @@
...
+		"last_commit": head,
+		"modules": [m for m in modules if m.get("id") is not None],
+		"contributors": contributors,
+	}
+	ctx.cache["gm4_manifest"].json = new_manifest
+
+	os.makedirs(f'release/{version}', exist_ok=True)
+	with open(f'release/{version}/meta.json', 'w') as f:
+		json.dump(new_manifest, f, indent=2)
+		f.write('\n')
```
RitikShah, line 74: I'd highly recommend using `pathlib`'s `Path` object for this portion. The base `Context` object already uses it and it makes dealing with pathing and directories much nicer.

`ctx.output_directory` would point to the `releases` folder (if this is a release) and you can do something like this:
```py
versioned_dir = ctx.output_directory / version
versioned_dir.mkdir(parents=True, exist_ok=True)
meta_json = versioned_dir / "meta.json"
meta_json.write_text(
  json.dumps(new_manifest, indent=2)
)
```

gm4/plugins/create_manifest.py
```diff
@@ -0,0 +1,74 @@
...
+	modules: list[dict[str, Any]] = [{"id": p.name} for p in sorted(ctx.directory.glob("gm4_*"))]
+
+	for module in modules:
+		id = module["id"]
+		try:
+			with open(f"{id}/beet.yaml", "r+") as f:
+				meta = yaml.safe_load(f)
+				module["name"] = meta.get("name", id)
+				module.update(meta.get("meta", {}).get("gm4", {}))
+		except:
```
RitikShah, line 26: I think bare `except`s are really error-prone, I'd recommend looking for the specific exceptions here.

gm4/plugins/output_release.py
```diff
@@ -0,0 +1,17 @@
...
+	ctx.data.save(
+		path=f"{base_path}_{version}.zip",
+		overwrite=True,
+		zipped=True,
+	)
+
+	os.makedirs(base_path, exist_ok=True)
+	for file in ["README.md", "CREDITS.md", "pack.png"]:
+		if file in ctx.data.extra:
+			ctx.data.extra[file].dump(base_path, file)
```
RitikShah, line 17: I would just have two plugins in the same `output` file:
```py
# output.py
def beet_default(ctx: Context):
  ...
  
def release(ctx: Context):
  ...

```

RitikShah: Tbf, this is preference, but I like having less files for these smaller plugins since they group nicely together.
The same could work for the `create_*` and the `populate_*`, etc one as well.

In `beet`, you reference them the same `gm4.plugins.output.dev`!

CONTRIBUTING.md
RitikShah: I'd leave a description here about what exactly goes in `meta.gm4`. It be the same as before but you can describe what you need and what you don't need and perhaps make it clear on what each field *expects*.

gm4_mysterious_midnights/beet.yaml
misode: For the same reason I want to keep fields like `notes` and `recommended` as empty lists. It gives the file a consistent structure and can give hints about what fields might be missing before publishing a module. ~~Also I just came up with this explanation as I'm typing this~~

gm4/plugins/create_manifest.py
misode: Because I don't use `output: ` in config, I don't think `ctx.output_directory` will be ser correctly. But I'll use the other parts 👍

misode:
Thanks for all the feedback @RitikShah. About the flat structure of our repo: This is a remnant of when we used to place this git repo inside our test world `datapacks` folder.  Without any symlinks or build tools this was the only way. We could change it now, but that would break all old links and conflict with all other PRs so we'd have to time it right.

RitikShah:
Looks good 🎉. Awesome work on this PR, def. the work you put in here extends further into the beet community as a great example.

In the future, when we have an updated beet vscode plugin, I'd even recommend using some more structure for the `meta.gm4` structure so that there's better auto-complete in the `.yaml` file, but we can cross that road when it exists haha.

CONTRIBUTING.md
misode: Made these changes in https://github.com/Gamemode4Dev/GM4_Datapacks/pull/828/commits/d5150c1dec57675cea8fc5952af91dbb7e81166a, let me know if things are unclear or missing.

misode:
I've just added the necessary code to publish modules to modrinth (for when they start supporting datapacks). The way this will work is when a commit is made that changes the version number of a module, it will upload a new version to modrinth.
image

Bloo-dev:
Currently checking this out, expect a review by the end of the year!

SpecialBuilder32 (changes requested):
Alright so I've spent some time with this PR, getting familiar with how it would change how we develop modules and the new pipeline system it entails.

**Overall Comments and Thoughts**
This PR would completely change how we approach developing modules, as now instead of beet just being used for the packager it becomes the core of the development process. I think to most people, the most notable change in their workflow is going to be that the `main` branch no longer can be directly loaded into the datapacks folder (`pack.mcmeta` shows up on build only). 
For reviewers and more tech-savy users, this isn't really a bad change, as they should just get their modules from the release branch, or a PRs artifacts. However for developers this might take some getting used to for sure. 

In terms of ease of use and install stuff, I actually found using poetry and the special workflow commands Misode wrote to be incredibly straightforward. The biggest hiccups I ran into were with interactions with syntax highlighting extensions and what I assume are minor bugs. DHP on my computer for some reason takes ages to scan through all 84 of our modules and build its cache, so if my vscode is open to the root folder of the repo I basically can't auto complete and the highlighting lags by several lines. I fixed this by just opening only the folder for the module I was working on though, and then DHP had no issues keeping up. 

**Comments / Requested Changes**
Here are some issues I ran into that I believe are bugs, or behavior I wished was different.
1. The auto-reload `--reload` flag seems to be broken when using `beet dev` to build only a single module. From what I could tell, the required `livereload` datapack never makes it into the `datapacks` folder of the world. It does when building all the modules, but that has other problems
2. I found myself frequently clearing out the contents of the `out` directory when I wanted to inspect what exactly the build process spat out when including things. Whenever I messed up syntax and accidently build all modules, or when I switched to a new module to build, I wouldve kinda liked if the `out` directory was cleaned up. How feasible is it to add a `beet clear` (or similar) custom command to assist in clearing out remnant builds?
3. I couldn't get a non-watching version of the `beet dev` command to work; it always started a watch session. The script powering it seems to allow for a one-time build task, but I couldn't get any flags to trigger that behavior. 
4. When building all 84 of our current modules, it takes almost 2 minutes to build and then transfer them to the linked minecraft world. Is there any way to speed this up or is this an artifact of us having way too many datapacks that we're handling unzipped.  (I do think having them unzipped in the `out` directory is useful however)

**Conclusions and Future Work**
Overall, I really like this, and am excited to start using some of the more powerful tools to manage this behemoth of a project. Here's some areas of future work I want to bring up so we can start thinking about them, and to help convince the hesitant. 
1. I think this opens the possibility of merging the datapack and resource pack repos, and would let us manage CMD and translations alot more easily, but this would be a big project.
2. This pipeline format makes it crazy easy to start using tools like bolt, which I think would open alot of convivence doors in the upcoming future. Repetitive files would be much simpler to read/write with bolt available.

RitikShah:
> When building all 84 of our current modules, it takes almost 2 minutes to build and then transfer them to the linked minecraft world. Is there any way to speed this up or is this an artifact of us having way too many datapacks that we're handling unzipped. (I do think having them unzipped in the out directory is useful however)
`beet` has aggressive caching by default so this is likely a one time build constraint.

> I couldn't get a non-watching version of the beet dev command to work; it always started a watch session. The script powering it seems to allow for a one-time build task, but I couldn't get any flags to trigger that behavior.
Currently, the `beet dev` forwards / sets up the `beet watch` command. Imo, there should be a `beet dev` which emulates `beet build` and a `--watch` flag or `watch` subcommand to emulate the `beet` command better.

> I found myself frequently clearing out the contents of the out directory when I wanted to inspect what exactly the build ...
Another note here is that linking your packs to a world will leave leftovers as well, if you switch to another command. This might be something to note / upstream to the official `beet` tool, but it shouldn't be too hard to implement here as a trial.
It also might be worth to have an option to have no output aka only link.

I'm really happy to see some perspectives from outside the beet community on integrating it with such a long-standing and active project. Props for Misode on all the work they have put into this 👏🏽!

.github/workflows/main.yml
```diff
@@ -42,37 +44,49 @@ jobs:
...
-        run: beet build
+        run: poetry run beet -p beet-release.yaml build
+        env:
+          BEET_MODRINTH_TOKEN: ${{ secrets.MODRINTH_TOKEN }}
 
-      - name: Build modules from pull request
+      - name: Build all modules for pull request
         if: github.event_name == 'pull_request'
-        run: beet --set 'meta.pull_request_base=origin/${{ github.base_ref }}' build
+        run: poetry run beet -p beet-release.yaml build
```
RitikShah, line 76: What was the reasoning being this change, is `meta.pull_request_base` not being used?

misode:
> DHP on my computer for some reason takes ages to scan through all 84 of our modules and build its cache.

Yeah this has been a problem even before his PR, probably because of the absurd size of our project. Hopefully this will be improved with Spyglass v4 :tm:

> I wouldve kinda liked if the out directory was cleaned up. How feasible is it to add a beet clear (or similar) custom command to assist in clearing out remnant builds?

I have added `beet clean` which clears the `out` folder, and also added `--clean` to `beet dev` (this only clears the folder once at the start of running the command)

> I couldn't get a non-watching version of the beet dev command to work; it always started a watch session.

I have added `beet dev --watch` to invoke the old behavior. The default behavior now is to only build once.

SpecialBuilder32 (approved):
All the features look to be working properly. :+1: 

In today's staff meeting we agreed to merge this and start figuring out what breaks over the coming week.

PR #833, opened 2022-12-14
Fix weld implementation of yellow_shulker_box

BluePsychoRanger:
The loot table works in vanilla, but when used with weld it targets the wrong path.


RitikShah:
I would hold on merging this, there might be smaller tweaks with `weld` (atm, `priority.default` is getting removed, though there's no harm in having it).

Bloo-dev:
Has the `weld` implementation stabilized by now? If so, this could be marked as a proper PR.

BluePsychoRanger:
> Has the `weld` implementation stabilized by now? If so, this could be marked as a proper PR.

Progress is slow, I think rx is still working on it

PR #841, opened 2023-01-09
Fix Relocators custom block handling

BluePsychoRanger:
There is some weird inconsistency with `execute store success score ... run function` that causes some blocks to fail in placement. This probably fixes it by removing the use of `store success score` and just setting the score within the function that handles the placement of custom blocks


BluePsychoRanger:
I've resolved the merge conflicts. This should be ready to merge

SpecialBuilder32 (approved):

PR #844, opened 2023-01-23
Intro Song on Install Only

EpyonProjects:
- ~~adds fake player $enabled in gm4_intro_song, used to disable song~~
- ~~removes the immediate creative flag to play song~~
- ~~increases delay of song start after advancement unlock~~

## Post Discussion Summary
After deliberation, this PR makes the following changes:
- The intro song now only plays to all players online when GM4 is first installed.
- The intro song is always accompanied with an advancement toast welcoming the player to gm4
- Previous intro song behavior (where the song plays to every player upon first time joining the world) can be restored by setting the config flag `scoreboard players set $first_join gm4_intro_song 1`


Bloo-dev (changes requested):
I will continue to veto making the song into an opt-in or removing it completely unless you download a separat Datapack.

Gamemode 4 needs some brand identity, and the intro song is part of that. Noone is going to download the "Gamemode 4 Song Datapack", it is practically an advertisement, and noone intentionally wants those, but it is a necessary evil.

This is in addition to the fact that this song plays once. It is not as big of an issue as a loud minority makes it out to be. An easier opt-out option for server owners (via a score) could still be added, but an opt-in is not the way.

misode:
People already hate data packs that print messages to chat, our data packs play a weird jingle out of nowhere without players asking for it. Our intro song has only existed for a few major versions, we did fine before that.

> Gamemode 4 needs some brand identity, and the intro song is part of that.

The only way we could get new users from this song is because they found it annoying, looked for where it came from, and then asked us how to get rid of it.

If you install a module for singleplayer you already know GM4, so we don't need to advertise ourself. If you run a server with GM4 modules you likely don't want to bother every player with a weird unknown jingle and you will try to find a way to disable it or risk confusing your player base. The only scenario where this makes sense is a custom GM4 focused server.

gm4_intro_song/data/gm4_intro_song/functions/init.mcfunction
```diff
@@ -0,0 +1,12 @@
+scoreboard objectives add gm4_intro_song dummy
+scoreboard objectives add gm4_intro_song_t dummy
+
+scoreboard objectives add gm4_intro_walked minecraft.custom:walk_one_cm
+
+# enable intro song by default, upon first installation
+execute unless score $enabled gm4_intro_song = $enabled gm4_intro_song run scoreboard players set $enabled gm4_intro_song 1
```
misode, line 7: If this is a separate module an enabled config may not be necessary? Except maybe for when you want to temporarily disable/enable it on a server?

EpyonProjects: I just left in the option to disable it while still having it installed.   I can take this out if you think it's unnecessary since it's standalone

EpyonProjects:
Bloo and Misode, is there a specific request from either of you?

Bloo has requested changes but only stated he's going to decline this PR.

Misode has brought up a possible change but it is unclear if you want me to implement that.  Either keeping the intro song in base with the config toggle, or removing the toggle and keeping the pack separate is fine.  Just want to know how to proceed from here.

misode (approved):

misode:
I'd like for this PR to go forwards. @Bloo-dev have you given our arguments another thought?

misode:
Do you know of any other mod or data pack that plays its marketing song when you first join the world? No. Imagine downloading a random zoom mod and a weird tune plays in-game when you first start up the game.

The song would not go unused, it's still in our YouTube videos. That's where it belongs. I don't hate the song, I just think this is not the correct place for it. Even in non-server environments it can be seen as obtrusive.

Bloo-dev:
> I'd like for this PR to go forwards. @Bloo-dev have you given our arguments another thought?

I have thought about this a bit more and came to the conclusion that, even if it only plays once, it plays without the player doing anything to make it play, which may indeed feel off -- like when a video suddenly auto-plays full volume on a website.

I still think this song has a place in our datapacks, though, just not as an autoplay without interaction. And instead of just removing it, there should be a replacement. I propose the following:
- Introduce a "This world is running Gamemode 4" toast using advancements, that displays under the condition that the song currently plays. It's less intrusive but still gets our name out there, and let's people know that Gamemode 4 is installed.
- Change the play condition of the song: It should only play for players who are online when Gamemode 4 is initially installed. That way the song never plays "uncommanded", making it less of a "jump scare". Together with the existing "Initializing Gamemode 4..." chat messages, and the toast, this would probably make for a better experience overall.

base/data/gm4_intro_song/functions/load.mcfunction
SpecialBuilder32, line 1: Load.mcfunction are no longer required by modules due to auto-versioning, but this feature should not get one.

SpecialBuilder32 (changes requested):
I think I like the direction this PR moved toward, but doesn't run for me. It seems to have some leftover changes from being its own separate module, and some from being tested in the same world multiple times.

SpecialBuilder32 (approved):
So I tried to fix the merge conflicts and somehow broke the universe so I'll call in the experts.  (feel free to rewrite history to obliterate my faulty 4 commits k thanks)

@EpyonProjects can you update this PR's primary description to describe the final set of changes and the how to config the new introsong?

Bloo-dev:
This is ready to be merged.
In the future we should make it so the toast plays upon first join (previous song condition), not only upon the song playing. I will create an issue for this so we don't forget.

PR #845, opened 2023-01-24
Forming Press: Add Packed Mud recipe

EpyonProjects:
5 Wheat and 4 Mud to create 8 Packed Mud, doubling the output of the regular recipe


Bloo-dev:
Seeing that forming press usually only does _forming_ recipes (aka compressing and heating), isn't this already covered by the mud bricks recipe?

I guess making the packed mud in the first place is also somewhat heating & _forming_, it's called _packed_ mud after all.
Open to hear what other people think!

SpecialBuilder32 (approved):
Tested and works as expected :+1:

ToffeeMax:
Packed Mud Gif

Added recipe rotating on wiki

PR #852, opened 2023-02-28
Shapeless Portals

TheThanathor:
Allows the creation of portals in any shape and with the use of crying obsidian


RitikShah (commented):
I scanned through the pack, and it looks good 👍🏽. Not placing approval since I didn't do anything that thorough so I didn't want to give a vote of false-confidence.

Potentially, could include a `distance=..` selector arg for the marker `@e`s, but tbh, I'm not even sure if that helps even more (esp with `limit=x` being optimized now).

Also, I'm guessing this is only for player-initiated portals.

gm4_shapeless_portals/data/gm4/advancements/root.json
```diff
@@ -0,0 +1,34 @@
+{
```
BluePsychoRanger, line 1: this advancement isn't necessary since this module doesn't add any visual advancements

gm4_shapeless_portals/data/gm4_shapeless_portals/advancements/light_portal.json
```diff
@@ -0,0 +1,45 @@
...
+          "block": {
+            "blocks": [
+              "minecraft:obsidian",
+              "minecraft:crying_obsidian"
+            ]
+          }
+        }
+      }
+    },
+    "module_loaded": {
```
BluePsychoRanger, line 22: This is not how to properly check that the module is loaded. First of all you're using `gm4_modules` instead of `load.status` which is the patch version thing, not the actual successful load score (this will never be equal to 1 for a module downloaded from the website). Second of all, putting it in its own criterion means that if this is EVER true, then whatever players are online will have this criterion fulfilled and if it ever becomes false it doesn't matter cuz the player already fulfilled this criterion. You should check this within the `player` conditions of the `item_used_on_block` criterion. See Undead Players for an example of proper module load checking.

gm4_shapeless_portals/data/gm4_shapeless_portals/functions/x/build.mcfunction
```diff
@@ -0,0 +1,7 @@
+# fill in the portal frame
+# @s = portal frame marker
+# at @s
+# run from x/flood_fill
```
BluePsychoRanger, line 4: This is run from `portal_found`

gm4_shapeless_portals/data/gm4_shapeless_portals/functions/z/build.mcfunction
```diff
@@ -0,0 +1,7 @@
+# fill in the portal frame
+# @s = portal frame marker
+# at @s
+# run from z/flood_fill
```
BluePsychoRanger, line 4: This is run from `portal_found`

gm4_shapeless_portals/data/gm4_shapeless_portals/functions/x/build.mcfunction
```diff
@@ -0,0 +1,7 @@
+# fill in the portal frame
+# @s = portal frame marker
+# at @s
```
BluePsychoRanger, line 3: `at @s` isn't very useful here. Perhaps change it to something like "at each closed space air block"

gm4_shapeless_portals/data/gm4_shapeless_portals/functions/x/check_location.mcfunction
```diff
@@ -0,0 +1,7 @@
+# check if new location is valid for a portal frame
+# @s = new portal frame marker
+# at @s
```
BluePsychoRanger, line 3: `at @s` isn't very useful here. Perhaps change it to something like "at a potential new portal block"

gm4_shapeless_portals/data/gm4_shapeless_portals/functions/x/flood_fill.mcfunction
```diff
@@ -0,0 +1,20 @@
+# fill new blocks in portal frame
+# @s = portal frame marker
+# at @s
```
BluePsychoRanger, line 3: `at @s` isn't very useful here. Perhaps change it to something like "at a potential new portal block"

BluePsychoRanger (changes requested):
There's an issue with how you do load checking and a few vague and incorrect headers.

I'm not sure how I feel about allowing crying obsidian to work for portals. It kinda defeats the purpose of needing to repair ruined portal frames, but that's a design decision and I'm open for discussion.

gm4_shapeless_portals/data/gm4_shapeless_portals/functions/x/build.mcfunction
BluePsychoRanger: This applies to `z` also

BluePsychoRanger (approved):

SpecialBuilder32 (approved):
I tested this out in game and could find no bugs. It fits in very naturally with regular vanilla portal creation. 

Looking over the code, I can't help but feel like it should be possible to accomplish this without summoning the temporary marked entities, but as there is: a reasonable hard limit, they are summon very infrequently, and don't even exist for the whole tick - I think the simplicity of using them is just fine.

Bloo-dev:
This module will have to be updated to 1.20, as some of the Advancement format changed.

Bloo-dev (approved):

PR #847, opened 2023-02-14
Module Marketing Readmes

SpecialBuilder32:
This PR seeks to unify and add download-page style descriptions of our modules for use on sites like PMC, Smithed and Modrinth. These descriptions are generated from the module's README.md, and formatted according to the available features of their respective platform. PMC is also roughly translated to the BBCode format. A number of formatting codes are used in the generation and are as follows. 

### Formatting Codes
`<!--$localAssetToURL-->` : Converts a local file embed to point to the raw.githubusercontent address on our, on either markdown images `[alt_text](img_src)` or html tag ``. Points to module-level folder in our repo. 
    ex) `Chairs Any Stair Example <!--$localAssetToURL-->` in gm4_chairs becomes !`[Chairs Any Stair Example](https://raw.githubusercontent.com/Gamemode4Dev/GM4_Datapacks/master/gm4_chairs/images/chairs_any_stair_example.png) `

`$dynamicLink:gm4_pack_name` : Converts a markdown link `[link_text]($dyna...)` to point to that module's download page on whatever site the readme is generated for. For example, in the smithed readme, this link points to the smithed download page. Requires the link data to be present in the `beet.yaml` file, under `meta.modrinth.project_id`. 
   *note: PMC and Smithed APIs are under flux, so `beet.yaml` format for these fields will update in the future to reflect their appropriate api names

`<!--$headerTitle-->` : Inserts the standard header, with image and formatting. Takse preceeding text as the proper name of the module
   ex) Boots of Ostata 

`<!--$footerInsert-->` : Inserts the standard footer, summarizing gm4's mission statement

`<!--$youtubeLinkInsert-->` : Injects the "Watch on Youtube" image and line, using the YT url in `beet.yaml`

`<!--$wikiLinkInsert-->` : Injects the "Read on Wiki" image and line, using the wiki url in `beet.yaml`

`<!--$creditsInsert -->` : Injects credits data, according to names in `beet.yaml` and links to profiles in `contributors.json`

`<!--$modrinth:replaceWithVideo-->` : replaces this line with a youtube video iframe for the video url in `beet.yaml`. If the field is null, leaves the line intact. Intended for the module gif at the top of each readme. 

`<!--$pmc:delete-->` : Deletes the line in the PMC readme. Intended for any images or title content otherwise duplicated in planetminecraft, either in the gallery or in the post title.

`<!--$pmc:headerSize-->` : Increases the font size of the line to 14pt. Intended for the short "eye catching" module summary at the top of the readme.

[...]

PR #856, opened 2023-03-08
Versioning and release

### LINKED ISSUE #818: Custom Crafters is not installed
**Describe the bug**
When installing a datapack that requires lib_machines (for example Liquid Tanks), without lib_machines installed, it gives an error telling you you need both lib_machines and Custom Crafters, while Custom Crafters isn't available anymore.

**To Reproduce**
1. Install a datapack requiring lib_machines, without lib_machines
2. `/reload` to see the install errors
3. Install lib_machines
4. `/reload` again and it should be installed properly without needing Custom Crafters

**Expected behavior**
The same, but without the error of the Custom Crafters requirement

**Fix**
Remove the Custom Crafters check?

**Screenshots**
[https://cdn.discordapp.com/attachments/801191203966156891/1037709606031470692/unknown.png](https://cdn.discordapp.com/attachments/801191203966156891/1037709606031470692/unknown.png)

**Environment**
 - Minecraft version: 1.19.2
 - Server version:
 - Installed Datapacks: base, Liquid Tanks
 - Installed Mods:

SpecialBuilder32:
# Overview
This PR overhauls how we handle versioning in our datapacks, with two primary goals.
1. Modrinth and Smithed require semver-style versioning for uploads
2. Automating our manual versioning control methods
Additionally, it completes the upload automation pipeline to smithed and modrinth utilizing their respective APIs, making use of the features from #847 for their description data.

## Base Changes
The gm4 base module is now version 1.3, with improved error logging for available updates or library conflicts
- Outdated module messages all get printed at once in a long list to prevent chat spam 
image
- Library version conflicts are printed with more data about which module the erring library is bundled inside 
image
- Removes the hardcoded 'gm4-1.2` version number from the namespace (part of later version generation plugins)
- Uninstalled modules no longer can cause "out of date" notices in chat
- Fixes #818 and related library install confusion

## Beet.yaml config changes
- Moves most existing data to a `meta.gm4.website` field, such as `description`, `recommended`, `notes`, `hidden` that are used purely by the gm4.co website
- Required modules and libraries are now listed along with their minimum version number in `meta.gm4.versioning.required`. Note that previously libraries were not listed here. ex) 
```meta:
  gm4:
    versioning:
      required:
        - gm4_liquid_tanks: 2.1.0
        - lib_brewing: 1.0.0
```
- Any schedule-clocked functions are now listed in the config

## Module Lantern-Load autogeneration
- Module no longer have `load` directories. The tags previously stored there are now generated on build from the dependency information in beet.yaml. 
- Modules no longer have `load.mcfunction`. This is also generated on build. To ensure modules properly disable themselves when they fail to load, schedule-looped functions are listed in the beet.yaml under `versioning/schedule_loops`. The namespace is assumed to be the same as the project id unless otherwise specified.
- All advancements in a module have version-checks injected into their conditions on build. This ensures advancements are not earnable if the module failed to load properly.
- Any optional cross-module interaction can still be detected with `if score module load.status matches 1..`, which will only return true for properly installed and running modules

## Library LL autogeneration
- Similar to modules, libraries have much lantern load information autogenerated, such as `enumerate`, and `resolve_load`.
- Libraries no longer have hard-coded version number in their directory names, rather the rename happens on build. References to functions within files are likewise updated automatically. 
- Libraries have an additional config for injection version data: `extra_version_injections`, which is useful for any functions or advancements which might also require version checks, like `resolve_tick` when using lib_forceload's command block. Libraries advancements do not have version checks automatically injected like modules do due to complexities with their inclusion method. 

### Separated lib_custom_crafters out of lib_machines
- `lib_custom_crafters` is now its own library, separate from lib_machines. This greatly simplifies versioning, and allows modules which require lib_machines but have no recipes (ie Soul Glass) to avoid unneeded bloat. 

# Guide to Incrementing Version Numbers
Here is a brief flowchart to determine if you need to increment a version number: 
1. Did your changes add support for a new Minecraft version. Y? -> Increment the Minor value
2. Did your changes add a new significant feature that other modules may wish to interact with? ex, new potions to Zauber that LT could use. Y? -> Increment the Minor value
3. Did you completely rewrite something important that requires other modules to also change to continue working? ex) Custom Crafters switching to use loot tables. Y? -> Increment the Major value and reset the Minor to 0.

Like before, the patch number for modules is automatically incremented whenever that module receives a commit, and so the version number in `beet.yaml` is formatted like `1.2.X`, with the `X` character serving as a reminder to the developer of this fact.

# Guide to updating dependency version numbers
How do you know what version number to require in the beet.yaml? The best rule-of-thumb would be to set it to whatever version number is the most current when developing that module. Ie, if `lib_custom_crafters` is `3.0.0` when the module is written, set the required version to `3.0.0`. That value can be left alone until the module's custom crafters features are changed again in the future.
Even if `lib_custom_crafters` gets a `v3.1.0` at some point, leave the module at `3.0.0` until any of its own custom recipes are changed. If in doubt when updating a module's features, just set its dependencies to the current version.


RitikShah:
*Side note, I find `util.py` type modules in python called `utils.py`. Do with that as you please lol.*

gm4/plugins/manifest.py
```diff
@@ -65,35 +62,43 @@ def update_patch(ctx: Context):
...
 		else:
-			# Changes were made or this is the first release, bump the patch
-			patch = released["patch"] if released else prefix
-			module["patch"] = patch + 1
-			print(f"[GM4] Updating {id} to {patch + 1}")
+			# Changes were made, bump the patch
+			version = dict(zip(("major", "minor"), map(int, module["version"].split(".")[0:2])))
+			last_ver = dict(zip(("major", "minor", "patch"), map(int, released["version"].split("."))))
+			
+			if version["minor"] > last_ver["minor"]:
```
misode, line 95: This also needs to check for major version bumps. If the logic gets too complicated there is probably a semver package out there.

SpecialBuilder32: Oh shoot you're right. Since a major increment necessarily comes with a minor reset, I was originally going to make it whenever the minor version changes be the trigger. But I should just be able to check if major increment or if minor increment

gm4/plugins/output.py
```diff
@@ -51,46 +59,126 @@ def release(ctx: Context):
...
+			if not (200 <= res.status_code < 300):
+				print(f"[GM4] [Smithed] Failed to delete version version: {res.status_code} {res.text}")
+			print(f"[GM4] [Smithed] {res.text}")
+		
+		# post new version
+		res = requests.post(f"{SMITHED_API}/packs/{smithed_id}/versions",
+		    params={'token': auth_token, 'version': version},
+			json={"data":{
+				"downloads":{
+					"datapack": f"https://raw.githubusercontent.com/Gamemode4Dev/GM4_Datapacks/release/{SUPPORTED_GAME_VERSIONS[0]}/{file_name}",
```
misode, line 169: Use `os.getenv("VERSION", "1.19")` for this. It's possible our supported versions in the future could be `[1.19.3, 1.19.4]` or similar.

SpecialBuilder32: Will do. That is much nicer

Smithed right now only has options for [1.18, 1.18.1, 1.18.2, 1.19] for its supported versions, hence why I've used this kinda ugly `SUPPORTED_GAME_VERSIONS[0]`. I'm hoping to let it just use the whole list in the future when smithed comes out of beta again.

gm4_animi_shamir/beet.yaml
```diff
@@ -2,19 +2,22 @@ extend: '../module.yaml'
...
 
 meta:
   gm4:
-    description: Adds the Animi Shamir to Metallurgy. Items with Animi will respawn with you when you die!
     required:
-      - gm4_metallurgy
-    recommended: []
-    notes:
-      - Modifies items right after a player died, and might fight over those items with other Datapacks that do the same.
+      - 'gm4_metallurgy: 1.2.0'
```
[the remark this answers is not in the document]
SpecialBuilder32: Is it considered, proper? to have the yaml keys be predetermined? Like the module here listed becomes the config option rather than the value. 

It's possible I just don't know the conventions.

lib_custom_crafters/beet.yaml
```diff
@@ -0,0 +1,31 @@
...
+    extend: 'beet.yaml'
+  - data_pack:
+      load: .
+  - lib_custom_crafters.generate_item_tags
+
+meta:
+  gm4:
+    required: ['lib_machines : 1.1.0']
+    schedule_loops: [main]
+    extra_version_injections:
```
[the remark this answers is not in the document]
SpecialBuilder32: The only reason I didn't do this was that the required field is also used by the website but I agree this would be better

gm4/utils.py
```diff
@@ -0,0 +1,17 @@
...
+	"""Run a shell command and return the stdout."""
+	return subprocess.run(cmd, capture_output=True, encoding="utf8").stdout.strip()
+
+def semver_to_int(version: str) -> int:
+	"""converts a semver string into a integer, for use in module scoreboards"""
+	return sum(map(lambda x,y: int(x)*y, version.split("."), (100_000, 1_000, 1)))
+
+def semver_to_dict(version: str) -> dict[str, int | None]:
+	"""converts a semver string into a dict with each of its components"""
+	return dict(zip(("major", "minor", "patch"), map(X_int, version.split("."))))
```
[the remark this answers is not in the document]
SpecialBuilder32: Made this change. It kinda felt like a weird data class but I like the syntax better

PR #877, opened 2023-06-08
BAS: interaction entities

Dennis-0:
Replaced the use of wandering_trader for left/right click detection with interaction entities.


SpecialBuilder32 (approved):
Tested briefly ingame and it all worked as expected!

Issue #875, opened 2023-06-05
Gamemode 4 Welcome Toast Condition


Bloo-dev:
The Gamemode 4 Welcome Toast (introduced in PR #844) is currently set to only play when the song plays, however, it should play upon first join, not only on Gamemode 4's first install.

This issue is a notice to fix this behavior later on.

PR #878, opened 2023-06-08
Intro Toast Visibility

Bloo-dev:
This PR addresses issue #875.

Moves the intro toast to a loaction-triggered advancement, thereby making it **appear upon first login** for every player, letting them know that Gamemode 4 is installed. Also **updates the text** of this 'Welcome Toast' to use its own translation key and fallback text.

Feel free to discuss the specific fallback message below, especially (check if resolved):
- [x] Datapacks or Data Packs?
- [x] Should we use a location trigger, or use an `impossible` trigger and grant it via functions? The latter would allow for a score-based toggle and/or a score based check instead of relying simple advancement existence.

image showing the toast popup

### DIFF
--- a/base/data/gm4/advancements/intro_song/root_display.json
+++ b/base/data/gm4/advancements/intro_song/root_display.json
@@ -17,7 +17,7 @@
   "parent": "gm4:intro_song/invisible_root",
   "criteria": {
     "requirement": {
-      "trigger": "minecraft:impossible"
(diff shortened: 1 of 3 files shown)


Bloo-dev:
I have updated the toast message and added a load score check to make sure the advancement doesn't grant itself if the base does not load correctly, as is standard with other Gamemode 4 advancements.

PR #881, opened 2023-06-10
Rope Ladders Update

MichaelMiner137:
- Replaced villager with interaction entity.
- Ladder can now extend into any replaceable block.
- Players in adventure mode can no longer extend ladders.
- Fixed bug related to incorrect ladder stack being extended.

### DIFF
--- a/gm4_rope_ladders/data/gm4_rope_ladders/functions/mechanics/ladder_placement/place_ladder.mcfunction
+++ b/gm4_rope_ladders/data/gm4_rope_ladders/functions/mechanics/ladder_placement/place_ladder.mcfunction
@@ -12,6 +12,10 @@ execute if entity @s[gamemode=!creative] run item modify entity @s weapon.mainha
 # grant advancement
 advancement grant @s only gm4:rope_ladders
 
+# break block
+execute unless block ~ ~ ~ #gm4:water run setblock ~ ~ ~ air destroy
+execute if block ~ ~ ~ #gm4:water run setblock ~ ~ ~ water destroy
+
 # place ladder
 execute unless block ~ ~ ~ #gm4:water run clone ~ ~1 ~ ~ ~1 ~ ~ ~ ~
 execute if block ~ ~ ~ #gm4:water run function gm4_rope_ladders:mechanics/ladder_placement/place_water_ladder
--- a/gm4_rope_ladders/data/gm4_rope_ladders/functions/mechanics/ladder_placement/place_ladder.mcfunction
+++ b/gm4_rope_ladders/data/gm4_rope_ladders/functions/mechanics/ladder_placement/place_ladder.mcfunction
@@ -13,5 +13,5 @@ execute if entity @s[gamemode=!creative] run item modify entity @s weapon.mainha
 advancement grant @s only gm4:rope_ladders
 
 # place ladder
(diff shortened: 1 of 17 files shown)


gm4_rope_ladders/data/gm4_rope_ladders/tags/blocks/water.json
```diff
@@ -0,0 +1,6 @@
+{
+  "values": [
+    "water",
+    "bubble_column"
+  ]
+}
```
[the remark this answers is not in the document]
MichaelMiner137: I didn't know that was a block tag that existed.

SpecialBuilder32 (changes requested):
Looks good! Only comment I have is that while standing above its difficult to target the interaction entity. Perhaps the height could be adjusted to be the full block heignt?

image

MichaelMiner137:
I made the hitbox extend all the way up, and even slightly above the top for easier interaction.
image

Bloo-dev (approved):
I tested this in-game and everything still works as expected.

I found some non-blocking issues that we should deal with in the near future, or in this PR if there's enough time before this is merged:
- when placing ladders, the setblock mode is currently `replace`, which means that tall grass or seagrass simply disappear, without any breaking sound or particles. Simply setting this to `destroy` would fix this.
- the block tag `gm4_rope_ladders:valid_replace` only contains `gm4:replaceable` and is there somwhat pointless. This could be removed.

Bloo-dev:
@SpecialBuilder32 please verify that your hitbox concerns have been resolved.

Bloo-dev (approved):

gm4_rope_ladders/data/gm4_rope_ladders/tags/blocks/valid_replace.json
TheThanathor, line 1: this file isn't used anywhere and could be removed

gm4_rope_ladders/data/gm4_rope_ladders/functions/mechanics/ladder_placement/place_ladder.mcfunction
```diff
@@ -12,6 +12,10 @@ execute if entity @s[gamemode=!creative] run item modify entity @s weapon.mainha
...
 advancement grant @s only gm4:rope_ladders
 
+# break block
+execute unless block ~ ~ ~ #gm4:water run setblock ~ ~ ~ air destroy
+execute if block ~ ~ ~ #gm4:water run setblock ~ ~ ~ water destroy
+
 # place ladder
-execute unless block ~ ~ ~ water run clone ~ ~1 ~ ~ ~1 ~ ~ ~ ~
-execute if block ~ ~ ~ water run function gm4_rope_ladders:mechanics/ladder_placement/place_water_ladder
+execute unless block ~ ~ ~ #gm4:water run clone ~ ~1 ~ ~ ~1 ~ ~ ~ ~
```
TheThanathor, line 20: technically this means that if the ladder above is waterlogged it will incorrectly copy a waterlogged ladder. However the only method I found to do this is by replacing light blocks (as they are replaceable but stop water flow), so it's not a big deal.

TheThanathor (approved):
two tiny comments, but nothing that's very important to fix

SpecialBuilder32 (approved):

PR #882, opened 2023-06-11
Orb of Ankou 1.20

BluePsychoRanger:
Feature Complete for 1.20:
- Sniffers can drop pneumas (Retreating and Sniffing)
- Sniffing (sneak hold) allows the player to sniff up torchflower seeds and pitcher pods. Once one is dug up there's an 8 minute cooldown. Sneak must be held for at least 5 seconds before an item can be dug up.
- Camels can drop pneumas (Retreating and Dashing)
- Dashing (sneak charge) allows the player to dash up to about 12 blocks forward (depending how long they sneak) and a bit more than 1 block up

Changes:
- Support for up to Looting X (previously anything above looting III wouldn't drop anything)
- Magma Cube drop rates reduced
- Dolphin drop rates reduced
- Axolotl drop rates increased

Technical:
- The beet script now reads a CSV file in the source code and generates the loot tables for each mob. This was previously done with an offline script and separate sheet.
- The beet script generates the predicates that indicate if a player has a pneuma equipped

Bug Fixes:
- Some pneumas were broken in 1.19.4 due to damage type changes
- Sneak charge pneuma would activate when the sneak number was greater than 3 which is too low after the tick change in #780  (it wouldn't do anything, but it was unnecessary code being run)


gm4_orb_of_ankou/soul_essence.csv
```diff
@@ -0,0 +1,182 @@
+-Run beet to update columns 4-9
```
RitikShah, line 1: I had no idea you could add comments to csvs

BluePsychoRanger: It's a normal cell actually, I just skip the first few lines when parsing it then print that out when rewriting it

Bloo-dev:
That's a nice use of a csv! I'll test this in-game this weekend.

TheThanathor (approved):
Tested this in game and everything works as expected

didn't take a full look at the python file, but from the output that seems to work as well

Bloo-dev (approved):

PR #887, opened 2023-06-13
Fix Arborenda Unbreaking Behaviour

### LINKED ISSUE #836: Arborenda incorrectly factors unbreaking into durability decreases
**Describe the bug**
Arborenda always takes away 1 durability from an axe for every block broken. This is due to a math error in the unbreaking-durability calculation done in `roll_binomial_distribution.json`

**Expected behavior**
Arborenda has code that uses the binomial number provider to give the unbreaking I, II, and III enchantment a 50%, 33% and 25% chance to remove a durability for each log broken. An unbreaking III axe should take much less damage than an unenchanted axe when breaking the same quantity of wood. 

**Fix**
Add `scale: 0.01` to the probability score provider to properly specify probability. Currently the passed values of 100, 50, 33, or 25 correspond to a 100% probability for each log to take durability, being greater than one. 

**Additional Notes**
This simple bug was not just patched in a short commit because it was discovered as a part of a larger bug investigation into paper/vanilla discrepancies on loot tables that spawn items with `count:0b`, which caused additional issues on the current Public Server where no arborenda axe takes durability from the logs it breaks.

Bloo-dev:
Fixes #836, which caused Arborenda to effectively ignore Unbreaking.

Note that Arborenda still only uses a single durability point per tree on Spigot due to a Spigot bug.

### DIFF
--- a/gm4_metallurgy/data/gm4_arborenda_shamir/loot_tables/roll_binomial_distribution.json
+++ b/gm4_metallurgy/data/gm4_arborenda_shamir/loot_tables/roll_binomial_distribution.json
@@ -17,7 +17,8 @@
                         "type": "minecraft:fixed",
                         "name": "$damage_chance"
                     },
-                    "score": "gm4_arb_data"
+                    "score": "gm4_arb_data",
+                    "scale": 0.01
                 }
             },
             "entries": [


Bloo-dev:
@SpecialBuilder32 summoning you as you originally discovered this bug.

SpecialBuilder32 (approved):

PR #888, opened 2023-07-26
Player head management plugin

SpecialBuilder32:
This PR is the ~~*mostly finished*~~ beet tooling that helps manage the custom-player-head-skins we use throughout our datapacks. It replaces the need for the developer to manually transform skin images into the ~~gibberish~~ base64 string that represents the image in an item's nbt, instead storing the skin image file directly in the source code, and using mineskin's api when a skin needs to be sent to Mojang's servers. 

This PR also introduces Mecha to the project for the first time, and serves as a prototype for the eventual CustomModelData handler that is planned down the road. 

## Usage
Skin files placed in the `data/namespace/skins` directory are accessible via their file names when writing nbt compounds:
```{SkullOwner:"$gm4_heart_canisters:heart_canister_tier_1"}```

Skin references are prefaced by `$`, and will assume the namespace to be the `project_id` (usually the root folder name) if omitted. The following shorthand and regular forms are accepted, and the `Name` and `Signature` fields are optional, to be used as needed by modules like lib_machines
- `{SkullOwner:"$skin_name"}`
- `{SkullOwner:{Value:"$skin_name"}`
- `{SkullOwner:{Value:"$skin_name", Name:"foo", Signature:"bar"}`
- `{SkullOwner:{Name:"foo", Properties:{textures:[{Value:"$skin_name", Signature:"bar"}]}}}`

Although mecha does not currently parse and process non-function files, nbt fields within loot tables, item modifier and advancements are manually passed into the mecha parser and will be properly filled out.

### Skin cache
All data about skins and their in-game texture values is stored in a `skin_cache.json` file. The hash of each skin image is used to ensure api calls are not duplicated and successive `beet build` calls produce consistent item data. If a developer is changing skin textures, this file will be changed by beet, and must be included in commits to the github repo. 

Since the github build action won't modify the master branch of the repo, it is unable to upload skins to mineskin and cache the result. Developers changing skin images must therefore run at least one `beet build` or `beet dev` on affected modules locally after editing images, and commit the resulting cache file. The GitHub action is marked as failed if it finds a skin image that has *not* been uploaded to mineskin by a developer.

If skin textures are removed from the repo, the skin cache will *not* be automatically cleaned up, and a message informing the developer of the unused cached texture data emitted to the log. Manually editing the skin cache to remove the unneeded entry is suggested

### Mineskin authentication
Since uploading skins to mineskin's api requires an api key, the beet build process will ask for an authentication token in the terminal if any skin file changes are made. Obtaining a token is simple; simply sign up for a mineskin account using a google account login, and create a token. It is *not* necessary to provide mineskin with your minecraft account details, but you may if you wish your account to be used to upload the skin changes and not queued to one of the many volunteer accounts mineskin has in its possession. 

Your authentication token will be cached locally by beet in `.beet_cache/mineskin/index.json`, and can be manually deleted or removed by using `beet cache --clear`. 

### Skin() class
Beet plugins can programmatically create and edit skin files much like other image files in the beet ecosystem. The `gm4.plugins.player_heads` python module defines a `Skin` class which attaches to the container skin image files are mounted to automatically. 
```py
from gm4.plugins.player_heads import Skin
from PIL import Image
def beet_default(ctx: Context):
    ctx.data[Skin]["gm4_end_fishing:enderpuff"] = Skin(Image.new("RGB", (64, 64), color="lime"))
```

## Mecha Details
So this PR introduces mecha to the project for the first time, and that both opens the floodgates for loads of new powerful beet tools (like bolt), and adds some more complexity to be aware of. Here are some of the details I thought were important to share. 

Mecha is a parser, which means that it reads in all function files, processes all the commands, nbt data ect into an abstract syntax tree. This means the build process now can detect most syntax errors without needing to check the output log of minecraft (at least in function files). You'll see nicely formatted diagnostic errors in the terminal for syntax errors.

We can write rules that interact with nodes of this tree very nicely; this PR adds a rule that edits specifically the `SkullOwner` nbt nodes. As it is now, mecha doesn't do anything with comments, and by default strips them out of the built files. There is a "development only" option to preserve the comments and newlines which I have applied to our project, but this may result in strange spacings if we dive more into bolt modules in the future. There are some plans to add a third formatting option that keeps comments and does bolt spacing better, but that is just a wishlist for now. Just keep in mind this fact if you start doing weird things

At the moment, mecha is only run on module files, not libraries (except CC). This is mostly to is to keep the nice isolated-subproject nature of the build process, and to keep processing to a minimal since libraries don't need mecha atm.


gm4_audere_shamir/data/gm4_audere_shamir/functions/init.mcfunction
```diff
@@ -6,7 +6,7 @@ scoreboard players set audere_shamir gm4_modules 1
...
+execute unless data storage gm4_player_heads:register heads[{id:"gm4_audere_shamir:band/v1"}] run data modify storage gm4_player_heads:register heads append value {id:"gm4_audere_shamir:band/v1",name:"[Drop to Fix Item] gm4_audere_shamir:band/v0",item:{gm4_metallurgy:{has_shamir:1b,stored_shamir:"audere",metal:{type:"barium",amount:[12s],castable:1b},item:"obsidian_cast"},SkullOwner:"gm4_metallurgy:band/barium",CustomModelData:3420120,display:{Name:'{"italic":false,"translate":"item.gm4.metallurgy.obsidian_cast","fallback":"Obsidian Cast"}',Lore:['{"italic":false,"color":"#F0EAD6","translate":"item.gm4.metallurgy.band","fallback":"Barium Band","with":[{"translate":"item.gm4.metallurgy.barium"}]}','{"italic":false,"color":"aqua","translate":"item.gm4.metallurgy.shamir","fallback":"Shamir"}','{"italic":false,"color":"gray","translate":"item.gm4.shamir.audere","fallback":"Audere"}']}}}
```
Bloo-dev, line 9: Isn't there a `$` sign missing infront of `gm4_metallurgy:band/barium`?

SpecialBuilder32: Probably yes, but I think all the player-heads stuff is a moot point anyway right now since its all done improperly. I think I'd like to fix lib player heads in another PR however.

PR #889, opened 2023-07-28
Publish libraries on smithed

SpecialBuilder32:
This PR adds a step to our build pipeline that will upload standalone copies of our libraries to smithed for other developers to use :tada:

Although smithed has some dependency-listing capability, these library zips will come with any sub-libraries already included (lib_machines will include lib_forceload for example), since that works very cleanly with our build configs

As a part of the PR, we will be relicensing the libraries on our main repo under MIT. This was discussed and happened on the now abandoned separate repos some of our libraries can be found on in the past, but never was completed. All other content in this project will remain licensed under GPLv3, and only the contents of `lib_*` folders (which will each contain a copy of the MIT license) will be affected. As such all contributors to the libraries will be asked to approve this PR.


gm4/plugins/output.py
```diff
@@ -68,7 +68,26 @@ def release(ctx: Context):
 
 	# publish to download platforms
 	publish_modrinth(ctx, release_dir, file_name)
-	publish_smithed(ctx, release_dir, file_name)
+	publish_smithed(ctx, file_name)
+
+
+def libraries(ctx: Context):
+	"""Saves the zipped standalong library to FIXME"""
```
Bloo-dev, line 75: I would assume `FIXME` is a placeholder?

SpecialBuilder32: Yes that's correct. I was planning to save the build to `release/libraries/lib_machines` but I'm not so sure about that.

misode:
I agree to license the libraries as MIT

Luexa:
I approve the relicensing to MIT. My contributions may be freely used under the MIT License.

Bloo-dev:
I agree to relicensing the libraries to MIT.

BluePsychoRanger:
I agree to relicensing the libraries to MIT

BluePsychoRanger:
Is there a reason the MIT license uses 2021? I'm not sure how the year works tho. If it's the first year the license is in affect or if it's the latest year the license has updated?

misode:
> Is there a reason the MIT license uses 2021? I'm not sure how the year works tho. If it's the first year the license is in affect or if it's the latest year the license has updated?

Some people update it every year, others keep it as the initial year. Afaict it doesn't really matter (not legal advice).

PR #890, opened 2023-07-28
Add Double Doors

Bloo-dev:
This PR adds the **Double Doors** module to the repository.

### Progress
- [x] Implement base case
- [x] Generate other material types using beet
- [x] Populate `beet.yaml`
- [x] Populate `README.md`
- [ ] Add Module Icon
- [ ] Add Guidebook information

## Features
This module allows players to open double doors with ease! Instead of having to click twice, clicking one of the doors will **automatically open the other**!
Additionally, this module also allows trapdoor placed above doors to open and close with the door! For this to work, the trapdoor must be a bottom half trapdoor, open, and its orientation must match the orientation of the door in at least one of the two positions.

Doors are detected as double doors if:
- both doors/trapdoors are of the same wood type
- **and** the non-hinge end of both doors touches whilst both doors are either open **or** closed (closed doors that are technically `open=true` are allowed, but mixing of technically open and truely open is not allowed)

Trapdoors are opened automatically, if:
- the trapdoor is placed above a door (or trapdoor) of the same wood type
- **and** the trapdoor is a bottom-half trapdoor (allows player to use top-half trapdoors for non-opening windows above doors)
- **and** the trapdoor is open
- **and** the trapdoor matches at least on of the possible orientations of the door below
- **and** the trapdoor stack above the door is at max 2 trapdoors tall (this can be configured by setting the fake player `$trap_door_limit` in `$gm4_double_doors_data`)

image


Bloo-dev:
It's not a surprise that the test@v1 check failed, the module is not executable as a data pack unless built using beet.

gm4_double_doors/generate.py
```diff
@@ -0,0 +1,99 @@
+from beet import Context, Function, Advancement, Structure
+from typing import Dict, List
+from dataclasses import dataclass
+
+NAMESPACE = 'gm4_double_doors'
```
SpecialBuilder32, line 5: This is probably better-accessed with `ctx.project_id`

gm4_double_doors/generate.py
```diff
@@ -0,0 +1,99 @@
...
+
+    advancements: Dict[str, Advancement]
+    functions: Dict[str, Function]
+    structures: Dict[str, Structure]
+
+
+def beet_default(ctx: Context):
+
+    # list of supported wood types. Add new wood types here.
+    material_names: List[str] = ['acacia', 'bamboo', 'birch', 'cherry', 'crimson', 'dark_oak', 'jungle', 'mangrove', 'oak', 'spruce', 'warped']
```
SpecialBuilder32, line 22: We can automatically pull this from the `beet.contrib.vanilla` Perhaps taking the wood types from `#minecraft:wooden_doors`? There might be a better source of just the wood types without extra *stuff* like the `_door`

gm4_double_doors/README.md
```diff
@@ -0,0 +1,9 @@
+# Bat Grenades<!--$headerTitle--><!--$pmc:delete-->
```
SpecialBuilder32, line 1:
```suggestion
# Double Doors<!--$headerTitle--><!--$pmc:delete-->
```

SpecialBuilder32 (commented):
Hey Bloo! Welcome to my ~~hell~~ world of beet plugins!

I think to get this to pass the github action, we might be able to move the template files out of the "executable" directories and into their own `templates` folder perhaps. 

But while on that topic, beet has jinja templating available and I suspect that would be a really nice way to do this whole "clone but replace" thing. I've never understood templating though so I've asked for some help and am doing some more looking into it.

Bloo-dev:
This now uses jinja templating implemented by @SpecialBuilder32.
I've tested @SpecialBuilder32 's version (the latest commit) in game and all is working fine.

Note:
This required an update to our beet version to allow access to the latest wood types (bamboo and cherry) via the vanilla beet plugin. Running `poetry install` and `poetry cache --clear` locally may be required on your end.

gm4_double_doors/README.md
```diff
@@ -0,0 +1,9 @@
+# Double Doors<!--$headerTitle--><!--$pmc:delete-->
+
```
SpecialBuilder32, line 2: I would *love* it if you could make a quick screenshot of the "3-tall" doors for here. (Or, even better, a GIF of the opening if you have time)

SpecialBuilder32 (approved):
I already reviewed this module and helped with its beet generation stuff, so it all looks good to me. I wish we had module icons for new modules but I'm also dont want that to be a blocking issue forever.

gm4_double_doors/README.md
Bloo-dev: A gif for this is in the works with @Dennis-0. ETA is next week!

PR #892, opened 2023-08-09
Pirate Translations

BluePsychoRanger:
Pirates from Tower Structures were spawning with old or nonexistent translation tags. This fixes the spawners to spawn skeletons with the correct nbt.
- Bows and Swords given to the skeletons have proper translation tags, so death messages appear correctly
- Pirate Hats (and Turtle Helmets) given to the skeletons have proper translation tags (unnecessary cuz it's not visual, but it should be done for consistency)
- Pirates' names use translations instead of plain text
    - Pirates' (default) names end with `§` to allow for future custom entity textures


Bloo-dev:
Tested this in-game and checked out the nbt of the structures, all looking good.

PR #898, opened 2023-09-09
Infinitas Shamir: Update to 1.20 and leftover fixes

EpyonProjects:
- Actually works in 1.20, after including the required location tags
- Addresses SneakingSpider's last mentioned issue regarding milk/cows
  - Turns out milk uses player_interacted_with_entity, not filled_bucket
- Addresses rx's concern about the powder snow toggle
  - It is now off by default
- Updates the bucket shamir nbt in init
  - As a result, this autoupdates pre-existing Infinitas buckets after use

### DIFF
--- a/gm4_metallurgy/data/gm4_infinitas_shamir/advancements/empty/pickup_mainhand.json
+++ b/gm4_metallurgy/data/gm4_infinitas_shamir/advancements/empty/pickup_mainhand.json
@@ -21,13 +21,18 @@
             }
           }
         ],
-        "location": {
-          "block": {
-            "blocks": [
-              "minecraft:cauldron"
-            ]
+        "location": [
+          {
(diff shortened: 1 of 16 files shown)


Bloo-dev:
Works as advertised, therefore it is ready to be merged.

A future PR should make it so that the lava toggle is loaded from the gamerule _once_ upon first install. The same goes for a water toggle that should be added:
```
execute unless score $water_infinitas gm4_ml_data matches 0..1 store result score $water_infinitas gm4_ml_data run gamerule waterSourceConversion
execute unless score $lava_infinitas gm4_ml_data matches 0..1 store result score $lava_infinitas gm4_ml_data run gamerule lavaSourceConversion
```
At that point, only Milk wouldn't have a toggle so I suggest also adding one for milk.

PR #899, opened 2023-09-18
Pydantic models for Plugin Config

SpecialBuilder32:
This PR refactors how our `beet.yaml` metadata is read into many of our build plugins, using the intended system of pydantic models rather than arbitrary dict access. This change allows for a much clearer definition of what options are available/expected and in what format. 

Also included in this PR (and its impetus), is the addition of the `search_keywords` config option for the website.


RitikShah:
I'll skimmed through it, happy to review when you think it's ready. I would stick with either `Optional[type]` or `type | None` (whichever you prefer),

BTW, `beet` is still on Pydantic v1.x and not v2.x so make sure you are reading the correct docs for that!

Bloo-dev:
I can't find any reference to `search_keywords` in the code of this PR, did you forget to commit that?

PR #907, opened 2023-10-16
Bookshelf Inspector

TheThanathor:
Shows a nametag when hovering over a book in a Chiseled Bookshelf


gm4_bookshelf_nametags/data/gm4_bookshelf_nametags/functions/evaluate/get_pos.mcfunction
```diff
@@ -0,0 +1,9 @@
+# spawn marker to get player position
+# @s = marker
+# at unspecified
+# run from evaluate/motion
+
+tp @s @p[tag=gm4_bookshelf_nametags_target]
```
Bloo-dev, line 6: Whilst it doesn't make a difference anymore, I'd prefer to use `@a` instead of `@p` for readability. (not critical)
```suggestion
tp @s @a[tag=gm4_bookshelf_nametags_target,limit=1,sort=nearest]
```

gm4_bookshelf_nametags/data/gm4_bookshelf_nametags/functions/evaluate/motion.mcfunction
```diff
@@ -0,0 +1,25 @@
...
+# at unspecified
+# run from evaluate/run
+
+# store player motion using a marker to avoid reading player NBT
+tag @s add gm4_bookshelf_nametags_target
+execute summon marker run function gm4_bookshelf_nametags:evaluate/get_pos
+tag @s remove gm4_bookshelf_nametags_target
+
+# round down to lowest integer, if all are 0 consider player to not be moving
+execute store result score $Mot_0 gm4_bookshelf_nametags_data run data get storage gm4_bookshelf_nametags:temp Motion[0]
```
Bloo-dev, line 12: Fake players should be in `snake_case` and not shortened unless absolutely required; furthermore, translating 0, 1, and 2 to x, y, and z would make this easier to understand i.e. `$motion_x` instead of `$Mot_0`.

gm4_bookshelf_nametags/data/gm4_bookshelf_nametags/functions/evaluate/get_pos.mcfunction
```diff
@@ -0,0 +1,9 @@
...
+data modify storage gm4_bookshelf_nametags:temp Motion set from entity @s Motion
```
Bloo-dev, line 7: You're technically reading player NBT twice here, once for `Motion` and once for `Rotation`. It could be cheaper to read it once and then extract `Motion` and `Rotation`, I don't fully remember though.

Bloo-dev (commented):
This is a preliminary review, someone with a better understanding of display entities will have to take a look at that part of the module. Nevertheless, here are some comments mostly regarding form, not function.

We may also want to consider another name for the module, as Nametags are an item in game.

gm4_bookshelf_nametags/data/gm4_bookshelf_nametags/functions/evaluate/get_pos.mcfunction
TheThanathor: I've changed this, but it does require me to read the entire marker NBT into storage so I'm not sure if it's actually faster

gm4_bookshelf_nametags/data/gm4_bookshelf_nametags/functions/evaluate/get_pos.mcfunction
TheThanathor: Going to disagree with this one, I find the simplicity of not having an extra `,limit=1,sort=nearest` to make it more readable

Bloo-dev (commented):
This has been tested in-game and works as promised. We also tested it in multiplayer and the name plates rendering for everyone seems reasonable considering the limitations we are working with (have you heard back from Compso?).

I do have one more quality-of-life suggestion; Once a player starts inspecting bookhshelves getting nudged slightly or moving a little bit should not immediately throw them out of "inspect mode". How about implementing a "cooldown" so that the player is only thrown out of "inspect mode" once they have moved for more than x ticks?

TheThanathor:
> This has been tested in-game and works as promised. We also tested it in multiplayer and the name plates rendering for everyone seems reasonable considering the limitations we are working with (have you heard back from Compso?).

Comp didn't remember the method.
I figure it might be using translation to keep the display close to the player with a very small `view_range`, might look into that

gm4_bookshelf_inspector/images/bookshelf_inspector.png
Bloo-dev, line 1: A square image would work better on most plattforms.

Bloo-dev (approved):

gm4_bookshelf_inspector/beet.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - gm4.plugins.extend.module
+
+meta:
+  gm4:
+    versioning:
+      schedule_loops: [main]
+    website:
+      description: See what's in your Chiseled Bookshelves without having to take all the books out!
+      recommended: []
+      notes: []
```
SpecialBuilder32, line 18: Since this only works on 1.20.2, there should be a note here that shows on the website

SpecialBuilder32 (approved):
Did a read-through of the code, and it looks pretty optimized in how it does things. Theres alot of clever modern datapack tricks in here that I quite like. 

Just a note for the future: When we release this on modrinth or smithed, there is a yet-unused config field that needs to be set so those sites list this module as only available for 1.20.2. `minecraft: ["1.20.2"]`.

PR #908, opened 2023-10-17
Update 1.20.2

misode:
This PR adds support for 1.20.2 using overlays. This means that data packs in the current version of the repo will work on both 1.20, 1.20.1 and 1.20.2.

The only breaking change were [effect ID changes and NBT tag renames](https://misode.github.io/versions/?id=1.20.2&tab=changelog&search=effect).

These changes were made with the help of [some scripts](https://gist.github.com/misode/50520dea40bf34e96daa5d5f8bc3eab4).


Bloo-dev (approved):
Tested the output both in 1.20.1 as well as 1.20.2 in-game. Both builds succeeded and worked fine in game. I did not test all modules altered by this, but a good chunk of them.

I also had a look through the changed files and couldn't spot any issues. However, I did not check for any missed updates.

As a side note: Building this causes a change to the `skin_cache.json` in which some lines are sorted differently.

PR #913, opened 2023-10-31
Add support for trapped hanging signs

### LINKED ISSUE #876: Trapped Signs missing support for Hanging Signs
This bug report describes and tracks a desired feature set for the Trapped Signs module. 

1.20 added a new set of writable signs, as well as the ability to write on both sides of the ordinary signs. These new variety of signs should be supported by the trapped signs module, as well as the ability to write on both sides. 

Some minor discussion about new features: 
- potentially changing the module from a machine-block placement to a modifier players can apply to existing signs to add the redstone capability

misode:
Closes #876 

Also fixes some issues introduced when adding support for bamboo, cherry and mangrove signs


Bloo-dev:
I tested this in game (I did not test all types of sign & trapped sign, but only tested a couple drawn at random).

Everything works as expected, however, there are raycasting issues: Due to the shape of signs it is quite easy to get the raycast to intercept a neighbouring sign when placing multiple next to each other. This causes the trapped sign not to form properly and the player loses the two trip wire hooks and three redstone.

This issue has been present before this PR though, hence it is **not** blocking. Overall this module may need some reimagining in the future to bring it up to date with our latest design standards or should be retired entirely as there are plenty of button alternatives in the base game these days.

PR #918, opened 2023-11-14
The great Resource Pack Merge

SpecialBuilder32:
This PR merges the [Resource Pack repo](https://github.com/Gamemode4Dev/GM4_Resources) into this repo, building the resource pack with beet in alongside the datapack.

## Resource Pack Feature Overview
All resource-pack features now live in the main repository, inside an `assets` directory of each respective module folder. Each module essentially creates a resource-pack with only its own textures/models, and then beet will collect and merge all the packs into the one master resource-pack we offer for download.

There are a number of development tools added to help manage CustomModelData values, translation keys, and model creation, which will all be detailed in their respective sections below.

When using the `beet dev` command, the combined resource pack for all listed modules will also be output to the `out` directory. Additionally, each module will link their own individual resource packs to the resource pack folder, alongside the combined resource pack under the name `DEV gm4_resource_pack`. This may be useful for debugging potential merge issues.

### CustomModelData `model_data` Config
<details open>
<summary>collapse</summary>
All custom_model_data values are auto-assigned by beet, and replaced by developer-friendly string references in the source code. e.g. `{CustomModelData:'bat_grenades:item/bat_leather'}`. These string references are all defined in the `beet.yaml` configuration, alongside what model file that cmd should display, or information to generate such the model.

`meta.gm4.model_data` is a list of compounds, each of which accepts the following fields
* `item`: (required) item name, or list of item names this CMD will be applied to.
* `reference`: (required) the unique string reference to be used in source files e.g. `item/bat_leather`. If no namespace is specified, the module's namespace will be assumed. 
It is recommended to format references according to the type of item it applies to. e.g. `item/...`, `block/...`, `gui/...` ect, and extra recommended to use the model/texture file path where possible.
* `model`: (optional) the model file(s) to display on items with this CMD. If not specified, will default to the same value as `reference`. 
Accepts a string or list of strings, one for each item present in `item`. May be formatted as a dictionary mapping items to models e.g. 
```yaml
item: [apple, potato]
model:
  apple: my_model_apple
  potato: my_model_potato
```
More complex model predicates for items with multiple vanilla models (e.g. elytra & broken elytra, clock ect) can be specified here as a list of predicates, following the same syntax as model files e.g. 
```yaml
model:
    - predicate: {broken: 0}
      model: item/elytra/captains_wings
    - predicate: {broken: 1}
      model: item/elytra/broken_captains_wings
```
This list of predicates may also be mapped to a specific item as above.
```yaml
item: [apple, elytra]
model:
  apple: my_model_apple
  elytra:
    - predicate: {broken: 0}
      model: item/elytra/captains_wings
    - predicate: {broken: 1}
      model: item/elytra/broken_captains_wings
```
Items who have multiple vanilla models, like clocks, who do not have the manual predicates specified in `model` will have the same provided model file applied to all variants.
* `template` (optional), a model-file generating template to apply. Accepts a string name of a template, or a compound containing template configuration values. Defaults to `custom`, which generates no Model files. See section below for details on available templates.
* `transforms` (optional), a list of model transforms to apply. Accepts a compound of configuration data. See section below for details on available transforms.
* `textures` (optional), a string, list of strings, or mapping of texture files to supply to the template.
* `broadcast` (optional*), a list of sub-configs which will inherit the current config values. This reduces redundancy when defining many similar references. Due to the inheritance, each config need not contain every required field, so long as after the broadcast is carried out all required fields have a value. e.g.
```yaml
model_data:
      - item: player_head
        template: generated
        broadcast:
          - textures: item/band/mundane_band
            reference: item/mundane_band
          - reference: item/lump/baryte
          - reference: item/lump/bauxite
          - reference: item/lump/bismutite
          - reference: item/lump/thorianite
```

</details>

### Model Auto-generation/Templates

<details>
<summary>Expand</summary>

Templates allow for easy generation of common model file structures. Uses the `textures` field (or its default value of `reference`) to find the texture file(s). Available default templates are:
* `custom`: does not generate a model file(s).
* `generated`: generates a model of the `minecraft:item/generated` parent. Often used to turn a single texture image into a flat held item ingame. e.g.
```yaml
- item: firework_star
        reference: item/everstone
        template: generated
```
* `generated_overlay`: generates a model with two layers, where the second texture is the same as the first, followed by `_overlay`. Intended for potions and other similar items
* `handheld`: generates a mode of the `minecraft:item/handheld` parent. Used for turning single textures into a tool item ingame.
* `vanilla`: use the vanilla item model. Often used when no textures exist yet but a CMD value should still be reserved.
* `block`: generates a model of the `minecraft:block/cube` parent. Textures are specified in the order [top, bottom, front, side], or by mapping to their names. e.g.
```yaml
- item: piston
  template: block
  reference: block/forming_press
  textures:
      front: block/forming_press_side
      side: block/forming_press_side
      top: block/forming_press_top_out
      bottom: block/forming_press_bottom
 - item: enchanting_table
   template: block
   reference: block/enchantment_extractor
   textures:
      - block/enchantment_extractor_top_out
      - minecraft:block/furnace_top
      - block/enchantment_extractor_side
      - block/enchantment_extractor_side
```
* `advancement`: generates a model to be used in an advancement icon. This is a different CMD from the actual item to facilitate retexturing just the advancement and not the whole item. Redirects to a specified `forward` model file. This template ignores the `model` config field.
```yaml
- item: gunpowder
  reference: gui/advancement/bat_grenades
  template:
    name: advancement
    forward: item/bat_leather
```
* `legacy_machine_block`: generates a model following the `block` template, but with a set of transformations that scale and translate the model to match the position/scale of the "block on the head of a small armor stand" that is used in so many of our machine blocks.

Modules may create and configure their own model templates by extending the `TemplateOptions` class in a beet plugin. The subclass must define its name as a class variable, and a `process` method which creates the models, mounts them to the pack, and returns them in a list.
<details>
<summary>expand to see example</summary>

```py
 class OreDisplayTemplate(TemplateOptions):
    name = "smelteries:ore_display"
    texture_map = ["top", "side"]
    
    default_transforms = [
        ItemDisplayModel(
            origin=[0.5, 0.5, 0.5],
            scale=[0.624, 0.624, 0.624],
            translation=[0, 0.19, 0]
        )
    ]

    def process(self, config: ModelData, models_container: NamespaceProxy[Model]):
        model_name = ensure_single_model_config(self.name, config)
        reference = config.reference.split('/')[-1]
        m = models_container[model_name] = Model({
            "parent": "gm4_smelteries:block/ore_display",
            "textures":{
                "top": f"gm4_smelteries:block/ore_display/{reference}_top",
                "side": f"gm4_smelteries:block/ore_display/{reference}_side",
            }
        })
        return [m]
```
</details>
At the time of writing, the following module-specific templates exist and may be imported and used.

* `smelteries:ore_display`: the ore-block shown inside a smelteries cauldron
* `tnt_landmines:landmine`: a landmine in the ground
* `trapped_signs:sign_overlay`: the trapped sign texture, overlaid on the vanilla item
* `pneuma`: generates the shard and essence models for pneumas/souls
* `shamir`: generates models for a shamir band, and tool/armor models with the shamir applied. Has the following sub-config.
<details>
<summary>expand</summary>

* `textures_path`: directory of textures to use when the shamir is on a tool. Textures from metallurgy will be substituted for any missing textures. Unless the shamir has unique textures, its recommended to just use the metallurgy folder corresponding to the metal type.
* `metal`: which metal this shamir is

This template also overrides the `item` field of the config to allow more broad item groups, such as `armor`, `tools`, `weapons`, `axes`, `helmets` ect...

e.g.
```yaml
- item: swords
    reference: shamir/corripio
    template:
      name: shamir
      metal: bismuth
      textures_path: gm4_metallurgy:item/shamir/bismuth
```
</details>

</details>

### Model Transformations
<details>
<summary>expand</summary>

The `transforms` field of `model_data` entries is used to apply transformations (rotation/translation/scale) to a template-generated (or custom provided) model. Right now there is only one available transform: `item_display`. This transform does the convenience calculations for an item model that will be displayed by an item display entity. By first setting up the display entity without the resource pack, and providing the entites display parameters, the specified model will be modified to a 1-block scale, aligned with the block grid. 
Accepts the required parameters `origin`, `scale`, and the optional parameters `translation=[0,0,0]`, `rotation=[0,0,0]`, `display=head`

example
```yaml
- item: glass
  template: custom
  reference: block/liquid_tank
  transforms:
    - name: item_display
      origin: [0.5,0.9,0.5]
      scale: [0.438,0.438,0.438]
      translation: [0,0,0]
      display: head
```
</details>

### Font-powered Gui Textures
<details>
<summary>expand</summary>

Custom textured GUIs using fonts can easily be setup using the `meta.gm4.gui_fonts` config entry. These will create a translation that displays a given image texture inside a container, like a dropper or hopper. Options are:
* `translation`: the translation key that will become the texture. Recommended to follow the format `gui.gm4.my_container`.
* `texture`: the image texture to display
* `container`: the type of container the gui will display inside. Currently accepts `dropper`, and `hopper`. Additional containers formats can be added by extending the `ContainerGuiOptions` class in a plugin
example.
```yaml
gui_fonts:
  - translation: gui.gm4.auto_crafter
    container: dropper
    texture: gm4:gui/container/empty_dropper
```
</details>

### Babelbox Translations.csv
Language translations are now all defined in a `translations.csv` file, either located in the module root or in the `assets` directory. This will allow for translations to easily be added for other languages in a spreadsheet format.

All English (en_us) translations used in a given module should be present in this spreadsheet, and mecha will issue linting warnings for any unused or undefined translation keys within the project.

## Additional Changes
* Upgrades beet to `0.100.0` and mecha to `0.86.2`, to use mecha compilation within json files. 
* Restructures the `beet-release` pipeline using beet workers to partially build each module, check its output for changes, and then increment its version number if needed.
* Tweaks some block models (smelteries, forming press ect) so the model file is a 1-to-1 scale.
* Updates block_compressors, tinkering_compressors, and ender_hoppers to use display entities instead of armor stands


gm4_metallurgy/beet.yaml
```diff
@@ -63,3 +67,108 @@ meta:
...
+      - item: player_head
+        reference: gui/advancement/metallurgy_blast_ore
+        template:
+          name: advancement
+          forward: item/lump/bauxite
+      - item: player_head
+        reference: gui/advancement/metallurgy_cast
+        template:
+          name: advancement
+          forward: shamir/defuse/band
```
Bloo-dev, line 174: I've been looking through this this morning and think we should re-think where we put these texture/asset definitions.

`beet.yaml` has so far been for meta-data, whereas this asset data isn't meta-data anymore if you think about it, it's resource definitions. That plus the fact that these make the `beet.yaml` extremely long and overwhelming to look at when a module introduces more than a couple of textures makes me think these should go in their own, separate `.yaml` file.

Bloo-dev (changes requested):

gm4_metallurgy/beet.yaml
misode: Agreed. I think similar to how we have `translations.csv`, maybe `assets.yaml` or `assets/items.yaml` could work.

SpecialBuilder32: So a couple of arguments for why I think it should stay in `beet.yaml`.
1. `beet.yaml` isn't for "module metadata", but rather "beet build metadata", with all the info required for beet plugins to work. 
2. Putting this data in `beet.yaml` lets its structure be easily read in and validated by beet. Malformed data structures get nice diagnostic data printed upon the beet build. Counter - I could reimplement this validation on my own.
3. The data here lets it be generated by a beet plugin easily, just by appending to the `ctx.meta['gm4.model_data']`. Most of the most "overwhelming" config is created this way, since those are also the most repetitive. Counter - I could create some file pinned to `ctx.data.extra` somewhere, but it feels strange to put it either on the datapack or texturepack, and it'd have to be cleaned up before we output.
4. Model data isn't just for defining items, but any CMD interactive assets, such as advancement icons or block models. About half our modules don't have any assets of their own, but do still define CMD for future compatibility or retexturing packs.

EpyonProjects: re: end of 3, how would moving this model_data into a standalone file change whether it is put on either the datapack or resource pack, if the format stays the same? (since we just want to move this list into a separate file)

SpecialBuilder32: Beet requires all files to be members of either the ResourcePack() or DataPack(), so any standalone file would need to belong to one of these to be accessed by plugins. `meta` is just a general purpose dict present on the `Context`, so its nicer to access by plugins.  

Its also just organizationally weird to have an asset that affects both the DP and RP only be a member of one of those containers, but I can get over that

SpecialBuilder32: An update on this thread from today's meeting: I'm going to add a config file that gets merged into `meta`, that way modules with very long model_data can move it out into a separate file to not clutter the beet.yaml, but modules with one or two definitions can still keep it in the beet.yaml directly.

gm4/skin_cache.json
```diff
@@ -1001,11 +990,11 @@
...
     "gm4_zauber_liquids": [
+      "gm4_potion_liquids:liquids/healing",
+      "gm4_potion_liquids:liquids/regeneration",
       "gm4_potion_liquids:liquids/poison",
       "gm4_potion_liquids:liquids/harming",
       "gm4_potion_liquids:liquids/strength",
-      "gm4_potion_liquids:liquids/regeneration",
-      "gm4_potion_liquids:liquids/healing",
       "gm4_potion_liquids:liquids/leaping",
       "gm4_potion_liquids:liquids/swiftness"
```
misode, line 999: Would it be difficult to sort this list?

resource_pack/beet.yaml
```diff
@@ -0,0 +1,31 @@
...
+name: Resource Pack
+version: 1.1.X
+
+resource_pack:
+  load:
+    pack.png: pack.png
+    # other files are inherited by the full-project build
+
+pipeline:
+  - dev_warning
```
misode, line 11: Should this not be `resource_pack.dev_warning`, like other module plugins?

beet-dev.yaml
```diff
@@ -4,7 +4,52 @@ ignore:
...
+  - gm4.plugins.finished
 
 meta:
   autosave:
     link: false
+  gm4_dev: True
+
+ignore: 
+ - gm4/skin_cache.json
+ - gm4/modeldata_registry.json
```
misode, line 55: Remove the `ignore` section at the top of this file

pyproject.toml
```diff
@@ -6,20 +6,24 @@ authors = ["Gamemode 4 Contributors"]
...
 
 [tool.poetry.dev-dependencies]
 black = "^22.6.0"
 isort = "^5.10.1"
 
 [tool.poetry.plugins.beet]
 commands = "gm4.commands"
+autoload = "gm4.plugins.autoload"
+
+[tool.beet]
```
misode, line 26: Is this correctly structured? Is `[tool.beet]` supposed to be empty?

misode (changes requested):

base/assets/gm4/lang/en_us.json
```diff
@@ -0,0 +1,3 @@
+{
+    "gm4.second": "%2$s"
```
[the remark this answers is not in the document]
SpecialBuilder32: This won't have separate values for each language, as the game always falls back to `en_us`.

gm4/plugins/autoload.py
```diff
@@ -0,0 +1,13 @@
+from beet import Context
+
+def beet_default(ctx: Context):
+    """Loads plugins auto-added to every beet subproject. Configured in pyproject.toml
+    Used primarily to load custom-merge-rules that will apply between broadcast packs, at least until a better option is available in beet"""
```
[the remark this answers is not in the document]
SpecialBuilder32: It's not needed in every subproject (eg libraries). Plus this implementation is a bit of a hack to deal with a beet limitation where broadcast makes an invisible subproject level you can't add merge rules too. 

The RP plugin only registers one merge rule (for optifine files) that I think isn't needed on the invisible subproject.

gm4/plugins/test.py
```diff
@@ -16,3 +17,9 @@ class TestFile(TextFile):
 def strip_tests(ctx: Context):
     for structure in ctx.data.structures.match("*:test_*"):
         del ctx.data.structures[structure]
+
+
+def skip_mecha_lint(ctx: Context):
```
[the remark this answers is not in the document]
SpecialBuilder32: I don't know how much time it saves, but when translations go wrong it's nice to not have the same error show up in both build and test.

resource_pack/beet.yaml
SpecialBuilder32: Within the scope of this subproject (the `resource_pack` folder), those two are identical. Beet uses python imports for this so it searches the local scope before escalating upward

gm4/skin_cache.json
SpecialBuilder32: It should be pretty easy. I just didn't want to do it in this PR. Ill add it though since it's tiny

gm4/plugins/test.py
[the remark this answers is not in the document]
misode: I see. I wrongly assumed that a lint error would halt the workflow but if this removes the duplicate error annotations that's good.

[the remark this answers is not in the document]
SpecialBuilder32: Most lint errors are logged at the `warning` level and so won't halt the build.

resource_pack/beet.yaml
misode: Is this just for preventing people running `beet build` in the `resource_pack` folder? We don't have that for any other module and I don't know of anyone that has gotten confused by it. It should just be clearly explained in the docs how to build the resource pack.

SpecialBuilder32: This is to prevent people from `beet dev resource_pack` which is otherwise valid and produces an empty pack. It's structured like a module and it's reasonable to think it would work like one.

gm4/plugins/output.py
```diff
@@ -42,6 +42,29 @@ def beet_default(ctx: Context):
...
+	version = os.getenv("VERSION", "1.20")
+	out_dir = Path("out")
+
+	ctx.assets.save(
+		path=out_dir / f"gm4_resource_pack_{version.replace('.', '_')}",
+		overwrite=True
+	)
+
+def release_resource_pack(ctx: Context):
+	"""Saves the resourcepack to the ./out folder."""
```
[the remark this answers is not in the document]
SpecialBuilder32: Yes, though how we package the RP on those sites is still undecided so I left it for future work. (Ie do we add it to every smithed project as the paired RP?). We need to get our own release working as a first priority 

I would like to abstract the release functions here to work with either the RP or the DP when we get to that.

Bloo-dev (approved):
I approve of the optional model data file.

pyproject.toml
SpecialBuilder32: Yes this is required to be a stub for.. reasons I don't understand. See https://discord.com/channels/900530660677156924/900533796179419226/1009269844501352468

base/assets/minecraft/atlases/blocks.json
```diff
@@ -0,0 +1,14 @@
...
+    "sources": [
+        {
+            "type": "directory",
+            "source": "gui",
+            "prefix": "gui/"
+        },
+        {
+            "type": "directory",
+            "source": "entity",
+            "prefix": "entity/"
```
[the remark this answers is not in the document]
SpecialBuilder32: Can you elaborate on what you mean by that? Don't these configure additional atlases to create?

base/assets/gm4/textures/gui/container/empty_dropper.png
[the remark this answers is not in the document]
SpecialBuilder32: `empty_hopper` and `empty_dropper` are here to serve as a template file for adding GUI textures to a hopper or dropper. They also can serve as the global placeholder for a machine without any gui texture (like autocrafters) since they are the right size for the font definition.

base/assets/minecraft/atlases/blocks.json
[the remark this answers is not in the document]
misode: The `blocks` atlas is a single texture atlas that is use by block and item model rendering (ignoring special entity renderers). So adding more texture to this makes the atlas larger. If the textures aren't used anywhere in item or block models they're just wasted space.

resource_pack/beet.yaml
misode: Oh interesting, I did indeed expect that to work. How do you currently build the resource pack? Is `beet-release` the only way?

base/assets/gm4/lang/en_us.json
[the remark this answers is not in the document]
SpecialBuilder32: The wiki no longer seems to document this behavior, but I just verified ingame that `gm4.second` does properly apply the `en_us` translation even when in a language without a more specific value.

This is used in beeps font magic gui retexturing and guidebook. To the best of my knowledge, its used to switch from a simple non-rp text component to the complex rp-dependant text component that uses fonts and other stuff to render properly. I think without this switching, the client would get errors about not finding fonts its being told to render?

base/assets/minecraft/atlases/blocks.json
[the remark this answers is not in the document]
SpecialBuilder32: I believe since all our textures are rendered by item model (on a display entity or actual item) they would need to be in this atlas then?

The `gui` directory is intended for things that show up in menus, so has stuff for guidebook and fonts. Do they need to get their own atlas, or would they not need an atlas at all?

resource_pack/beet.yaml
SpecialBuilder32: Yes, `beet-release` is the only way to get a full build of the RP. This is because it needs to go through the normal beet build process for each pack to load and process their assets. The RP is then assembled by the resource pack resulting from beet merging every module's individual resource pack.

In normal development, devs should only need to interact with the assets associated with their current module(s). `beet dev` is setup to output and link a smaller resource pack produced by just the modules listed. For example, `beet dev apple_trees zauber_cauldrons` will produce a single `gm4_resource_pack` in the out directory that contains both apple trees and zauber resource pack assets.

misode:
https://github.com/Gamemode4Dev/GM4_Resources/commit/43811f3fd52c910c2c4df590baf548fdc80cdf3b still needs to be incorporated into this

misode:
Was https://github.com/Gamemode4Dev/GM4_Resources/issues/66 addressed in this redo?

misode:
I made a diff of the release output comparing `master` and this PR. It looks fine at first glance, some CMD moves which I assume are intentional: https://github.com/misode/gm4-rp-merge-compare/commit/377c322ab690cbfa9a584c51d1e36082a5473e79

misode (approved):

PR #920, opened 2023-11-21
Double Doors: Close doors automatically after a player walks through

Bloo-dev:
## Progress
- [x] Introduce Auto Closing
- [x] Enable Auto Closing for All Wood Types
- [x] Add Module Icon

### The Problem
Double Doors is mainly a QoL module, and therefore should not interact with Redstone (nor can we easily implement Redstone interactivity). Still players enjoy the comfort of using pressure plates to automatically close their doors behind them.

Together with Double Doors, however, this causes an issue which can be reproduced as follows:
1. Place a Double Door with two pressure plates on one side
2. Open the Double Door by hand and walk through it
3. Observe how walking through it but only activating one of the pressure plates (e.g. because you tend to walk on the left side of the door due to your path layout) only closes one of the two doors, the other one remains opened

It can be argued that the player should link the pressure plates to both doors using Redstone but as Double Doors is mainly a QoL module with focus on making nice builds functional (e.g. through functional trapdoors above doors) this seems like an odd argument to make. Often builds simply lack the space to add a Redstone wire connecting both doors underneath the floor. Moreover, this would not close trapdoors above the door.

### The Solution
This PR introduces an auto-closing feature to Double Doors. When a player opens a door a marker is spawned which lingers around for a few seconds. If a player walks through that marker within that time, the associated door will be closed (i.e. returned to the state it was in before the player clicked it) automatically behind the player. As with the current Double Doors, sneaking prevents this behavior.

This auto-closing functionality does not make Redstone on doors useless, nor does it give the player any meaningful advantage; It's quite rare to see a player actually wire up doors to two pressure plates using Redstone wires, mostly players rely on walking over both pressure plates by pure chance.


gm4_double_doors/data/gm4_double_doors/functions/auto_toggle_marker/select_material.mcfunction
Bloo-dev, line 1: The main point of concern with this PR is this file. Checking for the material type and making sure the door sound only plays once requires a lengthy list of score checks.

This could theoretically be replaced with a function tree, however, this function isn't called very often (only called once when a door is actually toggled by an auto toggle marker) so I consider these checks to be ok to keep like this.

Bloo-dev:
This PR also adds an icon to Double Doors.

(just making this comment to test a webhook)

Bloo-dev:
Another test comment

SpecialBuilder32 (approved):
Took a look over the beet processing - everything looks good on that front.

PR #924, opened 2023-12-05
Fix AEC duration in overlay_18

misode:
Some area effect cloud `Duration` NBT keys were incorrectly made lowercase in #908


misode:
I'm going to merge this since it's a pretty important hotfix

PR #929, opened 2023-12-21
Fix Custom Cart Placement

BluePsychoRanger:
1.20.3 changed the string format once again. This fixes placement of our custom carts.

Auto Crafters are still broken, but #928 fixes that separately.

Also changes the name of newly placed tunnel bores to include a translate tag since the names are visible now.


SpecialBuilder32 (approved):

BluePsychoRanger:
this PR now bumps the beet version to 0.97.1 which allows for overlay specification inside the `beet.yaml` instead of requiring a `pack.mcmeta`. Once this is merged, the rest of the repo should be updated to use the new overlay settings.

misode:
This seems like a pretty important fix. Merging since special approved

PR #935, opened 2024-01-05
Add undead players and animi shamir compatibility

misode:
This adds the `gm4_player_death` library, which detects player deaths and then calls 4 main function hooks, on order:
* `#gm4_player_death:soul_bound`: catch items that should return to the player
* `#gm4_player_death:relocate`: catch items that should end up somewhere else
* `#gm4_player_death:grave`: summon a grave or catch the remaining items
* `#gm4_player_death:death`: anything else

These phases are up for discussion, and more can be added in the future without breaking compatibility.


gm4_undead_players/data/gm4_undead_players/functions/died.mcfunction
```diff
@@ -1,4 +1,4 @@
-#run from gm4_undead_players:death advancement
+#run from #gm4_player_death:grave
 #@s = player who took fatal damage
 
 # Reset death tracking scoreboard and advancement to allow for the player to die again.
```
BluePsychoRanger, line 4: the old scoreboard clear and advancement revoke should be removed

lib_player_death/data/gm4_player_death/tags/functions/_private/death.json
BluePsychoRanger, line 1: Is this supposed to be versioned? I think this is okay unversioned if we plan to never remove any of the values or rearrange the current values (which is a reasonable assumption to make; this would basically require us to never make a breaking change, thus major version will always be 1). Without versioning we can only insert new values to this function tag (tho depending on how MC merges function tags, that may also cause issues with the order of future insertions).

BluePsychoRanger (changes requested):
Looks good, but there are some leftover advancement revokes and scoreboard objective resets. I also left a comment about versioning the function tag this library calls.

Additionally, ZC has a death check, but doesn't use an advancement to detect it, instead relying on a schedule loop (player/submain) to check the score and run the event. It may be worth it to make ZC use this library.

misode:
I did notice the ZC check, but decided to leave it because it would probably conflict with #885

As for versioning the private tag, I don't think that's necessary since we wouldn't ever want to remove phases. Unless you see a good use case for it, and think it should include the major version?

Bloo-dev:
Zauber Cauldrons will no longer require any death checks #885 has been merged, so there's nothing to re-write there.

BluePsychoRanger (approved):
Looks good. I don't think versioning the function tag is necessary

lib_player_death/data/gm4_player_death/tags/functions/_private/death.json
Bloo-dev: Maybe one slight bit of pedantry that should at least be discussed:
Do we want to name these function tags after _explicit_ uses or _generic_ descriptions? E.g. should it be `#gm4_player_death:soul_bound` or `#gm4_player_death:item_manipulating`?

misode: I do think it's good to be explicit about these. All 4 phases can do item manipulation (soul bound, void fishing, graves). The point of these different phases is to make the priority predictable. I see `soul_bound` as being filtering on item properties, `relocate` on filtering based on the environment, `grave` as a catch all for item filtering, and `death` as anything after all items have been processed. If we have good generic names for those 4 phases I think I would prefer that but these explicit examples are fine for me as well.

Bloo-dev (approved):

PR #937, opened 2024-01-09
Allow lib_player_heads to Create Unstackable Player Heads

Bloo-dev:
Allows `lib_player_heads` to source restore data from loot tables directly instead of fixed a fixed nbt compound in the heads registry. This change removes the need to specify the item nbt twice (once in the heads registry and once in a loot table/function/etc which creates the item within a module) and allows for `lib_player_heads` to create unstackable skulls via loot tables, as this was previously impossible due to the static nature of the raw nbt compound in the heads registry.

~~Allows data packs to register their player heads as `unstackable` with `lib_player_heads`. This forces the library to insert a random number in the item's NBT at [`Item.tag.`]`gm4_player_heads.uuid`.
To mark a registered player head as unstackable, it must have `unstackable:1b` in the heads registry (so e.g. `{name:foo:bar/v0,texture:"foobar",unstackable:1b,item:{<item tag nbt>}}`~~

### ...but why?
Some module may want to make their player heads unstackable. However, the player heads library will always restore placed skulls to have fixed data specified within the registery. As a result, having consistently unstackable player heads would've required another clock previously.
This PR is a requirement for #885.

### Versioning
This change isn't breaking, yet using an old version would cause items to stack when they shouldn't. Hence the minor version was bumped up to `v1.2`.


lib_player_heads/data/gm4_player_heads/functions/apply_nbt_from_register.mcfunction
```diff
@@ -0,0 +1,14 @@
+# Apply the stored nbt data from the registry to the player head item on the ground
+# @s = player head with a match in the registry
+# at world spawn
+# run from repair_with_name and repair_with_texture
+
+# prepare register data
+data modify storage gm4_player_heads:register unstackable set value 0b
+execute store success score $enforce_unstackable gm4_player_heads run data modify storage gm4_player_heads:register unstackable set from storage gm4_player_heads:register temp[-1].unstackable
```
BluePsychoRanger, line 8: This should do the same thing with less nbt changes and lookups
```suggestion
scoreboard players set $enforce_unstackable gm4_player_heads 0
execute store result score $enforce_unstackable gm4_player_heads run data get storage gm4_player_heads:register temp[-1].unstackable
```

BluePsychoRanger (commented):
This looks fine. Initial player heads will still need to be unstackable (via the loot table), but that's easy enough with an attribute modifier in the loot table/item modifier (see mountaineering ski pole); we'll just need to make it clear in the documentation (that should be updated in this PR) that this only sets unstackable status during repairing since that's what this library is for.

misode:
Is there a specific module change you have in mind that requires this?

Bloo-dev:
Yes, I'm currenlty working on making Zauber Cauldrons' Crystals unstackable, but I didn't want to burry a library change in my large Zauber PR.

Bloo-dev:
After some consideration I've come to the conclusion that the previous implementation of simply adding a hook-in for unstackable items wasn't very elegant.
This new implementation allows for the use of loot tables in the heads register.

I have not yet tested this in-game.

Bloo-dev:
This has now been tested in-game.

PR #942, opened 2024-01-21
Zauber Liquids & Mob Curing: Enable Compatibility with Zauber Cauldrons v1.10

Bloo-dev:
Zauber Cauldrons v1.10 changed the name of some loot tables which these modules relied upon. This PR fixes the loot table references.

As Zauber Cauldrons v1.10 is only available for Minecraft 1.20.3+, this means Zauber Liquids will also only be available for those Minecraft versions.

### DIFF
--- a/gm4_mob_curing/beet.yaml
+++ b/gm4_mob_curing/beet.yaml
@@ -13,7 +13,7 @@ meta:
   gm4:
     versioning:
       required:
-        lib_trades: 1.3.0
+        lib_trades: 1.4.0
       schedule_loops: [main]
     website:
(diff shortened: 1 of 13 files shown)


gm4_mob_curing/beet.yaml
```diff
@@ -13,7 +13,7 @@ meta:
   gm4:
     versioning:
       required:
-        lib_trades: 1.3.0
+        lib_trades: 1.4.0
```
misode, line 16: This is a problem. The `lib_trades` version is still 1.3.X and the following warning also shows in the log
image

PR #944, opened 2024-01-22
Move CSV utils to gm4.utils

Bloo-dev:
Moves the CSV interface first introduced in Zauber Cauldrons utils into `gm4.utils`, as PR #896 also makes use of this functionality.


Bloo-dev:
Tested this in-game and Zauber Cauldrons works the same as it did before.

misode (approved):

PR #945, opened 2024-01-23
Your Sparks is in another castle!

sparks0:
My Twitter handle changed a while ago and I just noticed the old URL is still used by the site.

I did a search of the whole repo and I think the only place the link is referenced is contributors.json so hopefully it isn't listed somewhere else as well!

### DIFF
--- a/contributors.json
+++ b/contributors.json
@@ -97,7 +97,7 @@
     },
     {
         "name": "Sparks",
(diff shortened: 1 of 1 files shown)


Bloo-dev:
Correct, `contributors.json` is our central credits file from which the website and all modules pull their credits.

PR #951, opened 2024-01-29
Hotfix: Make destroying custom carts drop the correct item

BluePsychoRanger:
- fixes liquid minecarts and tunnel bores
- ender hopper minecarts seemed to not be affected


Bloo-dev:
We've been running this on the server for a week now and have not encountered any issues with it.

PR #952, opened 2024-01-30
Bookshelf Inspector: Optimise & Simplify Everything

TheThanathor:
This PR aims to optimize Bookshelf Inspectors by drastically cutting back on checks that need to be performed. This makes the module slightly more janky to use but with more than worthwhile performance gains (I hope)


misode (approved):
I can confirm the module is working fine. I am not bothered by the 8t clock, it still feels very natural.

However I am getting the following error in the log every time I look at a book. Checking the game code it's in the `Display` class, somewhere parsing some NBT data, but I haven't narrowed down the cause.
image

TheThanathor:
> I can confirm the module is working fine. I am not bothered by the 8t clock, it still feels very natural.
> 
> However I am getting the following error in the log every time I look at a book. Checking the game code it's in the `Display` class, somewhere parsing some NBT data, but I haven't narrowed down the cause. image

This is caused by the text_display not being summoned with `alignment` specified (since it is summoned through an execute summon), it can be fixed by spawning the entity first and then finding it again to run the rest of the init if that's preferred (would add an additional `@e` but it only runs when a new text display is spawned so probably fine

gm4_bookshelf_inspector/data/gm4_bookshelf_inspector/functions/evaluate/run.mcfunction
```diff
@@ -1,24 +1,13 @@
...
+# check player motion and if they are looking at a chiseled bookshelf
 # @s = player not in spectator mode
 # at unspecified
 # run from main
 
-scoreboard players set $evaluate gm4_bookshelf_inspector_data 1
+# do not check for chiseled bookshelves if the player moved anyway, allow some leniency for walking to avoid nudging removing the display 
+execute if score @s gm4_bookshelf_inspector_walk matches 100.. run return 0
+execute if score @s gm4_bookshelf_inspector_sprint matches 1.. run return 0
+execute if score @s gm4_bookshelf_inspector_fall matches 1.. run return 0
```
misode, line 9: Instead of doing these checks here, I think it'll be more optimal if they're all chained together in `main` on line 7

misode (changes requested):
The module in this PR only seems to work for chiseled bookshelfs facing west. I would suspect this is not the case for the released version.

TheThanathor:
> The module in this PR only seems to work for chiseled bookshelfs facing west. I would suspect this is not the case for the released version.

this should now be fixed, was not working due to a score not being set correctly anymore with the new executing entity

misode (approved):
Tested and seems to be fixed 👍

PR #953, opened 2024-02-01
Add Trident to Hypexperia

venomousbirds:
Hypexperia wasn't available to tridents and I didn't see why not.


TheThanathor:
I don't see why not to add tridents, especially since fishing rods are already valid.

There is the issue where since tridents are mostly a thrown weapon it may not be in the inventory when the xp drops, and therefor it will not be doubled

misode (approved):

PR #956, opened 2024-02-06
Zauber Cauldrons: Only Allow Magicol Bottling at Night

Bloo-dev:
This updates an oversight in the magicol bottling code to make it more closly align with the lore by only allowing bottling when the moon is actually visible.

Not tested in game yet.

### NOTE
This requires mecha for 1.20.3, aka mecha `v0.85.2` or above. I did not add that here as PR #918 updates mecha.


Bloo-dev:
I am not quite sure why the build fails after the update to the dependecies. Any help would be appreciated, I did not touch the player head plugin in this PR.

Bloo-dev:
This PR has been updated to rely on #961.

PR #958, opened 2024-02-07
Infinitas Shamir: Waterlogging blocks no longer breaks the shamir

EpyonProjects:
- Previously only targeted cauldrons, now uses gm4:waterloggable block tag
- #rails and Mangrove Propagule missing from base block tag
- Also specifically flower pots break the check for some reason

### DIFF
--- a/gm4_metallurgy/data/gm4_infinitas_shamir/advancements/water/place_mainhand.json
+++ b/gm4_metallurgy/data/gm4_infinitas_shamir/advancements/water/place_mainhand.json
@@ -314,8 +314,53 @@
             "predicate": {
               "block": {
                 "blocks": [
+                  "minecraft:flower_pot",
                   "minecraft:water"
                 ]
+              }
+            }
+          }
+        ]
+      }
+    },
+    "place_waterloggable": {
+      "trigger": "minecraft:placed_block",
+      "conditions": {
+        "player": [
+          {
+            "condition": "minecraft:inverted",
+            "term": {
+              "condition": "minecraft:entity_properties",
+              "entity": "this",
+              "predicate": {
+                "type_specific": {
+                  "type": "player",
+                  "gamemode": "creative"
+                }
+              }
+            }
+          },
+          {
+            "condition": "minecraft:entity_properties",
+            "entity": "this",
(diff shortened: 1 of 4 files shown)


gm4_metallurgy/data/gm4_infinitas_shamir/advancements/water/place_mainhand.json
```diff
@@ -314,6 +314,7 @@
             "predicate": {
               "block": {
                 "blocks": [
+                  "minecraft:flower_pot",
                   "minecraft:water"
```
misode, line 318: Why is flower pot special? Should this not include all blocks that break when you place water inside?

gm4_metallurgy/data/gm4_metallurgy/predicates/infinitas_active.json
```diff
@@ -5,14 +5,20 @@
       "condition": "minecraft:entity_properties",
       "entity": "this",
       "predicate": {
-        "nbt": "{Tags:[\"gm4_infinitas_mainhand_empty\"]}"
+        "nbt": "{Tags:[\"gm4_infinitas_mainhand_empty\"]}",
+        "flags": {
+          "is_sneaking": true
+        }
```
misode, line 11: I don't fully understand this part of the module, but why was this change necessary?

EpyonProjects: _This part was actually unrelated and a server test.  It probably should have been put in a different PR :grimacing:_

EpyonProjects: The reason it was added: Previously, Spiraculum was not usable as a tool, so you wouldn't be holding a Spiraculum bucket unless you were specifically looking for the vacuum effect.  Now, with empty Infinitas Buckets having a purpose, it becomes very frustrating when you frequently vacuum nearby entities.  Sneaking to trigger the vacuum effect seemed like a solid way to activate it (and it felt pretty good to use when I tested it on the server)

misode (approved):

PR #960, opened 2024-02-10
Double Doors: Add a setting to disable auto door open/closing after walking through

suppergerrie2:
This adds a trigger the player can call to toggle whether doors they go through will automatically open/close after walking through. 
This is for players who want to use the module but do not want their doors to open randomly after closing it.
By default no behaviour is changed, only after the player runs the trigger behaviour changes for that player only. 

I'm sure there are better ways to add a player-specific config option like this, if there is please let me know and I can update the PR.


gm4_double_doors/data/gm4_double_doors/functions/main.mcfunction
```diff
@@ -0,0 +1,4 @@
+execute as @a[scores={gm4_double_doors_auto_toggle_settings_trigger=1..}] run function gm4_double_doors:toggle_auto_setting
+scoreboard players enable @a gm4_double_doors_auto_toggle_settings_trigger
+
+schedule function gm4_double_doors:main 20t
```
Bloo-dev, line 4: Is any of this really necessary?
We could simply check the triggered score itself (`gm4_double_doors_auto_toggle_settings_trigger` side note: a shorter name might be nice) in the `.../toggle.mcfunction` functions and allow any non-zero positive score to inhibit the auto-closing feature (`unless score @s gm4_double_doors_auto_toggle_settings_trigger matches 1..`).

That way almost no code in `main` is necessary. Obviously, we still have to enable the trigger for everyone (and keep enabling it once they have used it... annoyingly), but maybe we can find a way of doing that without a `main` even? (to be researched)

misode:
I'm not a fan of adding triggers like this. If this is necessary, there clearly is something wrong with the module that should be fixed, rather than allowing some players to circumvent part of the module.

Bloo-dev:
> I'm not a fan of adding triggers like this. If this is necessary, there clearly is something wrong with the module that should be fixed, rather than allowing some players to circumvent part of the module.

The situation here is a muscle memory issue: Some people close doors behind them by muscle memory. This module makes that unnecessary, However, some people still click the door out of muscle memory, opening it when the module has already closed it.

Interestingly, @suppergerrie2 mentions here that the module randomly opens doors, which is not the case -- it follows a very strict and simple set of rules, yet apparently it seems random to some players. I thought about this a lot and tried to come up with ways we could tweak those rules to make it appear less 'random' to players, but all of those thoughts just let to an overcomplicated set of rules instead of the simple ones we have right now.

EpyonProjects:
> I'm not a fan of adding triggers like this. If this is necessary, there clearly is something wrong with the module that should be fixed, rather than allowing some players to circumvent part of the module.

As a utility module, I think it's especially important to give the player the choice to opt-in/out.  Similar to "accessibility" settings in Options, not everyone wants all of them.  Like Auto-Jump, which used to be enabled by default.  They've seen turned it off by default, but the option is still available for players if they want it :shrug:

Maybe a trigger isn't the most obvious choice, and you could allow the players to change it by the Guidebook? (suggested by gerrie, but I'm writing it first)

suppergerrie2:
With the changes to the module this seems to not be a huge issue anymore, I still think a utility module like this could benefit from settings like this but that would require some more thinking about how to implement this nicely without being able to add a config menu like mods can.

PR #961, opened 2024-02-10
Update to mecha 0.87.0 and pydantic v2

misode:
* Tested that the output matches exactly with the current release.
  * The only changes were nbt files and a bugfix in `upgrade_paths/run.json`. It seems like it was adding the function entry twice before.
* Updating to mecha `0.87.0` instead of `0.88.8` because the latter changed the effect amplifier range to 0-127.
* There are some `# type: ignore` in `prefabs.json` because it did not detect that nbtlib comes with beet. There may be a cleaner solution.
* `player_heads.py` required some changes. It seems like beet now needs the db key to be a `File`, so I just pass a `TextFile` with `source_path` set to the json file and content to the same unique string. I haven't tested that diagnostic messages still work.


EpyonProjects:
Using this PR, I was able to build successfully with my files that previously had issues :+1:

Bloo-dev (approved):
Code changes do make sense regarding the previous typing errors in builds.

I tested this using my changes from PR #956 and it builds and works fine.

SpecialBuilder32 (approved):
Changes look fine to me. Breaking diagnostic message is acceptable as #918 will restore them hopefully in the next few weeks.

As for nbtlib typing, the library is "missing a stub file" and so apparently just doesn't play nice with type linters atm. Theres some background work to get that updated by someone in beet. 

I don't see the supposed bugfix in upgrade paths other than the import migration, so I'm not sure whats up with that.

misode:
> I don't see the supposed bugfix in upgrade paths other than the import migration, so I'm not sure whats up with that.

If you download a module like block compressors right now, you can see that the file `data/gm4_upgrade_paths/tags/functions/run.json` currently looks like this:
```json
{
  "values": [
    "gm4_custom_crafters-3.1:upgrade_paths/run",
    "gm4_custom_crafters-3.1:upgrade_paths/run"
  ]
}
```

I'm assuming the JSON support for mecha fixed this, although it might still be useful to look into why this was happening in the first place.

PR #963, opened 2024-02-13
Double Doors: Disable for Powered Doors & Improve Raycast

Bloo-dev:
This is an update to Double Doors which aims to polish some of the rough edges of the initial release, namely:
- Improved Raycasts: Previously when opening double doors from a off-axis POV, the doors would -- especially when opening from the back-side (the side on which the door is not on the edge of the block) -- sometimes flicker and not open as inteded. This PR fixes this issue by introducing hinge-based parity and treating each case separately. I.e. if you click a left-hinged door, the raycast will ignore right-hinged doors.
- Improved Hitbox Detection: Only auto-toggle a door after a player walked through if the player walked through the region which was previously occupied by the door, not the entire block space (current behavior).
- Allow Manual Overwrites: Players manually toggling a door now removes all scheduled auto-close actions for that door and its neighbors, if the manual intervention by the player achieved the same as the auto-close action would have achieved
- Redstone Interface: When a door is powered by Redstone it is no longer affected by Double Doors. Previously this would cause undefined behavior, where the door was 'pseudo-locked' in the `open` block state but could be toggled via sneaking

Also adds Pack Tests.


Bloo-dev:
After a week of server testing I can conclude that this works fine.

However, this is not a solution to the fact that some players on the server simply don't want doors to automatically close behind them. For those players, a per-player setting would be the only option. Whether this is a classic public server issue or whether we need to introduce such a setting is up to debate.

gm4_double_doors/data/gm4_double_doors/tests/auto_toggle_east_inverted.mcfunction
```diff
@@ -0,0 +1,59 @@
+# @batch east
+# @template gm4_double_doors:tests/valid_door_east_inverted
+# @dummy
```
misode, line 3: Unless tests clash with each other, you usually don't need to manually batch them, when running them the game will still subdivide each batch into groups of 50 if they go over that, so you can probably leave this in the default batch

misode (commented):
I'm only commenting about the tests right now. Those are quite a lot of tests. Part of me wonders whether we need to test all the cases. Maybe a nice feature to add to packtest is some kind of parameterized test, and similarly auto rotated tests.

Bloo-dev:
This module has this many tests as it is rather cumbersome to test each direction and door state by hand. During development I've had one direction or state break on me multiple times; so it's nice to have these tests verify all directions and states.

misode (approved):

EpyonProjects:
*This is not a code review
Tested on server.  With the latest changes, I believe it is in the best condition it could be in functionally.

PR #966, opened 2024-02-18
Speed Paths: Advancement triggered clock

EpyonProjects:
Changes the constantly running clock to one that activates only after a player steps on a dirt path.  This would keep the clock off more often than not.

This only really makes sense as long as "advancements that constantly fail are very optimized" holds true.


gm4_speed_paths/data/gm4_speed_paths/functions/clock/check.mcfunction
```diff
@@ -0,0 +1,10 @@
+# Starts clock if not enabled
+# @s = player, stepping on dirt_path
+# at @s
+# run from advancement, gm4_speed_paths:on_path
+
+execute store result score $running gm4_speed_paths run schedule clear gm4_speed_paths:main
+execute if score $running gm4_speed_paths matches 1 run schedule function gm4_speed_paths:main 8t
```
misode, line 7: Why does this reschedule the clock when it's already scheduled? If multiple players stepped on a path every 4 ticks staggered for example, could they disable the main clock?

gm4_speed_paths/data/gm4_speed_paths/functions/main.mcfunction
```diff
@@ -1,4 +1,4 @@
 execute as @a[tag=!gm4_on_path] at @s if block ~ ~-0.9 ~ minecraft:dirt_path run function gm4_speed_paths:apply_effects/add_speed
```
misode, line 1: Since this is no longer the traditional main clock, add a header comment explaining this and where it is called from.

Additionally, remove it from the schedule loops in beet.yaml

misode (changes requested):

Bloo-dev:
Feedback from server testing: This works as expected.

gm4_speed_paths/data/gm4_speed_paths/functions/clock/check.mcfunction
EpyonProjects: Thought I replied to this.  I was trying to figure out a way to both check that the clock is running and not to have this issue by appending the schedule.  Though given the advancement change that uses the tag to block, I could alternatively check the tag for @a, which is the same trigger as scheduling the main function?

gm4_speed_paths/overlay_26/data/gm4_speed_paths/functions/apply_effects/add_speed.mcfunction
```diff
@@ -0,0 +1,11 @@
+# applies the Speed Paths attribute and gm4_on_path tag
+# @s = player upon stepping on a speed path
+# at @s
+# run from main
+
+# guard return
+execute unless block ~ ~-0.9 ~ minecraft:dirt_path run return fail
+
+tag @s add gm4_on_path
+scoreboard players set @s gm4_speed_paths 2
```
misode, line 10: Why do you need both a tag and a score to store the state?

gm4_speed_paths/overlay_26/data/gm4_speed_paths/functions/apply_effects/path_check.mcfunction
```diff
@@ -0,0 +1,8 @@
+# checks if the player is (still) on a speed path, otherwise break
+# @s = player
+# at @s
+# run from main
+
+execute if entity @s[tag=!gm4_on_path] unless function gm4_speed_paths:apply_effects/add_speed run return 0
```
Bloo-dev, line 6: Having the `if function` make changes to the state of the player is somewhat weird (we should probably restrict ourselves to using `if function` with pure functions only, or close-to).

Instead of the `unless function` use 
```suggestion
execute if entity @s[tag=!gm4_on_path] return run function gm4_speed_paths:apply_effects/add_speed
```
and adjust the returns within the add speed function accordingly.

gm4_speed_paths/overlay_26/data/gm4_speed_paths/functions/apply_effects/add_speed.mcfunction
```diff
@@ -0,0 +1,11 @@
+# applies the Speed Paths attribute and gm4_on_path tag
+# @s = player upon stepping on a speed path
+# at @s
+# run from main
```
Bloo-dev, line 4: Wrong call location.
```suggestion
# run from path_check
```

gm4_speed_paths/overlay_26/data/gm4_speed_paths/functions/apply_effects/keep_check.mcfunction
```diff
@@ -0,0 +1,9 @@
+# checks if player is still on speed path
+# @s = player, with tag gm4_on_path
+# at @s
+# run from function, apply_effects/path_check
```
Bloo-dev, line 4:
```suggestion
# run from function apply_effects/path_check
```

gm4_speed_paths/overlay_26/data/gm4_speed_paths/functions/main.mcfunction
```diff
@@ -0,0 +1,6 @@
+# Clock is started by tick advancement, on_path, only active while necessary
+# run from function, gm4_speed_paths:clock/start, or scheduled by self
+
+execute as @a at @s run function gm4_speed_paths:apply_effects/path_check
```
misode, line 4: If the goal is to optimize this module, then I don't think running a function `as @a` regardless of score or tag is not that smart. Is there no check you can add to this?

Bloo-dev (changes requested):

Bloo-dev (changes requested):
That complete rewrite (in-overlay) is very nice.

Here are two or three things I found.

gm4_speed_paths/overlay_26/data/gm4_speed_paths/functions/main.mcfunction
EpyonProjects: Previously it was 2 @a's, one that checks the tag and one that doesn't.  That effectively covers all of @a, and this reduces it to 1

gm4_speed_paths/overlay_26/data/gm4_speed_paths/functions/apply_effects/add_speed.mcfunction
EpyonProjects: Following up your question in main, is it better to check @a[tag= or @a[scores ?   If tag is not better, then we can use scores exclusively

misode: I'm pretty confindent that the difference is negligible.

gm4_speed_paths/overlay_26/data/gm4_speed_paths/functions/clock/check.mcfunction
```diff
@@ -0,0 +1,6 @@
+# Used to check if the clock is or should be active
+# @s = player or server
+# at @s or 0 0
+# run from function, clock/start and main
+
+return run execute if entity @a[tag=gm4_on_path,limit=1]
```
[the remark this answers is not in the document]
EpyonProjects: I was trying to use if function more, but it's probably best not to force it.  The normal if entity selector is perfectly fine here

gm4_speed_paths/data/gm4_speed_paths/functions/init.mcfunction
```diff
@@ -3,9 +3,6 @@ execute unless score speed_paths gm4_earliest_version < speed_paths gm4_modules
 scoreboard players set speed_paths gm4_modules 1
 
 scoreboard objectives add gm4_speed_paths dummy
-
-schedule function gm4_speed_paths:main 1t
-
-
+scoreboard objectives add gm4_speed_paths_leave minecraft.custom:leave_game
```
misode, line 6: What is this scoreboard used for? Is it reliable? A client forcefully disconnecting for example will not cause this scoreboard to increase.

EpyonProjects: When a player logs out with the path score, which keeps the clock active, there are no remaining players having the score and the clock stops.  But since they still keep the score, if the clock is not running when they rejoin, they would need to wait for someone else.

Using the rejoin, I clear their existing score so they can start up the clock immediately if they logged out off a path

misode:
I think it might be better to close this and revisit this idea in the future, when we're not in an updating phase.

PR #973, opened 2024-02-21
Boots of Ostara: Allow Offhand Item Usage When On Armor Stands

MichaelMiner137:
Allows grass/moss blocks to be used on armor stand offhand. Currently the block conversion order is inconsistent and will change at random. This update allows armor stands to consistently convert stone to grass blocks in the same tick and with 1 armor stand instead of 2.


gm4_boots_of_ostara/data/gm4_boots_of_ostara/functions/armor_stand_convert.mcfunction
```diff
@@ -6,3 +6,7 @@
 execute store success score $grass_check gm4_boo_data if predicate gm4_boots_of_ostara:armor_stand_grass if block ~ ~-0.5 ~ #minecraft:dirt run setblock ~ ~-0.5 ~ grass_block
 execute unless score $grass_check gm4_boo_data matches 1 if predicate gm4_boots_of_ostara:armor_stand_moss if block ~ ~-0.5 ~ #gm4_boots_of_ostara:moss_replaceable run setblock ~ ~-0.5 ~ moss_block
 scoreboard players reset $grass_check gm4_boo_data
+
+execute store success score $grass_check_offhand gm4_boo_data if predicate gm4_boots_of_ostara:armor_stand_grass_offhand if block ~ ~-0.5 ~ #minecraft:dirt run setblock ~ ~-0.5 ~ grass_block
+execute unless score $grass_check_offhand gm4_boo_data matches 1 if predicate gm4_boots_of_ostara:armor_stand_moss_offhand if block ~ ~-0.5 ~ #gm4_boots_of_ostara:moss_replaceable run setblock ~ ~-0.5 ~ moss_block
+scoreboard players reset $grass_check_offhand gm4_boo_data
```
misode, line 12: Why does this need a separate set of predicates and commands? Could the existing predicates not be modified with an `any_of` to allow for both the mainhand and offhand? If the order of moss and grass is swapped it should work. Right now I think it won't work if the armor stand has moss in its mainhand and grass in its offhand.

misode (changes requested):

misode (approved):

PR #976, opened 2024-02-27
Infinitas Shamir: Adds 2 packtests

EpyonProjects:
- The most basic, does it pick up liquids without breaking
- A more complicated check for data breaking between versions for unsmooshing

@misode, is there a way to build off of a previous test?  The second one is just the first one with extra steps

### DIFF
--- a/gm4_metallurgy/data/gm4_infinitas_shamir/tests/check_unsmoosh_after_use.mcfunction
+++ b/gm4_metallurgy/data/gm4_infinitas_shamir/tests/check_unsmoosh_after_use.mcfunction
@@ -13,18 +13,9 @@ setblock ~1 ~4 ~1 redstone_block
 
 await entity @e[type=item,distance=..6,nbt={Item:{id:"minecraft:obsidian"}}]
 
-summon armor_stand ~2.5 ~4 ~1.5 {Tags:["item_swapper"]}
-await delay 1t
-data modify entity @e[type=armor_stand,tag=item_swapper,limit=1] ArmorItems[3] set from entity @e[type=item,distance=..6,limit=1,nbt={Item:{id:"minecraft:bucket",Count:1b,tag:{gm4_metallurgy:{active_shamir:"infinitas"}}}}]
-item replace entity @s weapon.mainhand from entity @e[type=armor_stand,tag=item_swapper,limit=1] armor.head
-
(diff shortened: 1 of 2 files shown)


misode:
> is there a way to build off of a previous test? The second one is just the first one with extra steps

You can either delete the first test, since all those things are also covered by the second test. Or you can keep the duplicate test logic. There is no way to reuse a previous test, but you can manually create the success state of the previous test with NBT or a different structure template.

gm4_metallurgy/data/gm4_infinitas_shamir/tests/check_unsmoosh_after_use.mcfunction
```diff
@@ -0,0 +1,41 @@
...
+dummy @s drop
+tp @e[type=item,distance=..6,sort=nearest,limit=1,nbt={Item:{id:"minecraft:bucket",Count:1b,tag:{gm4_metallurgy:{active_shamir:"infinitas"}}}}] ~1.5 ~2.5 ~1.5
+
+await delay 1s
+
+setblock ~1 ~4 ~1 redstone_block
+
+await entity @e[type=item,distance=..6,nbt={Item:{id:"minecraft:player_head"}}]
+execute unless data entity @e[type=item,limit=1,nbt={Item:{id:"minecraft:player_head"}}] Item.tag.gm4_metallurgy.stored_shamir run fail "Shamir did not unsmoosh properly"
+execute unless data entity @e[type=item,limit=1,nbt={Item:{id:"minecraft:player_head"}}] Item.tag.display.Lore[2] run fail "Shamir did not unsmoosh properly"
```
misode, line 41: Unless you have a more specific failure message it's probably best to use `assert data` here, which gives better messages:

```mcfunction
assert data entity @e[type=item,limit=1,nbt={Item:{id:"minecraft:player_head"}}] Item.tag.gm4_metallurgy.stored_shamir
```

misode: Indeed, `assert` is basically `execute unless <condition> run fail <nice error message>`

EpyonProjects:
I'm not crazy, right??

EpyonProjects:
I've been thinking about the duplicated code more.  I could probably just have an assert in the middle of the longer one to check what the first does

PR #979, opened 2024-03-06
Zauber Cauldrons: Fix Rabbit Teleportation Easter Egg

Bloo-dev:
This easter egg was accidentally broken during some recent update. This PR fixes it and adds some particles & sounds to get it up to spec with our latest design practices.


misode (changes requested):
When the two cauldrons are close to each, it's possible for the rabbit to releport back to the original cauldron in the same tick. This depends on the arbitrary order of the zauber cauldron ticking.

Bloo-dev:
> back to the original cauldron in the same tick.

As in, it teleports from cauldron A to B, but B also has a player near it so it teleports from B to A within the same tick if B is ticked later than A?

Not sure whether it is worth fixing that -- this entire mechanic is an easter egg, and barely encountered in-game. To fix this, I'd have to leave an exclude tag on the rabbit accross ticks and clean it up later. The target cauldron is picked at random so in a real in-game situation those two cauldrons would have to be the only ones loaded.

misode (approved):

PR #982, opened 2024-03-12
Vertical Rails: Advancement grant fix

EpyonProjects:
Distance check seems to be failing for some reason, so instead check and grant advancement directly to the passengers


misode (approved):
I tested this in-game, and it seems like you now need `distance=..42`. But updating this to use `on passengers` is a good change regardless.

PR #984, opened 2024-03-15
Upload artifact of failing test worlds

misode:
Current problems:
- Tests are run at a random position in the world, so when you load the world you can't find the tests (there is a single line in the log with the coordinates, for example 4443907, -59, -13802014)
- Dummies are not saved, and these are often the things you want to check
- Ideally the game should be frozen when you enter it, with `/tick freeze`


misode:
I think all of the improvements will have to come from the packtest mod itself, and the test world is still usable, so I'll merged this and maybe it'll turn out to be useful.

PR #989, opened 2024-03-17
lib trees: Reduce selectors in tick

EpyonProjects:
Currently has 3 `@e` selectors in a tick clock.  This function is usually the highest percentage of the tick on the Public Server.

### DIFF
--- a/lib_trees/data/gm4_trees/functions/tick.mcfunction
+++ b/lib_trees/data/gm4_trees/functions/tick.mcfunction
@@ -1,4 +1,4 @@
 schedule function gm4_trees:tick 1t
 
 # process saplings
-execute as @e[type=marker,tag=gm4_tree_sapling] run function gm4_trees:sapling/process
+execute as @e[type=marker,tag=gm4_tree_sapling] at @s run function gm4_trees:sapling/process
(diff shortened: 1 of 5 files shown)


misode:
I'm not 100% sure this will improve performance, so please test and profile this on the server

EpyonProjects:
That was the goal!  We have a _lot_ of markers and the entity count is very high.  Just a quick test doesn't hurt

lib_trees/data/gm4_trees/functions/tick.mcfunction
```diff
@@ -1,9 +1,4 @@
 schedule function gm4_trees:tick 1t
 
-# upgrade old entities
-execute as @e[type=marker,tag=gm4_fruiting_sapling] run function #gm4_trees:upgrade_fruiting_to_general
-
 # process saplings
-execute as @e[type=marker,tag=gm4_tree_sapling] at @s run function gm4_trees:sapling/process
-# process leaves
-execute as @e[type=marker,tag=gm4_tree_leaf] at @s if block ~ ~ ~ #gm4_trees:empty_block run function #gm4_trees:destroy_leaf
+execute as @e[type=marker,tag=gm4_tree_sapling] run function gm4_trees:sapling/process
```
misode, line 4: Missing `at @s`

misode (changes requested):

misode (approved):

PR #992, opened 2024-04-09
Mountaineering: fix NBT not being saved when crafting new skis and crampons

AdequateUsername:
NBT of existing iron/chainmail boots were not previously saved when crafting skis/crampons, despite that being desired behaviour. This PR fixes that.


EpyonProjects:
Tested in singleplayer, looks good!

misode:
This has already been fixed in the 1.20.5 update, I think it'll be best to merge this with the future `ver/1.20` branch

misode:
I don't think anyone has tested this yet

misode (approved):
I'm a bit confused why this works, but I tested and everything seems to work...

PR #996, opened 2024-05-02
Combat Expanded 1.20.5

TheThanathor:
To-do list:

- [x] Move item tag to components
- [x] Update loot tables
    - [x] Armor
    - [x] Weapons
- [x] Move item_modifiers to inline with macro's
- [x] Test functionality of modifiers
- [x] Test functionality of mob modifiers
- [x] Remove modifiers that don't work well
_Burning_,_Sly_,_Twofold_
    - [x] Add / Update armor modifiers:
    - [x] NEW Acrobatic: improves movement when not under attack
    - [x] NEW Weightless: reduces gravity on wearer
    - [x] NEW Goliaths: increases player size
    - [x] NEW Shrinking: reduces player size
- [x] Add / Update weaponmodifiers:
    - [x] Giant: increased attack range
    - [x] NEW Long: reduced attack speed, increased attack range
- [x] Update Watchers
- [x] Update debug functions

Additional patch notes:
- removed compatibility with older versions - Devouring will no longer work
- improved chances for better stats on gear
- killing mobs now reduces regen cooldown down to 2.4 instead of 0.8 seconds


gm4_combat_expanded/data/gm4_combat_expanded/functions/armor/modifier/type/canine/wolf_remove_armor.mcfunction
```diff
@@ -0,0 +1,12 @@
+# remove armor from wolf
+# @s = wolf that is waering armor
+# at @s
+# run from armor/type/canine/wolf_interact_check
+
+# spawn item with the data
+data modify storage gm4_combat_expanded:temp body_armor_item.data set from entity @s body_armor_item
+function gm4_combat_expanded:armor/modifier/type/canine/eval_dropped_armor with storage gm4_combat_expanded:temp body_armor_item
```
misode, line 8: Instead of using a macro for this, you can summon the item and then use `item replace entity <item> contents from entity @s body_armor_item`

gm4_combat_expanded/data/gm4_combat_expanded/functions/armor/modifier/type/sword_ring/damage_dealt/apply_durability.mcfunction
```diff
@@ -8,10 +8,7 @@ execute store result storage gm4_combat_expanded:temp damage int 1 run scoreboar
 scoreboard players operation $sword_damage gm4_ce_data < $sword_durability gm4_ce_data
 
 # apply to correct piece of armor
-execute if score $sword_id gm4_ce_data matches 0 run item modify entity @p[tag=gm4_ce_target] armor.head gm4_combat_expanded:set_damage
-execute if score $sword_id gm4_ce_data matches 1 run item modify entity @p[tag=gm4_ce_target] armor.chest gm4_combat_expanded:set_damage
-execute if score $sword_id gm4_ce_data matches 2 run item modify entity @p[tag=gm4_ce_target] armor.legs gm4_combat_expanded:set_damage
-execute if score $sword_id gm4_ce_data matches 3 run item modify entity @p[tag=gm4_ce_target] armor.feet gm4_combat_expanded:set_damage
+execute as @p[tag=gm4_ce_target] run function gm4_combat_expanded:item_modify_eval/set_damage
```
misode, line 11: `set_damage` looks to be a macro function, yet it isn't given any parameters.

gm4_combat_expanded/data/gm4_combat_expanded/functions/item_modify_eval/chest_update.mcfunction
```diff
@@ -0,0 +1,2 @@
+
+$item modify entity @s armor.chest {function:"minecraft:set_components",components:$(components)}
```
misode, line 2: All of these `item_modify_eval` functions are missing a header comment with info about where it's called from.

gm4_combat_expanded/data/gm4_combat_expanded/loot_tables/armor/identification/modifier/vamp.json
```diff
@@ -13,6 +13,7 @@
               "modifiers": [
                 {
                   "attribute": "minecraft:generic.armor",
+                  "replace": false,
```
misode, line 16: This is not where `replace` goes. You need to put it at the same level as `function` and `modifiers`.

gm4_combat_expanded/data/gm4_combat_expanded/loot_tables/armor/identification/modifier/totem.json
```diff
@@ -188,6 +174,7 @@
         {
           "function": "minecraft:set_name",
           "entity": "this",
+          "target": "item_name",
```
misode, line 177: No other module uses `item_name` as of now. Are you sure that these items will still stack with existing items?

Imo we should keep using `custom_name` for the initial update, and then update all modules to switch to `item_name` when appropriate. This would also involve removing the unnecessary `"italic": false`, and using the `rarity` component instead of text component colors.

TheThanathor: These are all armor pieces, so there is no stacking to begin with. `item_name` seemed like the preferred thing to change

gm4_combat_expanded/data/gm4_combat_expanded/loot_tables/armor/identification/modifier/vamp.json
TheThanathor: fixed this, did not fix the bug however

misode:
I'll merge this so it can be tested with the rest of the 1.20.5 changes. Thanks!

PR #997, opened 2024-05-07
Lightning in a bottle 1.20.5

EpyonProjects:
Uses item modifier and `item modify block` instead of loot table and `loot replace block`


misode:
Looks good, this fixes all the liab issues afaict!

PR #998, opened 2024-05-07
Balloon Animals: 1.20.5 update

EpyonProjects:
- remove extra animal custom data value
- remove modifier since the loot table can now copy just the expected value
- Leash to leash
- loot table reference name key is now value
- animal type is now lead item name


misode:
Is it intended that purchased animals get some particle effects for a short moment?

EpyonProjects:
Why not :)

misode:
Animals before being purchased have the levitation and slow falling effects with particles hidden, after purchasing they have slow falling and resistance, both with particles shown. So it feels a little bit inconsistent.

Also I didn't remember that they had particles from testing in previous versions. But if it's always been like that, I'm happy to merge.

EpyonProjects:
I think if the balloon had visible particles, the color change of purchasing wouldn't be very obvious.  Doesn't matter to me whether it's mismatched, both showing, or both not.

But yes, I didn't touch those effects for this update, so it was like that before.

PR #999, opened 2024-05-07
Relocators: 1.20.5 update fix

EpyonProjects:
- copy_nbt data to copy_custom_data
- change nbt function to lore


misode:
Can confirm that this makes the module work correctly.

Issue #985, opened 2024-03-15
Guidebook markers are not cleaned up correctly


misode:
Guidebook is leaving around markers tagged `gm4_guide_<module>`. This may be an issue on Paper only.

BluePsychoRanger:
Fixed in #1015 with the removal of runtime alphabetical sort. All markers that were left over will not be automatically deleted though. This command can safely delete all extraneous markers:
`/execute as @e[type=marker] if data entity @s data.toc_line run kill @s`
`/kill @e[type=marker,tag=gm4_guide_sorter]`

Bloo-dev:
> Fixed in #1015 with the removal of runtime alphabetical sort. All markers that were left over will not be automatically deleted though. This command can safely delete all extraneous markers: `/execute as @e[type=marker] if data entity @s data.toc_line run kill @s` `/kill @e[type=marker,tag=gm4_guide_sorter]`



> Fixed in #1015 with the removal of runtime alphabetical sort. All markers that were left over will not be automatically deleted though. This command can safely delete all extraneous markers: `/execute as @e[type=marker] if data entity @s data.toc_line run kill @s` `/kill @e[type=marker,tag=gm4_guide_sorter]`



> Fixed in #1015 with the removal of runtime alphabetical sort. All markers that were left over will not be automatically deleted though. This command can safely delete all extraneous markers: `/execute as @e[type=marker] if data entity @s data.toc_line run kill @s` `/kill @e[type=marker,tag=gm4_guide_sorter]`

Closing this, even if it may still be an issue in 1.20.

PR #1015, opened 2024-08-03
Guidebook 1.21

BluePsychoRanger:
# Another Major Change (Guidebook v3.0.0)
All modules will need to be redownloaded to support the new guidebook module. Most of the guidebook json files haven't changed since the beet script handles the generate files.

## Changes
Removed runtime alphabetical sort
- it was only necessary for the distinction between Resurrecting Zombie and Resurrecting Skeletons
- it caused inconsistent, difficult to reproduce bugs that just broke the table of contents entirely

Removed binary search tree in favor of macros
- macros made the lectern table of contents much better as a direct insert of the lectern page number could be done instead

Removed separate loot tables for each module's book
- due to the way the books are generated, there is only a loot table for the default book
- this doesn't change survival gameplay
- lecterns don't even use loot tables at all anymore

Unlocked page info is probably faster now
- instead of building the book info upon changing the book, the book info is only updated once the player obtains an unlock advancement 
- book info is now stored in a Player Database which stored the book info for each module per player
  - when a player db entry is created, the player's advancements are checked to ensure their unlocked pages don't get lost

Update recipe files to use components
- only visually affects SCUBA Gear and Boots of Ostara
- Other recipes were update too in case we do want to use those to generate files in the future (unlikely cuz of the new nbt output crafting, but we'll see)

## Future Optimizations
When modules are changed, the entire player database is wiped (restored with advancements, so nothing breaks)
- Necessary since if a module is enabled/disabled it can change the book contents (e.g. the pages for desire lines change if boots of ostara and/or metallurgy are installed), and thus the page info needs to be updated for each player
- Currently deletes the entire player db since it's the easiest way to do it
- Can be sped up in "runtime" by deleting only entries for modules that have such complementary pages 
  - will slow down the initial time required right after a `/reload` when modules are changed since it would require checking each player db entry for matching modules (this is nontrivial since the player db entries are currently stored only as a map; may require storing an extra list with each key to iterate through the entire player db)

The player db can become very large with duplicate data
- Each player has a copy of their entire unlocked book data which may have duplicate pages as another player
- This could be solved with a mapping system which maps an index to a variation of a page, so only one of each variation is created once and a link is created in the player db entry
- Instead of storing the pages as strings, each player db entry (i.e. one for each player) would only contain integers which is significantly less storage
- The mapping system would reduce the file size since the only duplicated data would be the index values
- We usually don't consider space constraints, but the file may become significantly large on servers with many players
- This would technically add more time to access a book since it'd have to do an extra (macro) data read


BluePsychoRanger:
Fixes #985

TheThanathor (changes requested):
- text after recipe overlaps with RP (eg Boots of Ostara)
- books in lecterns do not work, `open_updating_lectern` is incorrect
- numbers in crafting recipes do not work with RP

BluePsychoRanger:
> * text after recipe overlaps with RP (eg Boots of Ostara)
> * books in lecterns do not work, `open_updating_lectern` is incorrect
> * numbers in crafting recipes do not work with RP
> * translations are missing from `translations.csv`

These have been fixed except for the numbers in crafting recipes

gm4_guidebook/data/gm4_guidebook/loot_table/guidebook.json
catter1, line 1: Based on conventions of the other modules, should this loot table be in an `items` subfolder? (`gm4_guidebook:items/guidebook`)

BluePsychoRanger: Yeah, in v2 each module had its own loot table, so to make it simple no subfolders were used. However v3 uses just the two loot tables, so they can be put into their proper subfolders

gm4_guidebook/data/gm4_guidebook/function/lecterns/restore/hotbar.mcfunction
```diff
@@ -3,12 +3,12 @@
 # at @s
 # run from gm4_hotswap_hotbars:swap/restore_box
```
misode, line 4: Copy-paste error in these header comments

gm4_guidebook/beet.yaml
```diff
@@ -1,6 +1,6 @@
 id: gm4_guidebook
 name: Guidebook
-version: 2.2.X
+version: 3.0.X
```
misode, line 3: `text.gm4.guidebook.jump_to_section` needs to be added to the `translation_linter_ignores` list in this file. Also double check whether there are any translation keys that can be removed.

misode (changes requested):
1. When you take a guidebook from a lectern, you need to refresh findings, the ToC starts empty. It also loses the active section info.
2. When you get your initial book, it would be nice if the ToC was already refreshed.
3. When you click the wiki link for included modules (like ender bolt) it doesn't go to the sub page.
4. With the RP active, there are no page numbers, not sure if this is feasible to add.

BluePsychoRanger:
> 1. When you take a guidebook from a lectern, you need to refresh findings, the ToC starts empty. It also loses the active section info.
WAI in order to keep the system simpler. I've updated it to populate the ToC though.
> 2. When you get your initial book, it would be nice if the ToC was already refreshed.
This has been implemented
> 3. When you click the wiki link for included modules (like ender bolt) it doesn't go to the sub page.
Fixed
> 4. With the RP active, there are no page numbers, not sure if this is feasible to add.
Yeah, this isn't really feasible without some complex logic
> 5. The boots of ostara recipe is a bit broken with the RP
This has something to do with the item overlays. I'm looking into it.

BluePsychoRanger:
Okay, I believe all issues have been addressed

gm4_guidebook/generate_guidebooks.py
```diff
@@ -2029,11 +1889,19 @@ def generate_toast_model(book: Book, ctx: Context) -> Model:
...
 
   # create model for new texture
+  textures = {
+    "layer0": f"{ctx.project_id}:gui/guidebook/{book.id}"
+  }
+   # some items will tint layer0, so we special case those here.
+  if book.icon.id.removeprefix("minecraft:") in ("leather_helmet", "leather_chestplate", "leather_leggings", "leather_boots", "leather_horse_armor", "potion", "splash_potion", "lingering_potion", "tipped_arrow"):
+    textures = {
+      "layer0": "gm4_guidebook:gui/font_magic/blank",
+      "layer1": f"{ctx.project_id}:gui/guidebook/{book.id}"
```
misode, line 1899: Both mecha and minecraft complain because the `gui/font_magic/blank` texture doesn't exist. Instead it is currently at `font_magic/blank`. However even after changing this, it still has the problem where these textures are not loaded to the `blocks.json` atlas, which I believe is required to render the textures on item models.

BluePsychoRanger: I've fixed the issue in-game, but can't figure out why mecha still complains. 
(This is pasted from commit e7ccfa2)
> Mecha still complains that the texture is missing, but the built resource pack works in game without throwing a warning. I've even tried modifying the `minecraft/atlases/block.json` in the `base` module, but it still shows a warning

SpecialBuilder32: These warnings are from other modules, which won't have the guidebook textures loaded when they build. I would recommend using `gm4:item/empty` from the base module, or if the image size is an issue, moving `font_magic/blank` to `gm4:gui/advancement/blank` in the base.

BluePsychoRanger: Oh that's perfect. I didn't know we already had an empty texture laying around.

gm4_boots_of_ostara/data/gm4_boots_of_ostara/gm4_recipes/boots_of_ostara.json
SpecialBuilder32, line 1: Is there a reason why this no longer references a loot table file? Defining the data of our custom items in multiple places makes it harder to deal with updating them.

gm4/skin_cache.json
SpecialBuilder32, line 1: I'm not quite sure why this file keeps getting generated on full builds, but as zauber liquids is currently on hold these changes should not be part of this PR.

gm4_guidebook/generate_guidebooks.py
```diff
@@ -34,6 +34,31 @@
...
+
+DEFAULT_OVERLAY_COLORS = {
+  "empty": 0
+}
+
+IS_DYED = [
+  "minecraft:leather_boots",
+  "minecraft:leather_chestplate",
+  "minecraft:leather_helmet",
+  "minecraft:leather_leggings"
```
SpecialBuilder32, line 52: Should also included `leather_horse_armor` and `wolf_armor` for completeness?

SpecialBuilder32 (changes requested):
The last translation warnings are from midnight menaces/mysterious midnights. Those just need a babelbox backfull run on them. `beet -s meta.gm4.babelbox_backfill=True guidebook mm1 mm2`

gm4_boots_of_ostara/data/gm4_boots_of_ostara/gm4_recipes/boots_of_ostara.json
BluePsychoRanger: The boots of ostara loot table was changed to reference an item modifier. Instead of basically rewriting mecha to resolve every loot table condition, I opted to just hardcode special cases like this

gm4_guidebook/generate_guidebooks.py
BluePsychoRanger: Done. I've unified the definition, so the tint checker uses this list instead of another inline list

BluePsychoRanger:
> There's also these warnings in my log, maybe that could be fixed? 
Found the issue! Should be fixed now

misode (approved):

gm4_boots_of_ostara/data/gm4_boots_of_ostara/gm4_recipes/boots_of_ostara.json
SpecialBuilder32: Well it doesn't bring me joy but thats probably the most efficient use of time/effort then 👍

SpecialBuilder32 (approved):
I did a quick test and everything seemed to work as its supposed to!

There are a few things I'm unhappy with from a beet standpoint, but they're pretty insubstantial and due to Mojang changing how things work all the time so I'll just deal with imperfection 😛 
(Having to duplicate loot table information multiple places, static item sprite sheets ect...)

And just as a final note, it is now more possible to auto-generate the font custom item sprite sheet from source textures due to some work by AirDox_ (https://discord.com/channels/900530660677156924/1250766751675977799). So this is an earworm about that being more doable in upcoming versions and maybe fun to implement

PR #1016, opened 2024-08-05
Fix Orb of Ankou Granting all Pneumas

BluePsychoRanger:
- fixes `equipped` predicates generated via the `generate_pneumas` script
- shows OoA on website
- fix strict typing error

### DIFF
--- a/gm4_orb_of_ankou/generate_pneumas.py
+++ b/gm4_orb_of_ankou/generate_pneumas.py
@@ -5,7 +5,7 @@ from typing import Any
 SUPPORTED_LOOTING = 10
 
 entities = {}
-pneumas = []
+pneumas:list[str] = []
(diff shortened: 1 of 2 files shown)


gm4_orb_of_ankou/generate_pneumas.py
```diff
@@ -95,15 +95,17 @@ def generate_pneuma_predicates(ctx: Context):
...
       "condition": "minecraft:entity_properties",
       "entity": "this",
       "predicate": {
         "equipment": {
           "offhand": {
-            "tag": "gm4_orb_of_ankou:pneuma_container",
-            "nbt": nbt
+            "items": "#gm4_orb_of_ankou:pneuma_container",
+            "components": {
+              "minecraft:custom_data": custom_data
```
misode, line 107: This will do an exact component check. Instead use an item sub predicate `"predicates"` instead of `"components"`

BluePsychoRanger: Thanks, I've updated it to use that

PR #1017, opened 2024-08-05
Add Crafter Support to Relocators

BluePsychoRanger:
This should be the only feature change for 1.21 since it's the only new block entity other than the vaults and spawners which can't be picked up. I've tested it and it seems to work properly.


misode (approved):
Works as expected!

PR #1018, opened 2024-08-05
Enchantment Extractors 1.21

BPR02:
Chances are up for debate, but these are what I decided for now. Tested this and it works for all enchantments with all levels.

Breach: 85, 70, 50, 35* (similar to piercing [90, 70, 50, 30])
Density: 95, 70, 60, 50, 30 (matches smite and sharpness)
Wind Burst: 70, 50, 30 (matches riptide)

*3% chance to have curse of vanishing

This PR also adds support for custom enchantments
- new function tag that runs when an item extraction occurs, so other packs can add their own enchantment extraction logic
- new function tag that runs before an item extraction occurs so other packs can treat some enchantments as curses (cursed items only extract the curse)
- new function tag that runs before the enchantments are cleared from the item in storage so other packs can save enchantments that shouldn't get cleared (i.e. technical enchantments)
- new function tag that runs after the item's enchantments are removed in storage, but before it's updated in the block so other packs can restore those technical enchantments that shouldn't have been removed


misode (commented):
> This PR also adds support for custom enchantments

I don't think we should implement this using function tag hooks, but instead try to do something like parsing the enchantment compound. I doubt packs that add enchantments will take the time to implement our 4 technical function hooks, they will just expect our module to work with them.

I'd be fine with merging just the 3 added enchantments without the rest though.

misode (approved):

PR #1027, opened 2024-08-12
Fix Holographic Tags Accepting Any Item

### LINKED ISSUE #1026: [Important] Holographic Tags Accept ANY item instead of just Nametags
**Describe the bug**
Placing any renamed item in an Item Frame deletes teh item and turns it into a Holographic tag, thereby causing item loss.

**To Reproduce**
1. Place an Item Frame
2. Rename some item
3. Place it in the Item Frame
4. Watch as the item is deleted and replaced by a Holographic Tag

**Expected behavior**
Only Nametags should be able to make Holographic Tags.

**Fix**
The issue was tracked down to a change in the data checks caused by the 1.20.6 update, causing the check to succeed for _any_ item with a custom name, not just `minecraft:nametag`.

**This fix should be backported to 1.20.6 as well as 1.21.**

**Environment**
 - Minecraft version: 1.20.6-1.21

**Additional context**
Spotted thanks to the reports of two users on our Discord:
- Original Report (May 2024) https://discord.com/channels/151141188961828864/1245110876705656912
- Second Report (August 2024) https://discord.com/channels/151141188961828864/801191203966156891/1272328270926581912

BluePsychoRanger:
Fixes #1026


misode (approved):
During the meeting we said this should be changed to move the `*[custom_name]` check into the predicate, but it isn't possible to just check for the existence of a component. Therefore I think this is the best possible fix.

PR #1031, opened 2024-08-20
Everstone 1.21

BluePsychoRanger:
### New Baby Mob Support:
Armadillos

### New Conversions
Skeleton to Bogged:
- Aging: in swamp and in water
- Deaging: not in swamp
- Average time: 1 hour

Bogged to Skeleton:
- Aging: sheared and not in swamp
- Deaging: in swamp
- Average time: 1 hour

### Vehicle Support
Converted mobs will mount the vehicle the original mob was riding.

### Sky-Access Optimization
Now uses the `can_see_sky` predicate instead of checking all blocks above the entity (plus, it was outdated and only checked up to Y=255)

### Biome Tag Support
Made biome checks target specific tags which by default target the `c` tags from fabric/neoforge. At the moment the required `c` tags are added directly (copied from the generated fabric files), but the implementation may change depending if the `c` tags become a smithed library.

#### Changelog Checklist
- [x] Add armadillos to baby list
- [x] Make converted mobs ride the original mob's vehicle
- [x] Add bogged -> skeleton conversion
- [x] Add skeleton -> bogged conversion
- [x] Simplify sky-access predicate
- [x] Add `c` tags for biome compatibility


misode (changes requested):
Preventing armadillo ageing works correctly.

I tested the mount keeping behavior, and at least for skeleton -> bogged, this did not work correctly. Instead the passenger mob simply vanished...

Additionally, although this may have been already the case, mobs in a vehicle do not age/convert.

TheThanathor (approved):

TheThanathor:
> Preventing armadillo ageing works correctly.
> 
> I tested the mount keeping behavior, and at least for skeleton -> bogged, this did not work correctly. Instead the passenger mob simply vanished...
> 
> Additionally, although this may have been already the case, mobs in a vehicle do not age/convert.

I did not run into the same issue, ran the `function gm4_everstone:aging_clock` clock every tick and tested bogged -> skeleton, skeleton -> stray, and skeleton -> bogged. all converted and the resulting mob rode the pig.

BluePsychoRanger:
Perhaps misode tested and ran into the opposite issue. If something was riding a skeleton and the skeleton converted to a bogged, I believe anything originally riding the skeleton would be teleported to the void.

misode:
> mobs in a vehicle do not age/convert

It turns out this issue is caused by the skeleton's hitbox clipping into the ground, and thus the skeleton is not considered to be in water. Even though this is an edge case, maybe it would make more sense to check for water at the mob's eye level?
image

misode (approved):

PR #1053, opened 2024-09-23
Port custom crafting recipes to official JSON recipes

### LINKED ISSUE #980: Moving custom recipes to official recipes
The 1.20.5 update is adding custom item support for the output item of recipes. While this doesn't solve all our problems, we can get pretty far and we should explore our path forward. This issue lists some of the changes required if we decide to move to official recipes.

## Advantages
- Players can easily look up our items in the recipe book
- Our recipes will automatically work with the crafter,

## Module changes
Some modules would no longer make sense or would need to be rebranded. These are just my initial thoughts, nothing has been decided yet!
- **Auto Crafting**: Deprecate and remove in 1.21, since the crafter will cover everything. If too complicated to upgrade this module we could also remove it in 1.20.5
- **Standard Crafting**: Would now just be a recipe pack, maybe rebrand?
- **Forming Press**: Consider if we still need this module. Do we move all recipes to standard crafting?
- **Record Crafting**: Same as above
- **Lib Custom Crafters**: Remove

## Recipe changes
We cannot check for item components in the recipe input, and we currently use that for a few recipes. Some recipes can be modified, for others we would need to be a bit more creative.
- **Auto Crafter**: uses custom crafter in input, propose replacing with crafting table
- **Boots of Ostara**: preserves input item data
- **Ender Hopper Minecart**: uses an ender hopper in input: propose hopper minecart with ender eye and respawn anchor
- **Heart Canister Tier 2**: uses tier 1 in input
- **Liquid Minecart**: uses a liquid tank in input: propose using liquid tank recipe with hopper minecart instead of hopper
- **Crampons**: preserves input item data
- **Skis**: preserves input item data
- **Relocator**: uses silk touch books in input
- **Scuba Tank**: uses potion of water breathing in input
- **Soft Defuser**: uses potion of harming 2 and landmine in input
- **Secret Trapped Sign**: uses normal trapped sign, propose incorporating black dye in original recipe

misode:
- Closes #980 

## Notes
- Leaving out: auto crafter, heart canister tier 2, relocators, scuba gear
- Slightly modifies: ender hopper minecart, liquid minecart
- Discarding item data: boots of ostara, mountaineering

## Todo
- [x] Create initial recipe JSON files
- [ ] Add recipe unlock advancements
- [ ] Update record crafting advancement
- [ ] Update wiki, guidebook, readme
- [ ] Maybe: overwrite some vanilla recipes to add `group` field for forming press recipes
- [ ] Maybe: rebrand record crafting, standard crafting, and forming press

## Future work
- Switch from `custom_name` to `item_name` and `rarity`
- Discuss whether we want to add a `[GM4] <module name>` tag line
- Remove auto crafting module

### DIFF
--- a/gm4_block_compressors/data/gm4_block_compressors/recipe/compressor.json
+++ b/gm4_block_compressors/data/gm4_block_compressors/recipe/compressor.json
@@ -1,0 +1,35 @@
+{
+  "type": "minecraft:crafting_shaped",
+  "category": "redstone",
+  "pattern": [
+    "IPI",
+    "pOp",
+    "CCC"
+  ],
+  "key": {
+    "I": {
+      "item": "minecraft:iron_ingot"
+    },
+    "P": {
+      "item": "minecraft:purpur_block"
+    },
+    "p": {
+      "item": "minecraft:piston"
+    },
+    "O": {
+      "item": "minecraft:obsidian"
+    },
+    "C": {
+      "item": "minecraft:cobblestone"
+    }
+  },
+  "result": {
+    "id": "minecraft:player_head",
+    "components": {
+      "minecraft:custom_model_data": "block/block_compressor_full",
+      "minecraft:profile": "$block_compressor",
+      "minecraft:custom_data": "{gm4_machines:{id:'block_compressor'}}",
+      "minecraft:custom_name": "{\"translate\":\"block.gm4.block_compressor\",\"fallback\":\"Compressor\",\"color\":\"white\",\"italic\":false}"
+    }
+  }
+}
--- a/gm4_boots_of_ostara/data/gm4_boots_of_ostara/recipe/boots_of_ostara.json
+++ b/gm4_boots_of_ostara/data/gm4_boots_of_ostara/recipe/boots_of_ostara.json
@@ -1,0 +1,41 @@
+{
+  "type": "minecraft:crafting_shaped",
+  "category": "equipment",
+  "pattern": [
+    " S ",
+    "MLG",
+    " W "
+  ],
+  "key": {
+    "S": {
+      "item": "minecraft:wheat_seeds"
+    },
+    "M": {
+      "item": "minecraft:moss_block"
+    },
+    "L": {
+      "item": "minecraft:leather_boots"
+    },
+    "G": {
+      "item": "minecraft:grass_block"
+    },
+    "W": {
+      "item": "minecraft:water_bucket"
+    }
+  },
+  "result": {
+    "id": "minecraft:leather_boots",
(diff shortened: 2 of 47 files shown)


BluePsychoRanger:
I think @Bloo-dev should weigh in on this convo, but I think porting the forming press recipes into normal crafting kinda bypasses the extra step that makes those double brick recipes warranted to be in the game. Perhaps we can switch it to just taking item stacks and using those (e.g. 8 cobblestone in one slot + 1 clay would create stone bricks). The original CC system wouldn't allow for differently stacked item counts, but given that system will eventually be gone we can rewrite Forming Press (again :P)

BluePsychoRanger:
> When guidebook added the `gm4_recipe` format for displaying recipes, it was done with future goal of being a format we could easily generate vanilla (or smithed crafter) recipes from. I'd prefer to only have one or the other, to keep the recipe defined in only one place.

I believe the ideal situation is to get rid of `gm4_recipe` completely which will only be possible when we get proper custom items. At that point the custom crafter can be discontinued and we'd need to update the guidebook to parse the (normal data driven) recipe files instead. Misode probably left the `gm4_recipe` files since removing them would mean updating the guidebook script...again.

> I also wish we didn't have to duplicate custom item definitions, once in the loot table (which is a nice /give alternative) and once in the recipe. I think it might be good to either generate/mutate the recipe files using the loot table definitions as a source of truth, or generate the loot table files from these recipe definitions as a source of truth.

We could probably do something similar to CMD where we inject custom data into the `recipe` files and replace them during the build. I'm in favor of using beet to parse the loot table and inject it into the recipe output, but that would take some work.

> On a side note, is smithed crafter mostly dead with componentized output recipes? Is smithed crafter cross-features still something we think would be good?

I think ideally the Smithed Crafter is phased out just like the GM4 custom crafter. Given we have had hints of custom items, I don't think the smithed crafter should be a priority. The smithed libraries are pretty neglected, but rx has been wanting to do a "weld plugin" which would take all data driven recipe files and add them to the crafter. So if that ever becomes a thing (unlikely to happen before custom items tbh), we'd just need our recipe files ready.

misode:
I think this PR is ready now. I did not switch `gm4_recipes` to mojang's recipe files, so there is still some duplicate information. Since not all of our recipes can be described using the official JSON recipes, it would mean maintaining two separate formats for the guidebook.

This PR is meant as an initial step towards more standardized items and recipes, to make these recipes more accessible to players. It can definitely be improved in the future.

SpecialBuilder32 (approved):
I like the automatic overlay stuff you've got here Misode 💙

PR #1055, opened 2024-09-26
Add Nullscape support to End Fishing

catter1:
Currently, End Fishing only supports the Vanilla biomes. This PR adds support for [Nullscape biomes](https://stardustlabs.miraheze.org/wiki/Nullscape#Biomes) - meaning loot can be gained by fishing in Nullscape's biomes. This is also important, since Nullscape prevents a couple of vanilla end biomes from generating.

### History
This was previously discussed over a year ago, but it was not easily possible due to how End Fishing is set up. Now that it is possible to reference biome tags in loot tables, this idea is possible (and can be expanded to other custom end biomes if needed in the future).

### Implementation
Three new biome tags were made, following the `gm4_end_fishing:nullscape/[biome]` schema. These tags reference the respective Nullscape biome, setting `required` to `false`. Then the loot tables reference the tags as they would any other vanilla end biome.

### Nullscape Biome Loot
- **Shadowlands** - Copies the End Midlands loot distribution, since End Midlands does not generate in Nullscape
- **Void Barrens** - Copies the End Barrens loot distribution, since End Barrens does not generate in Nullscape
- **Crystal Peaks** - Has the default loot distribution, except:
  - **Valuables** - Obsidian is replaced with Crying Obsidian, and added an extra Amethyst Block entry
  - **Scattered Treasure** - Made Captain's Wings *slightly* more common (since it's a mountain biome)

### Note
While working on the loot tables, I noticed that Captain's Wings (elytra) was an inline definition in the `scattered_treasure` loot table. Instead of making a new PR for it, I just moved it to `gm4_end_fishing:items/captains_wings` (just like `ravaged_wings` already is) in bb4c0aa. This can be reverted if for some reason it is not wanted.

### DIFF
--- a/gm4_end_fishing/data/gm4_end_fishing/loot_table/gameplay/fish/debris.json
+++ b/gm4_end_fishing/data/gm4_end_fishing/loot_table/gameplay/fish/debris.json
@@ -82,6 +82,12 @@
                   "predicate": {
                     "biomes": "minecraft:end_highlands"
                   }
+                },
+                {
+                  "condition": "minecraft:location_check",
+                  "predicate": {
+                    "biomes": "#gm4_end_fishing:nullscape/shadowlands"
+                  }
                 }
               ]
             }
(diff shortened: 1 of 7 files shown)


misode (approved):
I haven't tested this in-game, but the changes look good!

PR #1065, opened 2024-11-05
Metallurgy: Other explosions create ore lumps

EpyonProjects:
Adds creepers, withers, and wither skeletons alongside TNT creating ore lumps when destroying stone

### DIFF
--- a/gm4_metallurgy/data/minecraft/loot_table/blocks/stone.json
+++ b/gm4_metallurgy/data/minecraft/loot_table/blocks/stone.json
@@ -37,7 +37,12 @@
                   "condition": "minecraft:entity_properties",
                   "entity": "this",
                   "predicate": {
-                    "type": "minecraft:tnt"
+                    "type": [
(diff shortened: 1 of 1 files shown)


misode:
Would it be better to detect the `#is_explosion` damage type tag?

EpyonProjects:
Hmm.. I was thinking about how to future proof this or to add expansion compatibility, but I'm not sure how or where to input a tag

misode (approved):

EpyonProjects:
Oh, the stone entry at the bottom uses "survives explosion", wouldn't that be sufficient?  Just any explosion?

misode:
> Oh, the stone entry at the bottom uses "survives explosion", wouldn't that be sufficient? Just any explosion?

That condition always succeeds for tnt explosions and has a chance of failing (based on distance) for other explosions. But it also always succeeds for any other block destroy cause. It's present in most vanilla blocks.

PR #1070, opened 2024-11-24
Blossoming Pots

runcows:
Blossoming Pots initial commit.
Ready for code review and further discussion, I think...


gm4_blossoming_pots/data/gm4_blossoming_pots/function/main.mcfunction
```diff
@@ -0,0 +1,5 @@
+execute as @e[tag=gm4_blossoming_pots.display.decorated_pot] at @s unless block ~ ~ ~ minecraft:decorated_pot run kill @s
+execute as @e[tag=gm4_blossoming_pots.display.flower_pot] at @s unless block ~ ~ ~ minecraft:flower_pot run kill @s
+execute as @e[tag=gm4_blossoming_pots.data.flower_pot] at @s unless block ~ ~ ~ minecraft:flower_pot run function gm4_blossoming_pots:flower/item/break with entity @s data
```
misode, line 3: These `@e` checks should have a `type=`. But also I feel like this could be a single main command with a shared `tag=` and a block tag, running a cleanup function that does more work.

misode (dismissed):
I would probably want to limit the number of display entities per pot to something like 5. Anything above that I feel goes beyond of what "feels vanilla". Especially the coral decorations, where it's using 41 display entities for a single pot. That's not only a performance issue, but I also don't think it looks good visually, since there are too many small details and you just don't see that with vanilla. This already has the coral fans, I think those are a better fit.

I was a bit unsure on how this module worked tech wise, with the storage. From my point of view coming from beet, I would probably generate a mcfunction for each supported item, which just had the item display summon commands, that would probably allow you more flexibility without confirming to a strict storage schema.
I think it is probably fine to keep using the storage, unless you come across a major hurdle.

I didn't do a full technical review yet, but it seems fairly well put together after a first look.

runcows:
> I would probably want to limit the number of display entities per pot to something like 5. Anything above that I feel goes beyond of what "feels vanilla". Especially the coral decorations, where it's using 41 display entities for a single pot. That's not only a performance issue, but I also don't think it looks good visually, since there are too many small details and you just don't see that with vanilla. This already has the coral fans, I think those are a better fit.

Totally get it. I feel a little bad removing the coral blocks because I worked hard on them when I made them, but I totally get it and am down to axe them. I'm curious on your thoughts for a limit of display entities. There's only 5 other things that have more than 5 displays. The next highest count after Coral Blocks is the 3rd stage of Chorus Fruit with 18, which im much more hesitant to remove. After that is Kelp and Twisting Vines which go up to 16 and I'm willing to chop those down a good bit. Bamboo and Glow Berries have 8 and Weeping Vines has 9, again willing to chop those down a little bit.

gm4_blossoming_pots/overlay_48/data/gm4_blossoming_pots/function/decorated/as_player.mcfunction
```diff
@@ -0,0 +1,15 @@
+# Sets up raycast, then redirects to as_marker, then revokes advancement
+# @s = player who interacted with a decorated pot
+# at @s
+# run from advancement gm4_blossoming_pots:item_in_pot
+
+# Raycast for the pot (set up range and search)
+execute store result score @s gm4_blossoming_pots.range run attribute @s minecraft:player.block_interaction_range get 1
```
misode, line 7: You don't actually need these overlays. Our `gm4/plugins/backwards.py` script will automatically find any attributes and create overlays for them.

runcows: Ah, awesome, noted. I'll have it implement that in a few hours when I get home.

gm4_blossoming_pots/beet.yaml
```diff
@@ -0,0 +1,41 @@
...
+      #project_id: null
+    #smithed:
+      #pack_id: null
+    planetminecraft: 
+      uid: null
+    video: null
+    wiki: null
+    credits:
+      Creator:
+        - runcows
```
Bloo-dev, line 39: Don't forget to add yourself to `gm4/CONTRIBUTORS.json`. There you can add links to your social media accounts.

gm4_blossoming_pots/beet.yaml
```diff
@@ -0,0 +1,41 @@
...
+      Icon Design:
+        - <no idea>
```
Bloo-dev, line 41: You!

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/as_player.mcfunction
```diff
@@ -0,0 +1,14 @@
...
+
+# Raycast for the pot (set up range and search)
+execute store result score @s gm4_blossoming_pots.range run attribute @s minecraft:block_interaction_range get 200
+execute anchored eyes run function gm4_blossoming_pots:flower/raycast/loop
+
+execute as @e[type=minecraft:marker,sort=nearest,tag=gm4_blossoming_pots.temp.flower_pot] at @s \
+    align xyz positioned ~.5 ~.5 ~.5 run function gm4_blossoming_pots:flower/as_marker with entity @s data
+
+# Reset advancement
+advancement revoke @s only gm4_blossoming_pots:interact_flower_pot
```
Bloo-dev, line 14: Probably best to do this first, just in case there is some runtime error causing the function to abort prematurely

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/as_player.mcfunction
```diff
@@ -0,0 +1,14 @@
+# Sets up raycast, then redirects to as_marker, then revokes advancement
+# @s = player who interacted with an empty flower pot
+# at @s
+# run from advancement gm4_blossoming_pots:interact_flower_pot
+
+# Raycast for the pot (set up range and search)
+execute store result score @s gm4_blossoming_pots.range run attribute @s minecraft:block_interaction_range get 200
```
Bloo-dev, line 7: Reading the interaction range for this is quite smart and should possibly become the new standard for the whole of Gamemode 4. We should discuss the performance impact of it internally.

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/raycast/loop.mcfunction
```diff
@@ -0,0 +1,14 @@
+# main loop for raycasting for the flower pot the player interacted with
+# @s = player who interacted with a flower pot
+# at @s anchored eyes, then marched forward 0.005 over and over
+# run from flower/as_player
+# run from this function...
```
Bloo-dev, line 5:
```suggestion
# run from flower/as_player and self
```

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/raycast/in_pot.mcfunction
```diff
@@ -0,0 +1,20 @@
+# checks if execution context is inside the flower pots hit box
+# @s = player who interacted with a flower pot
+# at @s anchored eyes, then marched forward 0.005 over and over. Complicated.
+# run from flower/raycast/loop
+
+# early return if not flower pot
+execute unless block ~ ~ ~ minecraft:flower_pot run return fail
+
+# summon marker for checking
+summon minecraft:marker ~ ~ ~ {Tags:["gm4_blossoming_pots.temp.flower_pot","smithed.strict","smithed.entity"],CustomName:'{"text":"gm4_blossoming_pots.temp_marker"}',data:{id:null}}
```
Bloo-dev, line 10: What is this `id` field?

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/raycast/in_pot.mcfunction
```diff
@@ -0,0 +1,20 @@
...
+
+# if temp marker in physical hit box, return 1
+# dxdydz all 0 bc of MC-123441.
+execute as @n[type=minecraft:marker,tag=gm4_blossoming_pots.temp.flower_pot,distance=..0.1] \
+    align xyz positioned ~0.312499 ~ ~0.312499 if entity @s[dx=0,dy=0,dz=0] at @s align xyz positioned ~-0.312499 ~-0.625 ~-0.312499 if entity @s[dx=0,dy=0,dz=0] \
+    run return 1
+
```
Bloo-dev, line 17: You could avoid having to select the marker by using `execute summon marker run function...` and calling a function which checks the markers physical location and kills it afterwards.

Bloo-dev (commented):
This is not a full review (yet), but I had a couple of minutes spare. More to come later.

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/raycast/in_pot.mcfunction
runcows: The temp marker summoned here has the players SelectedItem stored on it, as well as the players UUID, and rotation. This happens in flower/raycast/player_rotation. id is set to null here as the fall back value for if the SelectedItem of the player is null and doesn't write. This is to distinguish an empty hand right click.

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/raycast/in_pot.mcfunction
runcows: Thats true yeah. It would have to take on a majority of the logic of this function, but it would be probably easier to read and I imagine not needing the selector would be a bit better for efficiency.

gm4_blossoming_pots/beet.yaml
```diff
@@ -0,0 +1,41 @@
+id: gm4_blossoming_pots
+name: Blossoming Pots
+version: 3.0.X
+  # should I start the version number over for gm4 release?
```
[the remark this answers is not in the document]
runcows: I completely forgot I put this comment here, uh yeah I think I'll go with my current version number :D

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/raycast/in_pot.mcfunction
runcows: Implemented this in most recent commit

Bloo-dev (changes requested):
Some more input from me... still haven't looked at everything tho ):

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/as_player.mcfunction
```diff
@@ -0,0 +1,14 @@
...
+# run from advancement gm4_blossoming_pots:interact_flower_pot
+
+# Reset advancement
+advancement revoke @s only gm4_blossoming_pots:interact_flower_pot
+
+# Raycast for the pot (set up range and search)
+execute store result score @s gm4_blossoming_pots.range run attribute @s minecraft:block_interaction_range get 200
+execute anchored eyes run function gm4_blossoming_pots:flower/raycast/loop
+
+execute as @e[type=minecraft:marker,sort=nearest,tag=gm4_blossoming_pots.temp.flower_pot] at @s \
```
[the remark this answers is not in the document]
runcows: I think I originally did this as a testing measure when markers weren't always being found and just left around. This would catch the stray markers. You're right and this can be changed to @n selector, since the pack works ^^

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/raycast/player_rotation.mcfunction
```diff
@@ -0,0 +1,16 @@
...
+data modify entity @n[type=minecraft:marker,tag=gm4_blossoming_pots.temp.flower_pot,distance=..0.01] data.UUID set from entity @s UUID
+
+# player facing west
+execute if entity @s[y_rotation=45..135] run return run data merge entity @n[type=minecraft:marker,tag=gm4_blossoming_pots.temp.flower_pot,distance=..0.01] {data:{rotation:0}}
+# player facing north
+execute if entity @s[y_rotation=135..225] run return run data merge entity @n[type=minecraft:marker,tag=gm4_blossoming_pots.temp.flower_pot,distance=..0.01] {data:{rotation:90}}
+# player facing east
+execute if entity @s[y_rotation=225..315] run return run data merge entity @n[type=minecraft:marker,tag=gm4_blossoming_pots.temp.flower_pot,distance=..0.01] {data:{rotation:180}}
+# player facing south
+execute if entity @s[y_rotation=315..405] run return run data merge entity @n[type=minecraft:marker,tag=gm4_blossoming_pots.temp.flower_pot,distance=..0.01] {data:{rotation:-90}}
```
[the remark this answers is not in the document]
runcows: The run-return-run was actually my solution to the inclusivity issue, since I don't know how to make the float range non inclusive, other than giving one end a .999999. I can do that; at the time I wrote this, I just didnt like the look of all the 9's and opted for this, figuring returning early to skip the extra checks was also beneficial.

I had typed out something for the marker selection bit, but I just realized something that might make this whole marker selection discussion irrelevant. The data on the temp marker here is used as a macro in the as_marker function run following the raycast. That function then stores that data to a temp storage anyway and then the temp marker's data is never used again... So it probably makes much more sense to just store the data in storage immediately right here and use that storage as a macro instead for as_marker, avoiding setting data on or selecting the temp marker here.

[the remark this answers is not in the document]
runcows: Just pushed the y-rotation changes to use the storage instead of the marker. 

Any advice on the range inclusivity?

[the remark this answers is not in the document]
Bloo-dev: Will have a look tomorrow evening, it's getting late here!

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/raycast/in_pot.mcfunction
runcows: This is no longer relevant, as selecteditem id is stored in storage directly now. Marking as resolved.

gm4_blossoming_pots/data/gm4_blossoming_pots/function/decorated/as_player.mcfunction
```diff
@@ -0,0 +1,17 @@
...
+
+# Raycast for the pot (set up range and search)
+execute store result score @s gm4_blossoming_pots.range run attribute @s minecraft:block_interaction_range get 200
+execute anchored eyes run function gm4_blossoming_pots:decorated/raycast/loop
+
+execute as @n[type=minecraft:marker,tag=gm4_blossoming_pots.temp.decorated_pot] at @s \
+    align xyz positioned ~.5 ~.5 ~.5 run function gm4_blossoming_pots:decorated/as_marker with block ~ ~ ~ item
+
+# kill temp markers that failed to instantiate function (pot empty)
+kill @e[type=minecraft:marker,tag=gm4_blossoming_pots.temp.decorated_pot]
```
Bloo-dev, line 17: Consider moving the kill command into an intermediary function. First select the marker as noted above (and add a range limit of ~32), then check if the `item` tag is present. If so, call your makro, else `kill @s`.

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/as_marker.mcfunction
```diff
@@ -0,0 +1,46 @@
...
+# decrement player mainhand by 1
+$execute as @p[nbt={UUID:$(UUID)},gamemode=!creative] run item modify entity @s weapon.mainhand gm4_blossoming_pots:count_decr
+# increment perma marker count by 1
+    # this is a nightmare of selectors
+        # could tag it at the start of the function to have this be just checking that one tag?
+execute store result score @n[type=minecraft:marker,tag=gm4_blossoming_pots.data.flower_pot,distance=..0.2] gm4_blossoming_pots.loop \
+    run data get entity @n[type=minecraft:marker,tag=gm4_blossoming_pots.data.flower_pot,distance=..0.2] data.count 1
+scoreboard players add @n[type=minecraft:marker,tag=gm4_blossoming_pots.data.flower_pot,distance=..0.2] gm4_blossoming_pots.loop 1
+execute store result entity @n[type=minecraft:marker,tag=gm4_blossoming_pots.data.flower_pot,distance=..0.2] data.count int 1 \
+    run scoreboard players get @n[type=minecraft:marker,tag=gm4_blossoming_pots.data.flower_pot,distance=..0.2] gm4_blossoming_pots.loop
```
Bloo-dev, line 30: These should be moved into a function executed as `@n[type=minecraft:marker,tag=gm4_blossoming_pots.data.flower_pot,distance=..0.2]`, so you can replace the selectors with `@s`

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/vanilla_pots/as_marker.mcfunction
```diff
@@ -0,0 +1,27 @@
+# If player is placing a vanilla plant in custom flower pot, prevent them and give them the item back
+# @s = temp marker inside filled vanilla flower pot
+# at @s align xyz positioned ~.5 ~.5 ~.5
+# run from flower/vanilla_pots/as_player
+
+#check block location for tagged display or perma marker?
+execute \
+    unless entity @e[type=minecraft:block_display,tag=gm4_blossoming_pots.display.flower_pot,distance=..0.2] \
```
Bloo-dev, line 8: Add `limit=1`

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/vanilla_pots/as_marker.mcfunction
```diff
@@ -0,0 +1,27 @@
...
+
+fill ~ ~ ~ ~ ~ ~ minecraft:flower_pot destroy
+
+# execute as the flower pot item exactly where it summons, kill 
+execute as @n[type=minecraft:item,distance=..0.5,nbt={Item:{id:"minecraft:flower_pot"},Age:0s}] run kill @s
+
+# execute as nearest item with 0 time removed, if predicate matches item tag for small pot items
+    # store item data in temp, or marker lol
+    # kill it
+    # Use marker data (which has player UUID) to give player the item back, might need a macro function for that
```
Bloo-dev, line 21:
```suggestion
# execute as nearest item with 0 time removed, if predicate matches item tag for small pot items
# | store item data in temp, or marker lol
# | kill it
# | Use marker data (which has player UUID) to give player the item back, might need a macro function for that
```

gm4_blossoming_pots/data/gm4_blossoming_pots/function/main.mcfunction
```diff
@@ -0,0 +1,5 @@
+execute as @e[type=minecraft:block_display,tag=gm4_blossoming_pots.display.decorated_pot] at @s unless block ~ ~ ~ minecraft:decorated_pot run kill @s
+execute as @e[type=minecraft:block_display,tag=gm4_blossoming_pots.display.flower_pot] at @s unless block ~ ~ ~ minecraft:flower_pot run kill @s
+execute as @e[type=minecraft:marker,tag=gm4_blossoming_pots.data.flower_pot] at @s unless block ~ ~ ~ minecraft:flower_pot run function gm4_blossoming_pots:flower/item/break with entity @s data
```
Bloo-dev, line 3: Each flower pot marker has an associated block display. Maybe move the block display kill into a function executed by the marker.

gm4_blossoming_pots/data/gm4_blossoming_pots/function/vanilla_pot_handling/give_item_back.mcfunction
```diff
@@ -0,0 +1,8 @@
+# Gives player the vanilla flower pot item back that they tried to plant in a custom flower pot
+# @s = temp marker inside filled vanilla flower pot
+# at @s align xyz positioned ~.5 ~.5 ~.5
+# with {UUID, id}
+# run from vanilla_pot_handling/as_marker
+
+$execute at @p[nbt={UUID:$(UUID)},gamemode=!creative] run summon minecraft:item ~ ~ ~ {Item:{id:"$(id)"}}
+    # Note: Possibly, if we exit out of as_marker, this can run as the player and we dont need to select the UUID? We'd need to signal as_player to do this.
```
runcows, line 8: Just pushed a bunch of changes, wondering about this. 
I think it might be possible to get rid of this player selector by moving this function to be called in the end of as_player rather than as_marker. as_marker would need to send some signal to as_player to call this (maybe storage, maybe a score) but then this function could be executed as the player thats being selected with this UUID. Would this be worth it?

Bloo-dev: Doing NBT checks on players should be avoided at all costs, I can't even state how expensive they are.

If you need this UUID matching you should do `execute as $(UUID) if entity @s[gamemode=!creative] at @s`. This is of course assuming that you didn't obtain that UUID by reading player NBT.... if you did that this won't help.

runcows: I did in fact obtain the UUID by reading player NBT.... I will attempt my little shuffle idea to avoid this though

Bloo-dev: Maybe you could tag the player to mark them? If you need an ID system you can implement your own quite easily by assigning each player a score ID on first join (incrementing numbers).

runcows: Yeah, I could definitely tag the player on getting the advancement and then select them that way if I can't get around the selector. I did this in 2 other places so I'll see what I can do o7

runcows: Removed the player nbt UUID checking entirely.

runcows:
Pushed more changes. I think this covered everything we discussed in the review session :D

gm4_blossoming_pots/beet.yaml
```diff
@@ -0,0 +1,41 @@
...
+meta:
+  gm4:
+    versioning:
+      schedule_loops: [main]
+    website:
+      description: Decorated Pots and Flower Pots extended to allow all kinds of things to be placed in them! Decorate in Style!
+      #recommended:
+        #- 
+      notes:
+        - <add notes on version or incompatibilities?>
```
TheThanathor, line 28:
```suggestion
      description: Decorated Pots and Flower Pots extended to allow all kinds of things to be placed in them! Decorate in Style!
      recommended: []
      notes: []
```

gm4_blossoming_pots/beet.yaml
```diff
@@ -0,0 +1,41 @@
...
+    #modrinth: 
+      #project_id: null
+    #smithed:
+      #pack_id: null
+    planetminecraft: 
+      uid: null
+    video: null
+    wiki: null
```
TheThanathor, line 36: Link to wiki should be added even if the page does not exist yet
```suggestion
    wiki: https://wiki.gm4.co/wiki/Blossoming_Pots
```

gm4_blossoming_pots/data/gm4_blossoming_pots/function/decorated/play_sound.mcfunction
```diff
@@ -0,0 +1,7 @@
+# Literally just plays sound
+# @s = temp marker inside decorated pot
+# with {place_sound}
+# at @s align xyz positioned ~.5 ~.5 ~.5
+# run from decorated/loop_init
```
TheThanathor, line 5:
```suggestion
# run from decorated/prepare_data
```

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/as_player.mcfunction
```diff
@@ -0,0 +1,20 @@
...
+
+# Reset advancement
+advancement revoke @s only gm4_blossoming_pots:interact_with_flower_pot
+# Reset temp
+data remove storage gm4_blossoming_pots:flower_pots temp
+
+function gm4_blossoming_pots:flower/get_player_data
+# raycast
+execute store result score @s gm4_blossoming_pots.range run attribute @s minecraft:block_interaction_range get 200
+execute anchored eyes run function gm4_blossoming_pots:flower/raycast
```
TheThanathor, line 14: This doesn't matter much here, but keep in mind that just using anchored eyes does not move the execute location unless you use ^, which means the first loop of the raycast checks at the player's feet.
e.g. running
`/execute anchored eyes run particle minecraft:happy_villager ~ ~ ~` plays the particle at your feet, while
`/execute anchored eyes run particle minecraft:happy_villager ^ ^ ^` plays it at your eyes

gm4_blossoming_pots/data/gm4_blossoming_pots/function/main.mcfunction
```diff
@@ -0,0 +1,5 @@
+execute as @e[type=minecraft:block_display,tag=gm4_blossoming_pots.display.decorated_pot] at @s unless block ~ ~ ~ minecraft:decorated_pot run kill @s
+
+execute as @e[type=minecraft:marker,tag=gm4_blossoming_pots.data.flower_pot] at @s unless block ~ ~ ~ minecraft:flower_pot run function gm4_blossoming_pots:flower/cleanup with entity @s data
```
TheThanathor, line 3: Do we know if using macro's here is better than setting the items data from the entity?

TheThanathor (approved):
Code looks great! I've left a few comments, but it seems good to go.

I've also tested it in game and found only one very minor bug:
When a custom plant is in a flower pot you can plant a vanilla flower in there by holding right click and walking in circles around the flower pot (the module tries its best to prevent you but you can get a flower to slip through). This does fix itself by dropping the custom flower though so I don't think it's a problem.

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/as_player.mcfunction
runcows: I did not actually know that. So the first loop of the raycast could fail to find the right pot if the player is clipped inside one. I assume then this could be resolved by changing the line in as_player to `execute anchored eyes positioned ^ ^ ^ run gm4_blossoming_pots:flower/raycast`?

TheThanathor: technically yes, but with the other checks it doesn't seem to ever fail. I tried many things in game to get it to run on the wrong pot and couldn't manage it

gm4_blossoming_pots/data/gm4_blossoming_pots/function/main.mcfunction
runcows: I have no idea. My thought a long time ago was that it was better to summon the item with the data than to modify it after summoning. 
I think we would have to use 2 selectors to set the item data from the marker; once for the id, and again for the item count. Personally, I just don't know the performance comparison between 2 selectors and a macro

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/as_player.mcfunction
runcows: I think I'll do that change just in case there's some way it could happen

Bloo-dev (approved):

PR #1075, opened 2024-11-25
Liab 1.20.5: Fix all brewed splash->lingering becoming lightning

EpyonProjects:
Backdate fix for #1074


EpyonProjects:
ugh, I didn't update my local branch

misode (approved):
Thanks again for the fix!

PR #1076, opened 2024-11-28
Fix broken quotes in init.mcfunction

misode:
I believe this should fix it, but I don't remember how to build the modules in this version so hopefully github will do it for me


misode:
Okay it isn't configured for pull requests so let's just merge it

PR #1078, opened 2024-12-01
Metallurgy 1.21: Lore Line Shrink

EpyonProjects:
Currently, Metallurgy shamir lore takes up 3 lines
```
Name: Obsidian Cast
Lore: <Metal Type> Band
      Shamir
      <Shamir>
```
This update removes that standalone "Shamir" line to only use 2 lines
```
Name: Obsidian Cast
Lore: <Metal Type> Band
      <Shamir> Shamir
```
This change is part of the larger Metallurgy redesign.
Existing items will not have " Shamir" appended to the end, but will function the same.  This also means old unattached bands will not be stackable with shamirs created after this update.


gm4_metallurgy/data/gm4_metallurgy/function/smooshing/remove_band/clear_data.mcfunction
```diff
@@ -2,12 +2,30 @@
 # run from smooshing/remove_band/check
 
 execute if score custom_model_data gm4_ml_data matches 0 run data remove entity @s Item.components."minecraft:custom_model_data"
+execute if items entity @s contents *[custom_data~{gm4_metallurgy:{active_shamir:"infinitas"}}] run data remove entity @s Item.components."minecraft:max_stack_size"
```
misode, line 5: Why does infinitas need special handling? Maybe include a comment about this in the function

Bloo-dev:
This is now waiting on some 1.21.5 changes.

PR #1082, opened 2024-12-20
Optimize book binders NBT modifications

misode:
I noticed that the `main` function of book binders was not super performant, and this boiled down to some NBT modification commands. The module would repeatedly modify the `DisabledSlots` NBT, even if the disabled/enabled state of a book binder hadn't changed. This simply adds some checks for the `gm4_book_binder_active` tag.

Not yet tested in-game

Performance report below is on an idle server with 2 book binder entities, but still book binders taking > 3% of the `scheduledFunctions` seems too much too me.
image


EpyonProjects:
Functions as expected in 1.21.4

Bloo-dev (approved):

Issue #902, opened 2023-09-25
Add Environment Checks to `base`


Bloo-dev:
## The Issue ⚠️
Some modded environments interfere with the way data packs function without communicating this to the user.
As a result, we get plenty of false positives of our data packs being brocken whilst what is really broken is **those modded environments not supporting basic vanilla features**.

The number one contender for this is `paper`/`spigot` which prevents non-player entities from being added to teams but never communicates this change of a basic vanilla mechanic to the player.
What makes this even worse is that the config option is confusingly named `scoreboards.allow-non-player-entities-on-scoreboards=false` but only affects teams, not scoreboards.

We have reached out to the `paper`/`spigot` development teams multiple times but they do not seem to understand how much of an impact this default value of theirs has and have shown no signs of wanting to fix this issue.

## What We Can Do 🔨
We can't fix `spigot`/`paper`, nor could we convince the respective development teams to fix this issue. However, this issue is negatively impacting our users, primerly ones without in-depth technical knowledge, which then blame our data packs. We need to do _something_.

By **adding some environment checks** in **base** we could detect if certain mechanics work as expected and print errors in chat if not.
Below is a list of mechanics to check (WIP):
- Test if a non-player entity can be added to a team

PR #1093, opened 2025-01-28
Environment Checks

Bloo-dev:
This PR adresses #902.

Environment checks (echecks) are specified in the `beet.yaml` of a module, as a list under `versioning`. E.g.:
```yaml
meta:
  gm4:
    versioning:
      required:
        gm4_metallurgy: 1.5.0
        lib_player_death: 1.3.0
      echecks: [gm4:echeck/score_on_non_player_entity]
      schedule_loops: [main]
```

Upon load, each module requests its environment checks to run. In the `post_load` stage, `base` (`gm4`) then calls the function tag `#gm4:evaluate_echecks`. This function tag is used to ensure versioning works for echecks defined in base or libraries.

New echecks may be defined by any module or library. To do so, simply write your echeck in `.mcfunction` files, keeping in mind, that you must ensure proper caller context (`as` and `at`) yourself. All echeck results are cached during a load, so each echeck will only run once. You do not need to implement caching yourself.
Once the `.mcfunction` code for your echecks is ready, add the `#gm4:evaluate_echecks` function tag to your namespace and point it to your echeck. Done.
You may now include your echeck in a module, a library, or base by adding it to the corresponding `beet.yaml`.

This PR includes the `gm4:echeck/non_player_entity_has_score` and `gm4:echeck/non_player_entity_on_team` checks. Other checks will have to be written in the future.


gm4_animi_shamir/beet.yaml
```diff
@@ -20,6 +20,7 @@ meta:
       required:
         gm4_metallurgy: 1.4.0
         lib_player_death: 1.2.0
+      environment_checks: [gm4:score_on_non_player_entity]
```
SpecialBuilder32, line 23: What was your reasoning for making this configurable? Is there any reason we wouldn't want to just check the environment in every module, especially since they all use scoreboards to initialize through lantern load.

gm4/plugins/versioning.py
```diff
@@ -56,6 +57,26 @@ def modules(ctx: Context, opts: VersioningConfig):
         lines.append(f"execute if score {dep_id} load.status matches 1.. unless score {dep_id} load.status matches {dep_ver.major} run data modify storage gm4:log queue append value {log_data}")
         lines.append(f"execute if score {dep_id} load.status matches {dep_ver.major} unless score {dep_id}_minor load.status matches {dep_ver.minor}.. run data modify storage gm4:log queue append value {log_data}")
 
+    # add required environment checks
+    for namespaced_environment_check in opts.environment_checks:
```
SpecialBuilder32, line 61: I almost think we should only run environment checks on first install, instead of every reload? Also, we only need to perform each check once. Putting them into load makes each module run the same check again. Maybe we can finally utilize pre_load for this kind of setup check?

BPR02: That would miss the edge case of importing an existing world to a new server.

Bloo-dev: > Putting them into load makes each module run the same check again
That's why these checks leave a score behind which is only reset post load. If the check was already done this reload, it won't need to check again and just uses the last result. That result is invalidated in post load, making sure the check is run again once next reload.

gm4_animi_shamir/beet.yaml
Bloo-dev: That's mostly true for this check, but not for all. E.g. `lib_machines` need command blocks to be enabled, but most modules don't need that.

Bloo-dev:
I have cleaned up the base python code to the point where I'm satisfied with functionality.

Here's what's left to do:
- [ ] Decide: Should environment checks reside in `function/environment_check/` during development, or should we place them in `environment_check/` (same level as `function/`) and move all those files over at build time (may cause issues when someone names a folder in `function/` the same as the folder we move the checks to.... Hence I don't really see a point in doing this.
- [ ] Write the remaining checks. We can always add more checks later, I'd rather have this merged with some checks and add more later then have it sit around forever.

Bloo-dev:
After some more thinking we have come to the following solution, which splits `load.mcfunction` into 3 functions:
- `pre_load.mcfunction`: Called by lantern load instead of load, starts environment checks (these may take longer than 1 tick). Then calls `await_environment_checks.mcfunction`.
- `await_environment_check_results.mcfunction`: Checks for the results of the environment checks started by this module. Defers (self schedule 1t) further execution until all checks have returned a result. When this has happend, `load.mcfunction` is called.
- `load.mcfunction`: Prints out error messages depending on the result of the environment checks. Then does all the version compat checking. Runs `init` if nothing fails

docs/utilities.md
runcows, line 1: We'll want to update the tag lists in this file to reflect changes in #1264

base/data/gm4/function/log.mcfunction
```diff
@@ -4,6 +4,8 @@ execute if data storage gm4:log log{type:"install"} run tellraw @a[tag=gm4_show_
 execute if data storage gm4:log log{type:"missing"} run tellraw @a[tag=gm4_show_log] [{"nbt":"log.module","storage":"gm4:log","color":"red"},{"text":" is disabled because ","color":"red"},{"nbt":"log.require","storage":"gm4:log","color":"red"},{"text":" is not installed."}]
 execute if data storage gm4:log log{type:"outdated"} run function gm4:outdated_logs/outdated_start
 execute if data storage gm4:log log{type:"version_conflict"} run function gm4:conflict_logs/version_conflict_start
+execute if data storage gm4:log log.echeck_id if data storage gm4:log log.result{passed:-1} run tellraw @a[tag=gm4_show_log] [{"text":"-","color":"white"},{"nbt":"log.echeck_id","storage":"gm4:log","interpret":true, "color": "yellow"},{"text":" has timed out!","color":"white"}]
+execute if data storage gm4:log log.echeck_id if data storage gm4:log log.result{passed:0} run tellraw @a[tag=gm4_show_log] [{"text":"-","color":"white"},{"nbt":"log.echeck_id","storage":"gm4:log","interpret":true, "color": "yellow"},{"text":" has failed! ","color":"white"},{"nbt":"log.result.probable_cause","storage":"gm4:log","interpret":true,"color":"white"}]
```
runcows, line 8: As far as I understand, this means that the logs are parsing all environment check results. Which, while it is unnecessary loops of this function being run, is probably not a huge deal since we're unlikely to have so many environment checks running for it to be an issue. But it is something to be aware of

base/data/gm4/function/echeck/score_on_non_player_entity.mcfunction
```diff
@@ -0,0 +1,7 @@
+# Tests if non-player entities can hold scores. Default-config Spigot fails this test.
```
runcows: note to us: make sure this check actually does fail on spigot with the relevant config set to false :P

runcows: the paper config actually controls behaviour of teams. It is also an option for Paper specifically and not Spigot
From the [papermc documentation](https://docs.papermc.io/paper/reference/world-configuration/):
> `scoreboards.allow-non-player-entities-on-scoreboards`
> Instructs the server to always treat non-player entities as if they are never on a team. Disabling this may slightly decrease the amount of time the server spends calculating entity collisions

base/data/gm4/function/log.mcfunction
Bloo-dev: Correct. We could re-implement the log printing logic for echeck result printing, but that seemed excessive to me.

base/data/gm4/function/echeck/non_player_entity_has_score.mcfunction
```diff
@@ -0,0 +1,7 @@
+# Tests if non-player entities can hold scores. We think spigot might have done this at some point, but are not sure.
```
misode, line 1: After thinking about this again, we should probably just remove this check. I'll get annoying adding this to every module when it doesn't even matter in the current versions.

base/data/gm4/function/load.mcfunction
```diff
@@ -1,4 +1,6 @@
 data merge storage gm4:log {queue:[],versions:[]}
+data modify storage gm4:log echecks set value []
```
misode, line 2: This command can be combined into the above command

base/data/gm4/function/announce_echeck_results.mcfunction
```diff
@@ -0,0 +1,17 @@
+# Copies the results of environment checks into the log queue to prepare for printing to chat
+# @s = none
+# at unspecified
+# run from await_echeck_results
+
+# if no check failed, no need to print
+execute unless data storage gm4:log echecks[{result:{passed:0}}] unless data storage gm4:log echecks[{result:{passed:-1}}] run return 0
+
+# copy results into queue, KEEP VERSION IN echecks, SO WE CAN INSPECT IT FOR DEBUGGING
+data modify storage gm4:log queue set from storage gm4:log echecks
```
misode, line 10: This feels really weird and hacky. The NBT objects in `echecks` don't have the same shape as what is expected by `queue`. This could cause subtle bugs in the future.

This command also completely overwrites the `queue` field. That feels wrong and might cause problems. For example GM4 base log printing is already delayed to wait for at least one player. Environment checks could finish in between and cause the normal logs to be overwritten.

I would suggest adding a separate recursive function that iterates over the echecks list, or alternatively appending to the existing queue and making sure that it runs a second time if necessary.

gm4/plugins/versioning.py
```diff
@@ -47,33 +53,63 @@ def modules(ctx: Context, opts: VersioningConfig):
...
-        lines.append(f"execute if score {dep_id} load.status matches 1.. unless score {dep_id} load.status matches {dep_ver.major} run data modify storage gm4:log queue append value {log_data}")
-        lines.append(f"execute if score {dep_id} load.status matches {dep_ver.major} unless score {dep_id}_minor load.status matches {dep_ver.minor}.. run data modify storage gm4:log queue append value {log_data}")
-
-    # finalize startup check
+        log_message_commands.append(f"execute if score {dep_id} load.status matches 1.. unless score {dep_id} load.status matches {dep_ver.major} run data modify storage gm4:log queue append value {log_data}")
+        log_message_commands.append(f"execute if score {dep_id} load.status matches {dep_ver.major} unless score {dep_id}_minor load.status matches {dep_ver.minor}.. run data modify storage gm4:log queue append value {log_data}")
+
+    # add environment check requests
+    echeck_requests: List[str] = []
+    for namespaced_echeck in opts.echecks:
```
misode, line 67: Now that environment checks no longer block the module from loading, we should consider whether we really want these checks to be triggered from `load`, and not from `init`.

For example: shamir X has a very specific environment check Y. But shamir X requires that metallurgy is present. When it is not, shamir X will run `load` but not `init` because it is disabled. I don't think we want that very specific environment check to complain in that case because the module isn't active anyways.

This could also happen when two versions of a library are present and the one that is disabled has an environment check that the one that is active does not have.

misode (commented):
I'm not sure that I'm a big fan of using the term `echeck` in the code and file paths. I don't think `environment_check` is that long.

Right now the only example checks are pretty simple.
- What about for example a check that checks the correct minecraft version (or pack format)?
- Will we need separate environment check identifiers for each new version?
- Do we have to keep all the old version checks in the base?
- What happens when there are multiple GM4 base versions present, some of them having environment checks that the others don't have, would everything still work?

My worry is that with this much abstraction around the environment checks it will be difficult to add some of the more complex checks that require multiple ticks or rely on overlays/json resources.

gm4/plugins/versioning.py
Bloo-dev: That seems reasonable.

base/data/gm4/function/announce_echeck_results.mcfunction
Bloo-dev: I think this would require reimplementing the chat announcing logic, which doesn't sound appealing to me

PR #1097, opened 2025-02-11
Small changes to the Contribution doc

catter1:
I had a couple of questions myself when I was getting back into uv/beet after a few months, and didn't find the answers in the Contribution page. So, a couple of additions:

- Specified that the current Python version being used is 3.10
- Specified that `--link` takes a directory to the main world folder
- Added a "Common issues" tab, where it explains how to refresh the beet cache

The "Common issues" could use some more stuff, so suggestions are welcome. This is the only thing I've personally come about so far, but I'm sure there will be more.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -4,6 +4,7 @@ This guide will show how to setup a local development environment to build, test
 ## Table of contents
 * [Installation](#installation)
 * [Building modules](#building-modules)
+* [Common issues](#common-issues)
 * [Submitting changes](#submitting-changes)
 
 ## Installation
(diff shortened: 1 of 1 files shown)


Bloo-dev (approved):

docs/getting-started.md
```diff
@@ -40,7 +41,7 @@ uv run beet dev --link <world_name> --watch metallurgy *_shamir
 ```
 * `uv run` makes sure that the command uses the correct Python dependencies
 * `beet dev` builds the data packs to the `out` folder
-* `--link <world_name>` copies the packs to a local world
+* `--link <world_name>` copies the packs to a local world (filepath to the world's directory)
```
misode, line 44: Is it not confusing to call this a "file path"? It's only the name of the world folder, not the entire file path.

catter1: Hm, I was reading [the world linking code](https://github.com/mcbeet/beet/blob/main/beet/contrib/link.py#L84), and maybe I'm misunderstanding how `pathlib` works. Followed the functions, and I never saw where the Minecraft folder was defined, so I assumed the user had to specify it.

misode: I guess we should document the `MINECRAFT_PATH` env here, and also the `--minecraft` flag. The beet docs touch on this a tiny bit: https://mcbeet.dev/getting_started/#linking-to-a-world

catter1: Ah, I should look for docs before reading the code :) But thanks, that makes a lot more sense now.

PR #1099, opened 2025-02-12
Blossoming Pots Pre-GM4 Upgrade Path

runcows:
This PR for Blossoming Pots adds an upgrade path from the pre-gm4 version of Blossoming Pots.

In addition to the standard upgrade path run conditions, the upgrade path will also run on pot interaction with a distance within 10 blocks of the player, if the pre-gm4 version of the pack was detected.

For Decorated Pots
- legacy block displays are deleted 
- current version block displays are summoned based on the number of items in the pot
- for plants removed in the GM4 conversion (like coral blocks) the block displays simply do not generate

For Flower Pots
- legacy block displays are deleted
- marker storing item data has its count set to 1, if the count was higher than 1 the extra items are dropped on the ground on location
- for plants removed in the GM4 conversion (like coral blocks) the block displays do not generate, and all the items stored on marker are dropped on the ground on location

For Flower Pot conversion I ended up choosing to drop extra items on the ground as it is the easiest option to implement. I could see leaving extra count on the marker for new versions of plants, to use on clean up. However, for removed plants the item has to go somewhere so as to not delete player data outright. It is though debatable whether dropping the items on the ground achieves the same thing as deleting the items, should the player not notice. Unfortunately, there doesn't seem to be any perfect solution to this, so I think dropping the items is good enough, but if anyone has any other ideas lmk and I'd be willing to try it out


gm4_blossoming_pots/data/gm4_blossoming_pots/function/upgrade_paths/3.1/check_legacy_flower_pot_in_storage.mcfunction
```diff
@@ -0,0 +1,7 @@
+# Check if flower pot id is in storage
+# @s = type=minecraft:marker, tag=rc_blossoming_pots.flowerPot
+# at @s
+# with {id}
+# run from upgrade_paths/3.1/legacy_flower_pot
+
+$execute if data storage gm4_blossoming_pots:flower_pots $(id) run data modify storage gm4_blossoming_pots:flower_pots temp.legacy.in_storage set value 1
```
BPR02, line 7: Is there a reason this is setting a value in storage instead of a scoreboard? Seems the usage would work with a scoreboard value which is better for performance.

BPR02 (commented):
Looks good to me, but I left one comment that may slightly improve performance

gm4_blossoming_pots/data/gm4_blossoming_pots/function/upgrade_paths/3.1/check_legacy_flower_pot_in_storage.mcfunction
runcows: Honestly, it's because my scoreboards in this module are a confusing nightmare and could be massively simplified if i collapsed some things into a misc scoreboard, but I didn't want to touch it here. So instead of finding a place for a score i just put it in storage. 

With this I could do my misc scoreboard idea or just find a place for this as a score and save the scoreboard simplification for a later PR

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/as_player.mcfunction
```diff
@@ -14,16 +14,17 @@ execute if score blossoming_pots gm4_earliest_version matches 0 \
...
 data remove storage gm4_blossoming_pots:flower_pots temp
+scoreboard players set $signal_item_decr gm4_blossoming_pots.misc 0
 
 function gm4_blossoming_pots:flower/get_player_data
 # raycast
 execute store result score @s gm4_blossoming_pots.range run attribute @s minecraft:block_interaction_range get 200
 execute anchored eyes positioned ^ ^ ^ run function gm4_blossoming_pots:flower/raycast
 
 # if signaled, decrement player mainhand (set in flower/prepare_data)
-execute if data storage gm4_blossoming_pots:flower_pots temp.signal_item_decr if entity @s[gamemode=!creative] run item modify entity @s weapon.mainhand gm4_blossoming_pots:count_decr
```
BPR02, line 17: Is line 22 in `flower/prepare_data` still necessary? Seems the usage of the data storage is no longer needed due to this change

BPR02 (commented):
Looks good. Found one instance of an unnecessary line of leftover code.

gm4_blossoming_pots/data/gm4_blossoming_pots/function/flower/as_player.mcfunction
runcows: Ooops I forgot to change that over to the scoreboard version. Thanks for the spot

BPR02 (approved):

PR #1100, opened 2025-02-16
Add Forming Press Crafting Recipe

runcows:
This PR for Forming Press adds a vanilla recipe for the Forming Press itself as it was absent.
It also fixes an error in the guidebook recipe file for the Forming Press.


misode:
Ah right, I removed all the forming press recipes in #1053, but of course the recipe to craft the forming press itself could still stay in the vanilla crafting table.

runcows:
I think this is missing is an advancement to give the recipe itself, looking into how that's done in other modules now

runcows:
Alright, this now has an advancement to give the player the recipe. Tested and working

runcows:
Weird inconsitency I noticed. In forming press on the main repository, right now there are 2 different recipes. There is one defined in `gm4_forming_press/function/check_recipes` for the custom crafter. This is the recipe listed on the wiki. But there is a second recipe defined in `gm4_forming_press/gm4_recipes`, which im not actually sure this folder is used for, but it has a repeater instead of a comparator in it. This is the version I based the json recipe on, which might be incorrect. 

What is the `gm4_recipes` folder used for?

misode:
`gm4_recipes` is used by the guidebook module to generate recipe previews. It's not used for actual crafting logic.

runcows:
Updated this to fix the json recipe and the guidebook recipe to both use the comparator. Would it be recommended to split the guidebook recipe fix into a separate PR as a fix?

Bloo-dev (changes requested):
Missing NL @ EoF

Bloo-dev (approved):

Issue #1037, opened 2024-09-08
Add Opt-In Logging to Modules Usable as Griefing Tools


Bloo-dev:
Servers often require actions of users to be trackable for moderation puposes, this is especially true if those actions have destructive consequences.
Some of our modules add rather destructive features which are not easily tracked by classical means. To countact this, we should add a storage-based logging system which can be enabled by server administrators.

It is to be decided whether this should be one unified Gamemode 4 log, or whether each action should have its own log (e.g. a log which tracks only uses of crossbow cartridges).

Tracked data should include:
- Time of use
- Location of use
- Name of the player responsible

Affected Modules:
- Crossbow Cartridges (primarily TNT launching)
- TNT Landmines

PR #1104, opened 2025-03-08
Reeling Rods Module

runcows:
Reeling Rods adds new fishing rod interactions to various entities that fishing bobbers can hook onto.
- Pull Paintings off walls
- Reel in items from Item Frames, and then the item frame off the wall
- Snatch leashed mobs from leash knots
- Try and fail to pull in an end crystal (it explodes)

Adds a new enchantment for fishing rods which allows you to
- Separate chest boats and all minecart variants
- Steal items from allays

This would be the first gm4 custom enchantment using the vanilla system. It is intentionally not a metallurgy shamir.

Functionally, the enchantment allows additional pulling options without entirely preventing the vanilla action of pulling the entity towards the player. All of the interactions possible without the enchantment, in vanilla simply do nothing but use durability. 

Additional ideas for this that I am unsure of and want feedback on are:
- Stealing armor or hand items from entities that can have them. This presents a gameplay balance issue that I am not knowledgable enough to have any insight on. It also could just be annoying for players in multiplayer.
- Dismounting mobs riding others. This would be added for parity with Bedrock Edition. As according to the [Minecraft Wiki](https://minecraft.wiki/w/Fishing_Rod#Usage), "in Bedrock Edition, the hooked mob dismounts".
- Pulling shulkers over a block towards the player. This would need checks to see if the shulker could be supported after being moved. Currently, in vanilla shulkers do nothing on being fished.


runcows:
Build for 1.21.1 is currently failing due to the boat entity type changes made in 1.21.2 and the #minecraft:boat entity type tag not existing then.

runcows:
Discussion from meeting:
- Need to add a not adventure mode check to advancements
- Additional ideas (Probably ok to implement previous ones)
  - \+ Steal potions from witches
  - \+ Steal items from villagers hands? Needs some penalty
  - Armor from mobs maybe shouldn't be a guaranteed drop
  - If additional ideas implemented, using enchantment makes sense since its a different utility / ability from the base rod + balance issues
- Theft Moderation could be an issue with servers (Might be a good time to look into #1037 )
- Enchantment could be exclusive to mending, so your yoinking days are limited
- Probably more discussion needed for gameplay balance of mechanics

runcows:
Added Barbed
- Mooshroom -> Cow Reeling + Barbed damage visual bug
- Enchantment stats data should be revisited
- Something Else I forgot

Bloo-dev:
Reeling and Barbed are now mutually exclusive

runcows:
Added drop chance usage to reeling, including an entity type tag for entities that should ignore drop chances.
When the drop chance fails, the equipment item in question just breaks.

Currently, the default drop chance is set to 8.5%, matching vanilla's base drop chance without looting.
Personally, I believe it should be more powerful than that, but I'm setting it here as a start to see what others think.

Bloo-dev:
> Added drop chance usage to reeling, including an entity type tag for entities that should ignore drop chances. When the drop chance fails, the equipment item in question just breaks.
> 
> Currently, the default drop chance is set to 8.5%, matching vanilla's base drop chance without looting. Personally, I believe it should be more powerful than that, but I'm setting it here as a start to see what others think.

Seeing that the player has to put in some effort to get this enchant, I think for armor and tools on most entities reeling success should be almost guaranteed; with the drop chance mode being the exception for things like tridents on drowned.

runcows:
> Seeing that the player has to put in some effort to get this enchant, I think for armor and tools on most entities reeling success should be almost guaranteed; with the drop chance mode being the exception for things like tridents on drowned.

Revisiting this after a few days, I definitely agree.
How does this sound then?
- If a slot has a 0% drop chance, it should fail.
  - There are exceptions to that, where the 0% should be ignored (villager and vex mainhand)
- If a slot has a non-zero drop chance, drop should be guaranteed.
  - There are exceptions to that, where a chance is checked (trident drowned)
### Remaining Questions
- Are there any other chances that should be checked, other than tridents on drowned?
- Is 8.5% a good drop chance for tridents off of drowned? It matches non-looting chances, but we can easily configure it to have a higher chance.
- For a 0% drop chance, is the fail behaviour breaking the slot or should it just not do anything?

Bloo-dev:
>     * If a slot has a 0% drop chance, it should fail.
>       
>       * There are exceptions to that, where the 0% should be ignored (villager and vex mainhand)
> 
>     * If a slot has a non-zero drop chance, drop should be guaranteed.
>       
>       * There are exceptions to that, where a chance is checked (trident drowned)

Agreed.

> ### Remaining Questions
> 
>     * Are there any other chances that should be checked, other than tridents on drowned?
> 
>     * Is 8.5% a good drop chance for tridents off of drowned? It matches non-looting chances, but we can easily configure it to have a higher chance.
> 
>     * For a 0% drop chance, is the fail behaviour breaking the slot or should it just not do anything?

- I can't think of any right now.
- 8.5% sounds good to me. I wouldn't be opposed to making it the looting III chance. When a chanced reel fails, it should break the equipment it was trying to reel.
- 0% drop chance equipment should just be ignored. You could play a sound akin to the hook "slipping off" the piece.

runcows:
Implemented the drop chance rework as discussed above.
- Currently drowned holding tridents are the only thing that will roll the random chance. Unless the drop chance is changed by something else, it will default to 11.5%, which is the same as looting 3.
- 0% drop chance now plays a sound when fails instead of breaking the item

Also updated the module to work with 1.21.6

gm4_reeling_rods/data/gm4_reeling_rods/function/id/loop.mcfunction
```diff
@@ -0,0 +1,9 @@
+# checks for bobbers and stops when there are none
+# run from player/cast_line \
+  & scheduled id/select_entities
```
Bloo-dev: Multi-line expressions in header comments are not allowed.
```suggestion
# run from player/cast_line and scheduled from id/select_entities
```

gm4_reeling_rods/data/gm4_reeling_rods/function/tick.mcfunction
```diff
@@ -0,0 +1,5 @@
+# checks if players have cast a line, then will assign ids to entities around bobber
+execute as @a[scores={gm4_reeling_rods.rods_cast=1..}] at @s run function gm4_reeling_rods:player/cast_line
```
Bloo-dev: I don't think the `at @s` is required here.
```suggestion
execute as @a[scores={gm4_reeling_rods.rods_cast=1..}] run function gm4_reeling_rods:player/cast_line
```

gm4_reeling_rods/data/gm4_reeling_rods/function/id/get_next.mcfunction
```diff
@@ -0,0 +1,55 @@
+# Gets the next id
+# run from id/set
```
Bloo-dev: Only the first line of the header comment (human-readable description) is optional. Executioner, location, and parent are required.

gm4_reeling_rods/data/gm4_reeling_rods/function/player/received_bit.mcfunction
```diff
@@ -0,0 +1,17 @@
+# Logic for when a player receives a bit
+# @s = player who received a bit
+# at @s
+# with {bit, UUID}
+# run from player/bit_{bit}_{value}
```
Bloo-dev, line 5: Whilst our header comment standard does not explicitly state how to handle macro information yet, we have been appending it to the parent function line, as a separate (optional) line seems a bit excessive. In the future we should manifest this in the standard.
You are right about only stating the keys, not the values, of the macro information. Make sure you state all keys supplied by the macro.

```suggestion
# run from player/bit_{bit}_{value} with {bit, UUID}
```

gm4_reeling_rods/data/gm4_reeling_rods/function/player/find_hooked_entity.mcfunction
```diff
@@ -0,0 +1,27 @@
+# Finds the entity the player hooked
+# @s = player who fished
+# at @s
+# with {bit_0, bit_1, bit_2, bit_3, bit_4, bit_5, bit_6, bit_7, bit_8, bit_9, bit_10, bit_11, bit_12, bit_13, bit_14, bit_15}
+# run from player/received_bit
```
Bloo-dev, line 5: See my comment on header comments in `received_bit.mcfunction`.

gm4_reeling_rods/data/gm4_reeling_rods/function/get_motion_to_player.mcfunction
```diff
@@ -0,0 +1,42 @@
+# Gets motion vector to launch item to player
+# run from pull_items
```
Bloo-dev: Incomplete header comment.

You should also use this header comment to explain what your motion algorithm does, it is pretty cumbersome to reconstruct from the mcfunction code and lookup table.

gm4_reeling_rods/generate_files.py
```diff
@@ -0,0 +1,119 @@
...
+                    continue
+                if potenital_key in lookup_keys:    # ignore if already found
+                    continue
+                lookup_keys.append(potenital_key)
+    lookup_keys.sort()
+    strList: List[str] = []
+    for key in lookup_keys:
+        value = math.floor(100*(0.08*math.sqrt(math.sqrt(key))))
+        strList.append(f"scoreboard players set ${key} gm4_reeling_rods.lookup {value}")
+    ctx.data["gm4_reeling_rods:set_lookup_table"] = Function(strList)
```
Bloo-dev: This is a very inefficient way of doing this. Lookups in lists scale with O(n) and sorts are even worse. If you want to stick with this system, the following one-liner should be functionally identical to your version's body (test it, I did not):
```suggestion
    ctx.data["gm4_reeling_rods:set_lookup_table"] = Function(
        [
            f"scoreboard players set ${key} gm4_reeling_rods.lookup {value}"
            for key, value in {
                (x**2 + y**2 + z**2): int(
                    100 * (0.08 * math.sqrt(math.sqrt(x**2 + y**2 + z**2)))
                )
                for x, y, z in product(range(0, 34), range(0, 34), range(0, 34))
            }.items()
        ]
    )
```

gm4_reeling_rods/data/gm4_reeling_rods/function/hooked_entity/item_frame.mcfunction
```diff
@@ -0,0 +1,19 @@
+# Action for hooked item frame or glow item frame
+# @s = item frame or glow item frame
+# at bobber in @s
+# with {type}
```
Bloo-dev, line 4: See my comment on header comments in `received_bit.mcfunction`. I will no longer highlight these from now on.

gm4_reeling_rods/data/gm4_reeling_rods/function/reeling/minecart/action.mcfunction
```diff
@@ -0,0 +1,21 @@
...
+$data modify storage gm4_reeling_rods:temp item_data.Item set value {id:"$(block)",count:1}
+function gm4_reeling_rods:pull_items
+
+# replace with minecart
+data modify storage gm4_reeling_rods:temp entity_data set from entity @s
+data remove storage gm4_reeling_rods:temp entity_data.UUID
+data remove storage gm4_reeling_rods:temp entity_data.Passengers
+data modify storage gm4_reeling_rods:temp entity_type set value "minecraft:minecart"
+execute at @s run function gm4_reeling_rods:reeling/summon_entity with storage gm4_reeling_rods:temp
+execute at @s on passengers run function gm4_reeling_rods:reeling/minecart/passenger_transfer
```
Bloo-dev: When does this take effect? You can't ride a chest/furnace/tnt minecart.

gm4_reeling_rods/data/gm4_reeling_rods/function/reeling/mooshroom/action.mcfunction
```diff
@@ -0,0 +1,25 @@
...
+data modify storage gm4_reeling_rods:temp item_data.Item.count set value 5
+function gm4_reeling_rods:pull_items
+
+# replace with cow
+data modify storage gm4_reeling_rods:temp entity_data set from entity @s
+data remove storage gm4_reeling_rods:temp entity_data.UUID
+data remove storage gm4_reeling_rods:temp entity_data.Passengers
+data modify storage gm4_reeling_rods:temp entity_type set value "minecraft:cow"
+execute at @s run function gm4_reeling_rods:reeling/summon_entity with storage gm4_reeling_rods:temp
+execute at @s on passengers run function gm4_reeling_rods:reeling/mooshroom/passenger_transfer
```
Bloo-dev: When does this take effect? You can't ride a cow. If this is for compatibility with custom entities that have been mounted using commands, those should probably no be modified in the first place and be excluded by checking for the smithed entity tag (I think you already do that).

gm4_reeling_rods/data/gm4_reeling_rods/function/reeling/stealable/steal_equipment.mcfunction
```diff
@@ -0,0 +1,13 @@
...
+# run from hooked_entity/select_type & reeling/villager/action
+
+execute store result score $order gm4_reeling_rods.math run random value 1..6
+
+execute if score $order gm4_reeling_rods.math matches 1 run return run function gm4_reeling_rods:reeling/stealable/order/equipment_1
+execute if score $order gm4_reeling_rods.math matches 2 run return run function gm4_reeling_rods:reeling/stealable/order/equipment_2
+execute if score $order gm4_reeling_rods.math matches 3 run return run function gm4_reeling_rods:reeling/stealable/order/equipment_3
+execute if score $order gm4_reeling_rods.math matches 4 run return run function gm4_reeling_rods:reeling/stealable/order/equipment_4
+execute if score $order gm4_reeling_rods.math matches 5 run return run function gm4_reeling_rods:reeling/stealable/order/equipment_5
+execute if score $order gm4_reeling_rods.math matches 6 run return run function gm4_reeling_rods:reeling/stealable/order/equipment_6
```
Bloo-dev, line 13: Instead of having 6 functions with the order written out starting at a different phase, couldn't you just determine a start value and then increment it 6 times, calling the appropriate steal function (1 -> mainhand, 2 -> head, 3 -> legs,...) on each increment. You'd of course have to take the modulo after every increment.

gm4_reeling_rods/data/gm4_reeling_rods/function/reeling/stealable/steal_hand.mcfunction
```diff
@@ -0,0 +1,9 @@
+# Pick an order to check hand slots for stealing
+# @s = #gm4_reeling_rods:steal_hand
+# at bobber in @s
+# run from hooked_entity/select_type
+
+execute store result score $order gm4_reeling_rods.math run random value 1..2
+
+execute if score $order gm4_reeling_rods.math matches 1 run return run function gm4_reeling_rods:reeling/stealable/order/hands_1
+execute if score $order gm4_reeling_rods.math matches 2 run return run function gm4_reeling_rods:reeling/stealable/order/hands_2
```
Bloo-dev: See comment in `steal_equipment.mcfunction`.

gm4_reeling_rods/data/gm4_reeling_rods/guidebook/reeling_rods.json
```diff
@@ -0,0 +1,110 @@
+{
```
Bloo-dev, line 1: NOTE for future us: I did not check this file.

gm4_reeling_rods/generate_files.py
```diff
@@ -0,0 +1,119 @@
+from typing import List
+from beet import Context, Advancement, Function
+import math
+from pathlib import Path
+from gm4.utils import CSV, CSVRow
+
```
Bloo-dev, line 6: Import needed for suggestion below.
```suggestion
from itertools import product

```

Bloo-dev (changes requested):
Did a full (excluding Guidebook and the folder `barbed/`) code review.

Nothing that's wrong, but found some nitpicks.
You should go over the header comments again and make sure they match the spec outlined in our docs (`docs/` folder)

gm4_reeling_rods/data/gm4_reeling_rods/function/tick.mcfunction
runcows: oh yep, that's a remnant from older code

gm4_reeling_rods/data/gm4_reeling_rods/function/player/received_bit.mcfunction
runcows: Talked in DMs, agreed to shelve the topic of macros in function headers for a future discussion.

Quick summary: There is no consistent standard for macro information in function headers currently. Most modules just don't include anything. A separate line in the style I've done here is what I've been doing since the Blossoming Pots code review. There are a few other styles used in the repo. We should revisit this in a future discussion on code standards where we can standardize everything properly. Ideally headers should remain machine parsable.

I will leave this conversation unresolved, so as to not hide it.

gm4_reeling_rods/generate_files.py
runcows: This is not functionally identical. It produces more than double the commands (1,131 more) because it's including values outside of the fishing bobber range (this is what line 73-74 in my version accounts for). 

I understand the desire for efficiency with this, but realistically since it's only run during compile time we probably don't need to obsess about it too much. There is definitely optimizations that can be made though while retaining the same function output.

gm4_reeling_rods/data/gm4_reeling_rods/function/reeling/mooshroom/action.mcfunction
runcows: Responding to this thread and not the minecart one since both are about the same thing.
Yes ,the idea behind transferring the passengers to the new entity was for compatibility with entities mounted with commands. Otherwise the passenger just gets yeeted into the void with the old entity.

Although, you make a very good point that entities like that probably shouldn't be modified in the first place. I'm thinking these should be changed to just fail if there is passengers.

Edit: I've implemented this now. Resolving this.

gm4_reeling_rods/generate_files.py
runcows: With the addition of a conditional to the suggested code, it is now functionally identical

gm4_reeling_rods/data/gm4_reeling_rods/function/reeling/stealable/steal_equipment.mcfunction
runcows: That would probably be easier to read and maybe easier to maintain, but I think that would result in a lot more commands in the worst case scenario. 

The worst case scenario (ignoring the steal_slot functions since they're the same for this) for each option is that the reeled entity has no item to steal, but we still need to check each slot, and then you random value a number that necessitates the max number of checks. 

For my method with the orders, the max number of commands follows a random value 6 which needs 6 score checks to get to that order, and then 6 slot checks then return fail. 

For a random starting point with a modulo, i think you need to random value and then 6*(check all 6 scores for the current one, increment, modulo). At least if I'm understanding your suggestion correctly

gm4_reeling_rods/data/gm4_reeling_rods/function/reeling/stealable/steal_hand.mcfunction
runcows: This one can probably be simplified to 3 commands (check mainhand, check offhand, check mainhand) with a random boolean determining if the first command should be skipped.

Edit: I've done this

gm4_reeling_rods/data/gm4_reeling_rods/function/reeling/stealable/steal_equipment.mcfunction
Bloo-dev: That makes sense. Let's stick with the current implementation that uses a bit more disk space, but disk space is basically free these days.

Bloo-dev (approved):

runcows:
A couple notes pulled from `generate_files.py` to clean it up:
- Stealing is treated as an action for reeling, and the entities that can be stolen from are listed in type tags. To give one of these entities an alternate action, remove them from the relevant `steal_*` entity type tag and implement the stealing following the example of how villagers are set up.
- Villager stealing: Currently, armor theft does not cause reputational harm to the player. However, this is easy to change.
- Illager types: can only have hand items stolen, despite being able to wear armor through commands. This is because armor does not render on them, nor is there a way to put armor on them in survival like in the case of villagers.
  - A consequence of this is that pillager captain banners cannot be stolen
- Entities rejected from special interactions
  - Enderman: Wanted to steal held block, but it is stored as a block state, not as item data, making it difficult to steal the item. 
  - Pufferfish: Wanted to puff them up, but setting PuffState has issues. Set once is fine. Then after it deflates a bit, setting again flashes and then reverts, having a jarring visual effect. This is probably an MC bug.

runcows:
As a last minute addition, I added a reeling interaction for enderman. You can steal the enderman's held block as a falling block entity that gets launched at you

PR #1111, opened 2025-04-17
Fix balloon animal rarity weighting

runcows:
Noticed that the loot table for whether to choose a rare or common balloon animal had the weights set up in a way that didn't make much sense.
Rare was weighted as 7, Common was weighted as 3.
This PR swaps this so the names common and rare align with the chances.

### DIFF
--- a/gm4_balloon_animals/data/gm4_balloon_animals/loot_table/technical/random/enumeration_value.json
+++ b/gm4_balloon_animals/data/gm4_balloon_animals/loot_table/technical/random/enumeration_value.json
@@ -5,12 +5,12 @@
       "entries": [
         {
           "type": "minecraft:loot_table",
-          "weight": 3,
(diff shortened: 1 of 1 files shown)


EpyonProjects:
So, this has been brought up before.  I should probably document it better.

This is already accurate.  The meaning of common and rare is the type of animal, but the goal of Balloon Animals (other than to showcase cool balloons you don't get to keep as balloons) is to give the player easier access to the more rare animals.  The 3-7 distribution for common-rare tries to weigh decently towards those harder to find animals.

runcows:
Ah that makes a lot of sense, thank you.

> So, this has been brought up before. I should probably document it better.
Do you mind if I make a suggestion for something that could clarify it?

runcows:
I'm going to close this PR now, since it's not actually suggesting a real change anymore

PR #1113, opened 2025-04-18
Fix Slab uncrafting overlapping with some Vanilla Recipes

### LINKED ISSUE #1105: Slab uncrafting recipes overlaps with some vanilla recipes
Suggested fix: change the recipe to a 2x2

runcows:
This PR is focused on Standard Crafting's slab and stair uncrafting
- Fixes #1105 by changing slab uncrafting to 2x2 slabs -> 2 blocks
- Updates custom crafter recipes to match
- Adding custom crafter support for 1.21 - 1.21.4 blocks
- Changes to generated function definitions and loot tables for the custom crafter


gm4_standard_crafting/generate_recipes.py
```diff
@@ -129,8 +127,67 @@ def recursive_apply(items: list[str], dir: str, shape: list[str], output_count:
...
+                        ]
+                    }
+                ]
+            })
+
+            command: str = "execute if score $crafted gm4_crafting matches 0 store success score $crafted gm4_crafting if data storage gm4_custom_crafters:temp/crafter {Items:["
+            slot = 0
+            while slot <= 4:
+                command += "{Slot:" + str(slot) + "b,id:\"" + item + "\"}"
+                if slot == 4:
```
misode, line 164: I'm hesitant to merge this because this is some ugly looking code if I'm being honest. Even though this PR removes more lines than it adds, it makes this module significantly harder to upgrade in the future.

runcows: The segment you highlighted is yeah, its not great. I was struggling with how to do the function definitions for each slot without just writing out the whole command, formatted strings weren't being nice either.
What would you recommend I do here?
Would it be better to roll back on the generated function definitions, but keep the rest?

misode: I'd probably duplicate the 4 slots. But also the whole "since_61" is not ideal here and `function_map` being passed to a function and getting mutated... Can the mcfunctions not be generated inside the `recursive_apply` function? Also what kind of special syntax is `"command":"##stairs"` even?

*Side note: I'm of the opinion that we should discontinue the custom crafter, and modify the remaining 4ish recipes that currently include NBT, so that's maybe why I don't want to spend too much time on custom crafter code*

runcows: I'm in agreement on discontinuing the custom crafter, but I don't think we should leave it behind in changes to standard crafting, which seems like what's been happening since the custom crafter currently doesn't have 1.21+ blocks. 

I just pushed a commit hopefully improving on the points you made, if I understood them correctly.
The mcfunctions can't be initialized inside `recursive_apply`, but I'm now appending directly to the functions rather than using `function_map`, which is indeed a lot simpler.

misode (approved):

Bloo-dev:
One more thought I had in hindsight:
Going with a 3x2 of slabs would be nicer, as slabs craft in lots of 6, not 4. If that change is too lengthy to make we can merge as-is and change the recipe again later.

runcows:
Ideally we only change the recipe just this once. For that matter though, the recipe could also be 2x1 (sideways instead of vertical) which would retain the same ratio as before

misode:
I think keeping it 2x2, consistent with stairs uncrafting, is acceptable

Bloo-dev:
After some more discussion, I agree.

PR #1114, opened 2025-04-23
Fix Gemini Shamir extra wolf baby not matching variant of other baby / parents

runcows:
Adds a command to set the variant of the extra baby from the first baby in `baby/spawn_wolf.mcfunction`


misode (approved):
Looks fine

PR #1122, opened 2025-05-10
1.21.5 heart canisters update

TheThanathor:
Heart Canisters no longer equip to the head slot and stack to a max of 5.

Tier 2 Canisters now work regardless of the amount of Tier 1 Canisters in the inventory

Heart Canisters now work even when spread around the inventory

### DIFF
--- a/gm4_heart_canisters/data/gm4_heart_canisters/function/check_heart_canisters.mcfunction
+++ b/gm4_heart_canisters/data/gm4_heart_canisters/function/check_heart_canisters.mcfunction
@@ -1,13 +1,21 @@
-# gets called for every player with at least 1 heart canister tier 1 item
+# gets called for every player with at least 1 heart canister item
 
-scoreboard players set @s[nbt={Inventory:[{count:1,components:{"minecraft:custom_data":{gm4_heart_canister_tier:1b}}}]}] gm4_heart_can 1
-scoreboard players set @s[nbt={Inventory:[{count:2,components:{"minecraft:custom_data":{gm4_heart_canister_tier:1b}}}]}] gm4_heart_can 2
-scoreboard players set @s[nbt={Inventory:[{count:3,components:{"minecraft:custom_data":{gm4_heart_canister_tier:1b}}}]}] gm4_heart_can 3
-scoreboard players set @s[nbt={Inventory:[{count:4,components:{"minecraft:custom_data":{gm4_heart_canister_tier:1b}}}]}] gm4_heart_can 4
-scoreboard players set @s[nbt={Inventory:[{count:5,components:{"minecraft:custom_data":{gm4_heart_canister_tier:1b}}}]}] gm4_heart_can 5
+# tier 1
+scoreboard players set $count_total gm4_heart_can 0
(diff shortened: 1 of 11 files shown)


gm4_heart_canisters/data/gm4_heart_canisters/function/check_heart_canisters.mcfunction
```diff
@@ -1,13 +1,21 @@
...
-scoreboard players set @s[nbt={Inventory:[{count:5,components:{"minecraft:custom_data":{gm4_heart_canister_tier:1b}}}]}] gm4_heart_can 5
+# tier 1
+scoreboard players set $count_total gm4_heart_can 0
+data modify storage gm4_heart_cannisters:temp heart_cannister_data set value []
+data modify storage gm4_heart_cannisters:temp heart_cannister_data append from entity @s Inventory[{components:{"minecraft:custom_data":{gm4_heart_canister_tier:1b}}}]
+execute store result score $count_add gm4_heart_can run data get entity @s equipment.offhand{components:{"minecraft:custom_data":{gm4_heart_canister_tier:1b}}}.count
+scoreboard players operation $count_total gm4_heart_can += $count_add gm4_heart_can
+function gm4_heart_canisters:count_heart_canisters
+scoreboard players operation $count_total gm4_heart_can < $heart_cannister_cap gm4_heart_can
+scoreboard players operation @s gm4_heart_can += $count_total gm4_heart_can
```
misode, line 11: This is an abysmal amount of player NBT checks, I'd much rather see 10 `if items` checks in a single mcfunction. And this score is no longer capped at `5`. I would either pick one of these changes:
- heart canisters stack to max 5
- heart canisers can now be spread around in the inventory

If you implement the latter, there is no longer any advantage of making them only stack to 5 imo.

misode (changes requested):

gm4_heart_canisters/data/gm4_heart_canisters/function/check_heart_canisters.mcfunction
TheThanathor: Unless I'm misunderstanding how player checks work this shouldn't be more than before, since the data is stored in storage first. I could change it to 20 `if items`. Making them stack to 5 feels like a good way to show players that that is the max amount that work

TheThanathor: Altough I guess `execute store result score $count gm4_heart_can if items entity @s container.* <heart can>` would work to count them as well, if that's better than looping over a storage

misode: Okay yeah, much better. Although the test is still failing, is the cap working?

TheThanathor: Test was failing since it assumed 6 canisters resulted in no health gain, while it now grants the effect of 5 canisters, I've updated the test

misode (approved):

PR #1127, opened 2025-05-10
1.21.5 Heart Canisters (& Crossbow Cartridges)

TheThanathor:
- changes Heart Canisters to use poisonous potato's instead of player heads
- fixes Crossbow Cartridges losing the "type" data of shot salmon


misode:
Will this not cause issues for people not using the resource pack?

BPR02:
> Will this not cause issues for people not using the resource pack?

`item_model` allows us to set the item to use the player head model, so it displays as a player head without the resource pack.

TheThanathor:
Still needs an upgrade path to transform old heart canisters

misode:
What was the problem with player heads?

TheThanathor:
it mostly came from trying to make them not equipable on the head slot, which broke when the canisters were placed (as it's done through not having a component, which can't be copied by the loot table). Then we just wanted to see if using poisonous potato's work, so this is mostly a test. Added benefit is that they are no longer placeable

PR #1054, opened 2024-09-25
Orb of Ankou 1.21 features

BluePsychoRanger:
### Soul Essence for New 1.21 Mobs
Armadillo:
- Enduring
- Retreating

Bogged:
- Hawkeye
- Lifeless
- Tranquilizing

Breeze:
- Agile
- Bounding
- (Breeze may get a new pneuma in the future, but I'd like to rework the system and would rather the feature update be implemented in the meantime)

### New Pneuma
Tranquilizing: makes all arrows shot from a bow into poison arrows

### Changed Pneuma
Freezing: makes all arrows shot from a bow into slowness arrows

### Optimized Nether Checks
There was a bug that caused the nether clocks to continuously run even when there were no players in the nether (this was due to players keeping the in_nether tag if they died)

Issue #1010, opened 2024-07-22
1.20.5 and 1.21 Feature Compatibility Updates


Bloo-dev:
This issue exists to keep track of _feature updates_ still needed to provide _feature compatibility_ with 1.20.6 and 1.21 features, e.g. support for new mobs or blocks.

Libraries:

- Gamemode 4 Base (`base`):
  - [x] Add new blocks to `full_collision` #1039
    - `chiseled_copper`
    - `exposed_chiseled_copper`
    - `weathered_chiseled_copper`
    - `oxidized_chiseled_copper`
    - `waxed_chiseled_copper`
    - `waxed_exposed_chiseled_copper`
    - `waxed_weathered_chiseled_copper`
    - `waxed_oxidized_chiseled_copper`
    - `chiseled_tuff`
    - `copper_bulb`
    - `exposed_copper_bulb`
    - `weathered_copper_bulb`
    - `oxidized_copper_bulb`
    - `waxed_copper_bulb`
    - `waxed_exposed_copper_bulb`
    - `waxed_weathered_copper_bulb`
    - `waxed_oxidized_copper_bulb`
    - `copper_grate`
    - `exposed_copper_grate`
    - `weathered_copper_grate`
    - `oxidized_copper_grate`
    - `waxed_copper_grate`
    - `waxed_exposed_copper_grate`
    - `waxed_weathered_copper_grate`
    - `waxed_oxidized_copper_grate`
    - `crafter`
    - `trial_spawner`
    - `vault`
    - `polished_tuff`
    - `tuff_bricks`
- Auto Crafting:
  - [x] Confirm 1.20.5 & 1.21 vanilla recipes are included
- Balloon Animals:
  - [x] Add `armadillo` to animal pool #1039
  - [x] Add `wolf` variants (otherwise all wolfs are pale wolfs) #1040
- Crossbow Cartridges:
  - [x] Add `wind_charge` as valid projectile item (#1039)
    - ~~Discuss: Fire Charges are a valid projectile, however, those can't be launched by hand~~ 
- Dangerous Dungeons:
  - [ ] Discuss switching to `trial_spawner`
  - [ ] Make use of `bogged` and `breeze`
- Double Doors:
  - [x] Enable for Copper Doors
- Enchantment Extractors: #1018
  - [x] Verify Mace compatibility
- Everstone: #1031
  - [x] Verify Armadillo compatibility
  - [x] Add Bogged aging 
- Forming Press: #1020
[...]

PR #1128, opened 2025-05-10
Orb of Ankou 1.21.5

BPR02:
Fixes various pneumas (agile, anchoring, bashing, blasting, dashing, enduring, pricking, feigning, soaring).

Uses fall damage multiplier attribute for agile instead of a clock.

Uses [player motion library (v1.4.2)](https://modrinth.com/datapack/player_motion/version/1.4.2) to implement dashing mechanic. The library has been copied into our own `lib_player_motion` library to be rename-spaced and version checked.

Loot tables now exist for all soul shards.

Soul Shards and Orbs of Ankou are now non stackable (stack size of 1).

Upgrade paths are implemented by throwing the item into a soul forge and extinguishing it. The entity data for failed items is now based on loot tables instead of copying the exact item data from the soul forge.

The soul forge will no longer spawn wither rose particles on obsidian.


gm4_orb_of_ankou/beet.yaml
```diff
@@ -3,7 +3,10 @@ name: Orb of Ankou
 version: 1.7.X
 
 data_pack:
-  load: .
+  load: [
+    .,
+    https://cdn.modrinth.com/data/oDhxdGVZ/versions/aW4ph2OC/player_motion.zip
+  ]
```
misode, line 9: You should format the yaml array like this imo:
```yaml
load:
  - '.'
  - 'https://cdn.modrinth.com/data/oDhxdGVZ/versions/aW4ph2OC/player_motion.zip'
```

BPR02: true; I forgot how to properly format yaml

Bloo-dev:
Does this include souls for the Armadillo, the Bogged, and the Breeze? Asking as they are not ticked off yet in #1010.

BPR02:
Armadillo, Bogged, and Breeze souls were added in #1054

BPR02:
This uses the player motion library, which is a potential issue since it's the first time adding a non-gm4 library as a dependency. AFAIK there is no way for our dependency checker functions (now automatically generated based on beet.yaml) to account for external libraries and disable the module if an unsupported version of player motion is installed. At the very most, the dashing pneuma would change behavior, which isn't that big of a deal. Should this be addressed before merging this change?

Bloo-dev:
I'm not to keen on relying on external data pack libraries

misode:
I don't think relying on external libraries is a bad thing, as long as we properly pin the version or copy it into our sources. But I'm more wondering why this library is suddenly necessary and wasn't in the previous version. Could we not update OoA to 1.21.5 without this new mechanic?

BPR02:
The library isn't _necessary_, but it provides a much better UX and UI. The previous system locked the player into the trajectory, not allowing them to move away from the initial dash (since it makes the player ride an armor stand which is given motion) and it also used weird armor stand motion which broke. Instead of fixing the broken armor stand motion I decided it was better to do it the "proper" way with player motion.

I can try to fix dashing without player motion if we don't wanna deal with it for this update.

misode:
Using the player_motion library seems fine, as long as:
1. We won't run into problems when multiple modules try to rely on it
2. We make sure the build is still 100% reproducible
3. We can handle updating the library to a new version
4. We can handle when player_motion is included in a non-GM4 data pack, possibly a different version

BPR02:
Okay, this PR now adds player motion as a GM4 library (lib_player_motion) which has the following differences:
- the targeted player motion library is locked to one specific version of the player motion library (v1.4.2 rn since it's the latest)
- the library uses `gm4_player_motion` instead of `player_motion` namespaces
- `#minecraft:load` has been deleted to allow for our lantern load and versioning plugin to handle load
- Function tags `#gm4_player_motion:launch_xyz` and `#gm4_player_motion:launch_looking` are added to allow for version-checked function calls similar to our other libraries
- The marker spawned at 0 0 0 (for trig) is spawned (and targeted) with a different UUID to prevent conflict with the normal library marker

This handles all 4 criteria specified in the last message.

Bloo-dev (changes requested):
After some more discussion we have concluded that inclusion of the player motion library as-is is not possible due to
- it forceloading `0 0`
- it not using lantern load

We should instead fork the library and make it use our forceloaded chunk and lantern load, as well as proper versioning.

BPR02:
The player motion library has been forked to use lantern load and our forceloaded chunk. Full changes can be found in the README of lib_player_motion. 

I'm not entirely sure if the licensing is done correctly. MIT doesn't require us to list the changes made (like GPL3 does), so I think we should be fine just putting the original MIT license file in the data/gm4_player_motion folder? But I don't think that would be carried onto the download zip since it only carries over the main LICENSE file from the library root folder. Or do we just add the copyright onto our license file?

TheThanathor:
From testing in-game:

- Wither roses in the very far corner during Fusing Ritual do not work, since the marker is outside the distance check
- Wither rose markers don't get removed when the ritual is ended (by punching out the fire)

**Pneuma's**
Withering: does not work
Soaring: the no-gravity stays when a block is placed during flight, or when the Soul Shard is thrown out
Striding: does not remove the markers
Expeditious: does not work

Blinding, Synergetic and Draining were not tested as I didn't have another player

BPR02:
Okay, all pneuma issues should be addressed now. I tested the ones Thanathor couldn't (Blinding, Synergetic, Draining, Depriving) and they are working correctly.

misode (approved):

PR #1131, opened 2025-05-12
[Double Doors]: Backport Copper Door Fix from 1.21.5 Update

Bloo-dev:
Applies a fix from the 1.21.5 update of Double Doors, making the module work with Copper Doors.

Will create merge conflicts with the 1.21.5 update.


misode:
That was the reason why I didn't add copper doors to the 1.21.4 version... I though that was acceptable.

Bloo-dev:
Seems reasonable. Closing this PR.

PR #1133, opened 2025-05-16
1.21.5 - Heart Canisters upgrade path

TheThanathor:


gm4_heart_canisters/beet.yaml
```diff
@@ -19,17 +19,17 @@ meta:
         lib_custom_crafters: 3.4.0
       schedule_loops: [main]
     model_data:
-      - item: poisonous_potato
+      - item: player_head
```
misode, line 22: Why were these models changed back to player heads?

TheThanathor: The way the RP works it uses CMD on the vanilla item model, so since the heart canisters use the player_head item_model it needs to alter that instead of the poisonous potato for the RP to work

misode: Ah right of course

PR #1150, opened 2025-08-17
Fix pale oak not working with Arborenda Shamir 1.21.5+

runcows:
I missed this file when updating it. Woops (Again)


misode (changes requested):
Looking at this function, this can probably be optimized by skipping the storage copy and using `execute if items` directly on the item entity.

runcows:
Made that change, which meant we could remove that function entirely

PR #1161, opened 2025-09-17
Cosmetic Updates for Particles Pack

kindabland:
- Updated the `autumn` and `blossom` particles to use the newer falling leaves particles, instead of redstone dust particles

because i'm kind of a toddler to GitHub, this might not be done in exactly the right format. I really did read the docs and I did my best though!


Bloo-dev (approved):
Couldn't test in-game, but looks good code wise.
If one wanted to be nitpick, measuring performance of the three `if score matches 11` checks vs. a single check & function call would be interesting. I doubt it makes a significant difference.

PR #1162, opened 2025-09-25
Fix Reeling Rods Item Dupe/Deletion for container entities

runcows:
This must have been something I changed quite late, forgetting how it worked.

The issue stems from entity data not being stored before `pull_items` is called, which handles emptying container entities. This meant that it would delete items from the container entity, and spit them out again only if the next entity reeled is a container entity. If there is no leftover data from the previous entity, it would simply dupe the container block (chest or hopper).

Then there was another issue with the motion for the container entity's items coming from the previously reeled entity and not the current primary item.


TheThanathor (approved):
Tested in game.

This should be merged asap as it fixes a gamebreaking bug

PR #1164, opened 2025-10-01
Update to 1.21.9

misode:


gm4/plugins/resource_pack.py
```diff
@@ -68,7 +68,7 @@
 JsonType = dict[str,Any]
 
 CUSTOM_MODEL_PREFIX = 3420000
-MINECRAFT_REFERENECE_VERSION = "1.21.6"
+MINECRAFT_REFERENECE_VERSION = "1.21.9"
```
EpyonProjects: Reference is spelled wrong

misode:
I have tested the failing tests in-game and they are all working, which leads me to believe it might be an issue with packtest.

I consider this PR ready.

PR #1165, opened 2025-10-04
runcows misc 1.21.9 commits

runcows:
- What I've done so far with copper update stuff
- ~~Snuck in the rope ladders translation key fix~~

I'm going to try to add copper lanterns to Lively Lily Pads soon, but that might take a little bit of time, so it might end up as a separate PR


gm4_reeling_rods/data/gm4_reeling_rods/tags/entity_type/steal_equipment.json
```diff
@@ -4,6 +4,7 @@
     "minecraft:bogged",
     "minecraft:drowned",
     "minecraft:husk",
+    {"id": "minecraft:mannequin", "required": false},
```
misode, line 7: Are we sure we want to include this?

runcows: Im not entirely sure. The idea was that people are going to do some crazy stuff with mannequins and it might be weird if reeling doesn't work with them. But it's definitely weird because the rest of the design is only really to support vanilla survival features

PR #1167, opened 2025-10-10
Metallurgy adjust gold max damage + remove redundancy

runcows:
Gold tools had the wrong max damage (33 instead of 32) causing an off by 1 error

Both Audere Shamir and Moneo Shamir set a max durability custom data on tools, this redundancy has been eliminated, favoring just the function from Moneo Shamir. 
However, there is potentially an open question of if we should move this function into the `gm4_metallurgy` namespace, since it's common functionality. That has not been done here yet, but can be.


gm4_audere_shamir/data/gm4_audere_shamir/function/active_tool.mcfunction
```diff
@@ -4,7 +4,8 @@
 
 data modify storage gm4_audere_shamir:temp/player/tool SelectedItem set from entity @s SelectedItem
 execute store result score $tool_current_damage gm4_ml_data run data get storage gm4_audere_shamir:temp/player/tool SelectedItem.components."minecraft:damage"
-execute store result score $tool_max_damage gm4_ml_data run data get storage gm4_audere_shamir:temp/player/tool SelectedItem.components."minecraft:custom_data".gm4_audere_shamir.max_durability
+execute store result score $tool_max_damage gm4_ml_data run data get storage gm4_audere_shamir:temp/player/tool SelectedItem.components."minecraft:custom_data".MaxDurability
```
misode, line 7: It seemed like this was not a backwards compatible change, but I realize that moneo is always installed, and thus the `MaxDurability` tag should be present on all tools. (I think?)

The comment in moneo's `store_maximum` function could probably make it clear that it runs for all shamirs.

Related, I think right now infinitas' make_unstackable runs for every shamir, not only for infinitas, which maybe isn't a problem, but still weird.

PR #1168, opened 2025-10-12
More 1.21.9 Feature Updates

runcows:
This PR
- removes `CanDuplicate` nbt field from Gemini Shamir, this field was removed in MC 1.21.5
- adds copper torch to Crossbow Cartridges
- Updates Metallurgy and its expansions for copper tools and armor
  - Including RP changes
  - Implements changes from #1167 
- Replaces `#gm4:boots` with `#minecraft:foot_armor`
- Swaps most 1.21.9 overlays to `backport_81`
  - left out from this is Disassemblers (`since_88` contents are generated) and Standard Crafting

### DIFF
--- a/gm4/modeldata_registry.json
+++ b/gm4/modeldata_registry.json
@@ -240,11 +240,76 @@
       "gm4_fulcio_shamir:shamir/fulcio": 115,
       "gm4_animi_shamir:shamir/animi": 124
     },
+    "copper_axe": {
+      "gm4_metallurgy:shamir/hypexperia": 110,
+      "gm4_metallurgy:shamir/arborenda": 112,
+      "gm4_metallurgy:shamir/moneo": 113,
+      "gm4_percurro_shamir:shamir/percurro": 117,
+      "gm4_audere_shamir:shamir/audere": 120,
+      "gm4_animi_shamir:shamir/animi": 124
+    },
     "copper_block": {
       "gm4_auto_crafting:guidebook_icon/auto_crafting": 1
     },
+    "copper_boots": {
+      "gm4_weighted_armour:shamir/helious": 103,
+      "gm4_metallurgy:shamir/defuse": 106,
+      "gm4_desire_lines:shamir/celaro": 118,
+      "gm4_vecto_shamir:shamir/vecto": 121,
+      "gm4_metallurgy:shamir/vibro": 123,
+      "gm4_animi_shamir:shamir/animi": 124
+    },
+    "copper_chestplate": {
+      "gm4_weighted_armour:shamir/helious": 103,
+      "gm4_metallurgy:shamir/defuse": 106,
+      "gm4_iacio_shamir:shamir/iacio": 122,
+      "gm4_animi_shamir:shamir/animi": 124
+    },
+    "copper_helmet": {
+      "gm4_weighted_armour:shamir/helious": 103,
+      "gm4_metallurgy:shamir/defuse": 106,
+      "gm4_animi_shamir:shamir/animi": 124
+    },
+    "copper_hoe": {
+      "gm4_metallurgy:shamir/gemini": 109,
+      "gm4_metallurgy:shamir/tinker": 111,
+      "gm4_metallurgy:shamir/moneo": 113,
+      "gm4_audere_shamir:shamir/audere": 120,
+      "gm4_animi_shamir:shamir/animi": 124
+    },
+    "copper_leggings": {
+      "gm4_weighted_armour:shamir/helious": 103,
+      "gm4_metallurgy:shamir/defuse": 106,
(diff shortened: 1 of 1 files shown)


gm4_metallurgy/since_88/data/gm4_metallurgy/function/init.mcfunction
```diff
@@ -0,0 +1,63 @@
...
+scoreboard objectives add gm4_ml_ore_ba dummy
+scoreboard objectives add gm4_ml_ore_bi dummy
+scoreboard objectives add gm4_ml_ore_th dummy
+
+#arborenda
+scoreboard objectives add gm4_use_axe_net minecraft.used:minecraft.netherite_axe
+scoreboard objectives add gm4_use_axe_dia minecraft.used:minecraft.diamond_axe
+scoreboard objectives add gm4_use_axe_gol minecraft.used:minecraft.golden_axe
+scoreboard objectives add gm4_use_axe_iro minecraft.used:minecraft.iron_axe
+scoreboard objectives add gm4_use_axe_cop minecraft.used:minecraft.copper_axe
```
misode: Instead of creating an overlay for this entire function, and thus needing a changes in the manifest plugin (which I'm not a fan of), you should just add a function call here, and then add an overlay for only that one-line function.

runcows:
This PR has been updated a lot, and things that are more unrelated to the 1.21.9/10 update have been split into their own PRs. Will update the comments in a little bit.

gm4_audere_shamir/data/gm4_audere_shamir/function/active_tool.mcfunction
```diff
@@ -4,18 +4,21 @@
 
 data modify storage gm4_audere_shamir:temp/player/tool SelectedItem set from entity @s SelectedItem
 execute store result score $tool_current_damage gm4_ml_data run data get storage gm4_audere_shamir:temp/player/tool SelectedItem.components."minecraft:damage"
-execute store result score $tool_max_damage gm4_ml_data run data get storage gm4_audere_shamir:temp/player/tool SelectedItem.components."minecraft:custom_data".gm4_audere_shamir.max_durability
+execute store result score $tool_max_damage gm4_ml_data run data get storage gm4_audere_shamir:temp/player/tool SelectedItem.components."minecraft:custom_data".MaxDurability
+# | MaxDamage is set by gm4_moneo_shamir:store_maximum
```
TheThanathor, line 8: Does this mean Audere is now reliant on Moneo? While that's technically fine while Moneo is in base Metallurgy it might cause problems if Moneo is ever updated / removed since it's unlikely this would be thought of then

gm4_crossbow_cartridges/data/gm4_crossbow_cartridges/function/projectile/copper_torch/check.mcfunction
```diff
@@ -0,0 +1,19 @@
...
+execute if block ~ ~ ~ #gm4_crossbow_cartridges:torch run loot spawn ~ ~ ~ loot minecraft:blocks/copper_torch
+
+# place torch if possible
+execute if block ~ ~ ~ #gm4:replaceable run function gm4_crossbow_cartridges:projectile/copper_torch/place
+
+# summon torch item if placing torch failed
+execute unless block ~ ~ ~ #gm4_crossbow_cartridges:torch run loot spawn ~ ~ ~ loot minecraft:blocks/copper_torch
+
+# place block sound
+playsound minecraft:block.wood.place block @a[distance=..15]
```
TheThanathor, line 16: Also for the other materials, but I feel like this shouldn't play if the torch didn't actually place

TheThanathor (approved):
Looks good! Left two comments but they're not really blocking anything

We should maybe look at updating Metallurgy at some point to also accept custom items or items with custom durability, but for now it's fine.

gm4_audere_shamir/data/gm4_audere_shamir/function/active_tool.mcfunction
runcows: Yes, although this is a change from https://github.com/Gamemode4Dev/GM4_Datapacks/pull/1167 which is only mixed in here because I needed to handle overlays of the same files for copper tools. We can definitely move the `store_maximum` function into the metallurgy namespace if needed though

gm4_crossbow_cartridges/data/gm4_crossbow_cartridges/function/projectile/copper_torch/check.mcfunction
runcows: Oh huh, yeah. Likely a change for a different PR though so I'll make an issue to track it

PR #1181, opened 2025-10-30
Lively Lily Pads 1.21.9 Feature Update

runcows:
- \+ Copper Torch
- \+ Copper Lantern
  - when unwaxed, oxidize with a 1/6500 chance every 8 ticks. This works out to give a 50% chance of oxidization in 30 minutes
  - can be waxed or scraped, both offhand compatible w/ vanilla parity in interactions
- Reworked the punch candle interaction function chain to be generic for permanent interactions, which is used for Copper Lanterns
  - this meant an upgrade path adding a generic tag to the candle interactions

### DIFF
--- a/gm4_lively_lily_pads/data/gm4_lively_lily_pads/function/mechanics/interactions/unwaxed_copper_lantern/attempt_oxidization.mcfunction
+++ b/gm4_lively_lily_pads/data/gm4_lively_lily_pads/function/mechanics/interactions/unwaxed_copper_lantern/attempt_oxidization.mcfunction
@@ -6,7 +6,7 @@
 # this is only run once per entity every 8 ticks (0.4 seconds)
 # with a 1/6500 chance every 0.4 seconds, in 30 minutes there will be a ~50% chance of oxidization
 execute store result score $rand gm4_llp.data run random value 1..6500
-execute if score $rand gm4_llp.data matches 2..6500 run return fail
+execute if score $rand gm4_llp.data matches 2.. run return fail
 
 # oxidize
 execute if data entity @s {block_state:{Name:"minecraft:copper_lantern"}} run return run data modify entity @s block_state.Name set value "minecraft:exposed_copper_lantern"


gm4_lively_lily_pads/data/gm4_lively_lily_pads/function/mechanics/interactions/waxed_copper_lantern/interact_rcd.mcfunction
```diff
@@ -0,0 +1,26 @@
...
+
+# reset scoreboards
+scoreboard players set $mainhand gm4_llp.data 0
+scoreboard players set $offhand gm4_llp.data 0
+scoreboard players set $holding_axe gm4_llp.data 0
+scoreboard players set $scraped gm4_llp.data 0
+
+# item data, only care about axes
+execute if items entity @s weapon.mainhand #minecraft:axes store success score $holding_axe gm4_llp.data run scoreboard players set $mainhand gm4_llp.data 1
+execute unless score $mainhand gm4_llp.data matches 1 if items entity @s weapon.offhand #minecraft:axes store success score $holding_axe gm4_llp.data run scoreboard players set $offhand gm4_llp.data 1
```
TheThanathor, line 16: technically offhand axes should only work if the mainhand is empty

runcows: Testing in vanilla, with an offhand axe and a mainhand honeycomb, right clicking on a `waxed_exposed_copper_lantern` for example, it scrapes the lantern, removing its wax.

runcows: Oh but I do see what you mean though. I believe it's actually that if the mainhand is placeable the offhand axe wont activate

PR #1184, opened 2025-10-31
Update `manifest.py` to allow overlays of init functions

runcows:
Cherry picking from #1168

Small update to the manifest plugin to allow overlays of `init`.
While currently we do not have any overlays of `init` functions, the current behaviour is undocumented and confusing.
I've had issues with this behavior before (resolved in #1099 commit [ec0e9fc])

### DIFF
--- a/gm4/plugins/manifest.py
+++ b/gm4/plugins/manifest.py
@@ -10,7 +10,7 @@ from pathlib import Path
 from typing import Any, Optional
 
 import yaml
-from beet import Context, InvalidProjectConfig, PluginOptions, TextFile, load_config
+from beet import Context, InvalidProjectConfig, PluginOptions, TextFile, load_config, Function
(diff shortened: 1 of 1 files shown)


runcows:
I can't argue with that, and can't provide any use-cases right now. 
However, there has already been one instance where an overlay of `init` has made it past review, in Blossoming Pots first PR (#1070). Not supporting overlays of `init` feels like begging for more similar issues, since things like this can and will be missed in review. 
If for no other reason than to save ourselves that trouble later, I think we should support overlays of `init`, even if we don't endorse them.

misode (approved):

runcows:
Yeah, I had, but the reason for it was silly. I was using it to check the version of the big data storage to see if it needed to be changed. Overlaying `init` was to check if the current overlays format was loaded. It was a holdover from how I had blossoming pots before GM4 and definitely not necessary. That, in combination with the overlay of `init` breaking versioning through manifest, is why I removed the overlay in #1099.

Issue #387, opened 2020-04-05
Sneak-depositing 800 or more levels exceeds the max command length


misode:
### Describe the bug
It is possible to hit the max command chain length limit (65536). This will cause the `gm4_xp_storage:main` function to stop running.

### To Reproduce
1. `/xp set @s 800 levels`
2. Stand on an enderchest and deposit one level
3. Sneak and notice your xp level drops to 749
4. Notice you can no longer deposit xp and the actionbar message is not appearing anymore

### Fix
This is a temporary fix, but decreasing the 50 levels limit would decrease the chance of this happening in survival. In the future we should look into optimizing the xp depositing algorithm.

Bloo-dev:
Using macros we can now directly update XP from the scoreboard value, negating the need to recurse at all.

PR #1190, opened 2025-11-06
Xp Storage - play sound once per clock cycle

TheThanathor:


Bloo-dev (approved):

Bloo-dev:
This addresses #387, as it now caps the depositing process once the command length has been reached.

gm4_xp_storage/data/gm4_xp_storage/function/init.mcfunction
```diff
@@ -1,5 +1,6 @@
 scoreboard objectives add gm4_stored_xp dummy
 scoreboard objectives add gm4_xp_calc dummy
+scoreboard objectives add gm4_xp_data dummy
```
misode, line 3: Why was `gm4_xp_calc` not reused here?

TheThanathor: I dont like using random scoreboards for unrelated things, so I added a general `data` score

PR #1192, opened 2025-11-08
Metallurgy item selector optimizations

runcows:
Reworks metallurgy's main function to be more resistant to crashes from hitting the max command limit. 
Schedule now happens first so that it doesn't need to be rescheduled.
Item entity processing has now moved into main, with more limits on the main selector.

### DIFF
--- a/gm4_metallurgy/data/gm4_metallurgy/function/main.mcfunction
+++ b/gm4_metallurgy/data/gm4_metallurgy/function/main.mcfunction
@@ -7,7 +7,7 @@ effect give @a[gamemode=!spectator,gamemode=!creative,predicate=gm4_metallurgy:c
 execute as @e[type=armor_stand,tag=gm4_sand_ring] at @s positioned ~ ~1.23 ~ run function gm4_metallurgy:casting/sustain_mould
 
 # check for moulds waiting to be created
-execute as @e[type=item,predicate=gm4_metallurgy:is_obsidian,predicate=gm4_metallurgy:on_ground] at @s if block ~ ~-0.1 ~ #minecraft:sand run function gm4_metallurgy:casting/check_mould_creation
+execute as @e[type=item,predicate=gm4_metallurgy:is_obsidian,predicate=gm4_metallurgy:on_ground] at @s if block ~ ~-0.1 ~ #minecraft:sand align xyz unless entity @e[type=armor_stand,tag=gm4_sand_ring,dx=0,dy=-1,dz=0,limit=1] if entity @e[type=item,dx=0,dy=0,dz=0,predicate=gm4_metallurgy:is_clay_ball,predicate=gm4_metallurgy:on_ground,limit=1] run function gm4_metallurgy:casting/create_mould
 
 # check for shamir on anvil
 scoreboard players reset found_item_on_anvil gm4_ml_data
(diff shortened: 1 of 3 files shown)


runcows:
Reworked roughly following the above comment.
Any entity selector in metallurgy that is checking for OnGround nbt has been updated to use a predicate.

gm4_metallurgy/data/gm4_metallurgy/function/main.mcfunction
```diff
@@ -1,12 +1,17 @@
...
-# manage items
 scoreboard players reset found_item_on_anvil gm4_ml_data
-execute as @e[type=item] run function gm4_metallurgy:item
+# check for moulds waiting to be created
+execute as @e[type=item,predicate=gm4_metallurgy:is_obsidian,predicate=gm4_metallurgy:on_ground] at @s if block ~ ~-0.1 ~ #minecraft:sand run function gm4_metallurgy:casting/check_mould_creation
+
+# check for shamir on anvil
+execute as @e[type=item,tag=!gm4_ml_smooshed] if items entity @s contents *[count=1,custom_data~{gm4_metallurgy:{has_shamir:1b}}] at @s if block ~ ~-1 ~ #minecraft:anvil run function gm4_metallurgy:smooshing/item_on_anvil
 
 #============================== SHAMIR CLOCKING ================================
```
Bloo-dev, line 16: I'm not a big fan of increasing the amount of `@e[type=item]` selectors on the main clock.

Its by far the most expensive selector.

runcows: I've limited the 2nd selector further. I think that these 2 limited item selectors will be better than just the one from before, but I am not entirely sure

Bloo-dev:
I'm not sure if limiting the selector by tags or predicates does anything significant for performance.

Modifiers like `distance`, `dx` (or its siblings), and `limit` improve performance by actually limiting the list of entities that is checked vs the selector (`distance` and `dx` limit the amount of chunks that are checked, `limit` stops selection after the limit has been reached).
This is, as far as I am aware, not the case for `tag` or `predicate`, and having to check those against all entities may even be more expensive.

gm4_metallurgy/data/gm4_metallurgy/function/casting/add_metal/initialize.mcfunction
```diff
@@ -25,4 +25,4 @@ scoreboard players reset $is_obsidian_cast gm4_ml_data
 data remove storage gm4_metallurgy:temp/item/ore gm4_metallurgy
 
 # make all ores on top jump
-execute as @e[type=item,tag=!gm4_ml_in_animation,dx=0,dz=0,predicate=gm4_metallurgy:is_castable,nbt={OnGround:1b},limit=1] run data merge entity @s {Motion:[0.0,0.35,0.0],PickupDelay:30,Tags:["gm4_ml_in_animation"],Item:{components:{"minecraft:custom_data":{gm4_metallurgy:{ore_in_animation:1b}}}}}
```
misode, line 28: Where did this ore_in_animation check go to?

runcows: It was removed. I looked through the repo and couldn't find anything that was actually using the custom data. In current metallurgy, it's set here in this function and removed in `metallurgy:item`, but never read or referenced out side of that.

Bloo-dev (approved):

PR #1194, opened 2025-11-08
Lavish Livestock

Bloo-dev:
Introduces the Lavish Livestock module which allows players to gain access to slightly larger versions of livestock via selective breeding.

## How it works
When a player breeds one of the supported mobs (chickens, cows, mooshrooms, pigs, rabbits, sheep, and hoglins), there is a 5% chance for the baby to be one-size-up from one of the parents (the parent is picked at random).
Two size variations exist; +12.5% and +25%.

On death, these size variations drop more loot; up to 3×base loot for +12.5% and up to 5×base loot for +25%.

image
Pigs of all sizes in a pen.

image
Chickens of all sizes in a pen.

## TODO
- [ ] ~~Compatibility with Gemini (currently load order dependent)~~ Should be done in a separate PR
- [ ] ~~Compatibility with Pig Tractors (larger pigs should til more tiles)~~ Should be done in a separate PR

### DIFF
--- a/gm4_lavish_livestock/README.md
+++ b/gm4_lavish_livestock/README.md
@@ -4,4 +4,4 @@ Selectively breed your livestock for increased yields! Commercialize your farmin
 
 ### Features
 - Breeding livestock may result in offspring that is abnormally large
-- Larger individuals drop more loot
+- Larger livestock drops more loot
--- a/gm4_lavish_livestock/generate.py
+++ b/gm4_lavish_livestock/generate.py
@@ -14,7 +14,7 @@ def beet_default(ctx: Context):
     # prepare list of supported entities
     entity_ids = [e.get('entity_id', "").removeprefix('minecraft:').strip() for e in csv]
 
-    # for each wood type in the vanilla doors tag, render a copy of the "templates" directory with the appropriate wood-type
+    # for each supported entity, render a copy of the "templates" directory with the appropriate entity-id
     for entity in entity_ids:
         subproject_config = {
             "data_pack": {


Bloo-dev:
After some back and forth I have discarded the plan to make this work with Gemini.

Gemini needs a more comprehensive update in a separate PR, see #1196.

gm4_lavish_livestock/data/gm4_lavish_livestock/function/init.mcfunction
runcows, line 1: `init` should not be in the templates, as this will put it under `gm4_lavish_livestock:<animal>/init` which is incorrect for a whole host of reasons and I'm actually not sure how/if it was running at all.

gm4_lavish_livestock/data/gm4_lavish_livestock/templates/function/determine_parent.mcfunction
```diff
@@ -0,0 +1,11 @@
...
+# @s = parent of the baby
+# at location of player who has fed the parent
+# run from gm4_lavish_livestock:{{ entity_id }}/revoke_advancement
+
+# init score if parent does not have a score
+scoreboard players add @s gm4_lavish_livestock_size 0
+
+# store size of parent into fixed fake player (only one of these triggers, the other one triggers for the other parent)
+execute unless score $parent_a gm4_lavish_livestock_size matches -2147483648..2147483647 run return run scoreboard players operation $parent_a gm4_lavish_livestock_size = @s gm4_lavish_livestock_size
+execute unless score $parent_b gm4_lavish_livestock_size matches -2147483648..2147483647 run return run scoreboard players operation $parent_b gm4_lavish_livestock_size = @s gm4_lavish_livestock_size
```
runcows, line 11: It might be more readable to initialize `$parent_a` and `$parent_b` to -1 as a default "uninitialized value" in `revoke_advancement` and test for that here, rather than testing for any score at all.

gm4_lavish_livestock/data/gm4_lavish_livestock/templates/function/find_baby.mcfunction
```diff
@@ -0,0 +1,12 @@
+# Selects babies near the breeding site
+# @s = breeding site
+# at @s
+# run from gm4_lavish_livestock:{{ entity_id }}/find_marker
+
+# cache parent size
+scoreboard players reset $size gm4_lavish_livestock_size
```
runcows, line 7: Resetting the size is unnecessary when assigning the value one command later.

gm4_lavish_livestock/README.md
```diff
@@ -0,0 +1,7 @@
+# Lavish Livestock<!--$headerTitle--><!--$pmc:delete-->
+
+Selectively breed your livestock for increased yields! Commercialize your farming!<!--$pmc:headerSize-->
+
+### Features
+- Breeding livestock may result in offspring that is abnormally large
+- Larger individuals drop more loot
```
runcows: I think it would be more clear to swap "individuals" with "livestock" or "animals". While it's unlikely to cause much confusion, I still think it would be better to be more specific.

gm4_lavish_livestock/data/gm4_lavish_livestock/templates/advancement/breed.json
```diff
@@ -0,0 +1,15 @@
+{
+    "criteria": {
+        "breed_pigs": {
```
runcows, line 3:
```suggestion
        "breed_livestock": {
```

gm4_lavish_livestock/generate.py
```diff
@@ -0,0 +1,39 @@
...
+
+from beet import Context, subproject
+from gm4.utils import CSV
+
+logger = logging.getLogger(__name__)
+
+def beet_default(ctx: Context):
+
+    # load csv
+    csv = CSV.from_file(Path('gm4_lavish_livestock', 'raw', 'livestock.csv'))
```
BPR02, line 12: do we need a CSV for this? Why not a json file or even just a list in this python script

gm4_lavish_livestock/generate.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # prepare list of supported entities
+    entity_ids = [e.get('entity_id', "").removeprefix('minecraft:').strip() for e in csv]
+
+    # for each wood type in the vanilla doors tag, render a copy of the "templates" directory with the appropriate wood-type
```
BPR02: comment is copied from double doors

gm4_lavish_livestock/data/gm4_lavish_livestock/templates/function/determine_parent.mcfunction
BPR02: Yeah, resetting is worst than setting to a known "uninitialized value"

BPR02 (changes requested):
Needs some changes as mentioned by @runcows. But looks pretty good. Could use a guidebook entry.

PR #1204, opened 2025-11-10
Crossbow Cartridges 1.21.5: Fixes projectile ownership

EpyonProjects:
Looks like this was incorrectly set from the start. projectile.uuid set from @s UUID cannot work entities cannot have the same UUID.  This most likely was meant to set the projectile.Owner (or .owner, in the case of tnt) as to attribute a projectile to the original player.  This works for arrows but not for the ones this PR modifies (TNT, fireballs, and wind charges) as those are freshly spawned.

This should address #1037 where moderation tools were not able to log potential grief/pvp cases

### DIFF
--- a/docs/resource-pack-management.md
+++ b/docs/resource-pack-management.md
@@ -196,6 +196,50 @@ model_data:
       - reference: item/lump/thorianite
 ```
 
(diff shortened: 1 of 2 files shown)


misode:
This PR is impossible to review because it has so many unrelated commits. Please rebase or cherry-pick this and select the right base.

PR #1209, opened 2025-11-14
Vibro Shamir - use attributes

TheThanathor:
Changes Vibro from using effects to using attributes, this allows it to stack with jump boost, and makes it more responsive to use.

_note: this does not apply the safe fall distance from jump boost effect, but that is normally gone on landing anyway_

### DIFF
--- a/gm4_metallurgy/data/gm4_vibro_shamir/function/remove_jump.mcfunction
+++ b/gm4_metallurgy/data/gm4_vibro_shamir/function/remove_jump.mcfunction
@@ -1,3 +1,5 @@
+# @s = vibro player no longer sneaking on the ground
+# run from gm4_metallurgy:tick
 
 scoreboard players reset @s gm4_vibro_sneak
 attribute @s jump_strength modifier remove gm4_vibro_shamir:jump_boost


gm4_metallurgy/data/gm4_vibro_shamir/function/jump.mcfunction
```diff
@@ -5,10 +5,16 @@
...
 
 # apply jump boost based on sneak number
-effect give @s[scores={gm4_vibro_sneak=1..5}] jump_boost 1 3 false
-effect give @s[scores={gm4_vibro_sneak=6..10}] jump_boost 1 5 false
-effect give @s[scores={gm4_vibro_sneak=11..15}] jump_boost 1 6 false
-effect give @s[scores={gm4_vibro_sneak=16..}] jump_boost 1 9 false
+attribute @s[scores={gm4_vibro_sneak=1}] jump_strength modifier add gm4_vibro_shamir:jump_boost 0.4 add_value
+
+attribute @s[scores={gm4_vibro_sneak=6}] jump_strength modifier remove gm4_vibro_shamir:jump_boost
+attribute @s[scores={gm4_vibro_sneak=6}] jump_strength modifier add gm4_vibro_shamir:jump_boost 0.6 add_value
```
misode, line 11: If you add a modifier with the same ID, does it not overwrite the previous modifier value?

gm4_metallurgy/data/gm4_vibro_shamir/function/remove_jump.mcfunction
```diff
@@ -0,0 +1,3 @@
+
+scoreboard players reset @s gm4_vibro_sneak
+attribute @s jump_strength modifier remove gm4_vibro_shamir:jump_boost
```
misode, line 5: Add a function header

gm4_metallurgy/data/gm4_vibro_shamir/function/jump.mcfunction
TheThanathor: Nope, you have to remove it first

PR #1210, opened 2025-11-14
Spawner Minecarts - feature update

TheThanathor:
Updates (or rather removes) features from Spawner Minecarts.

- Removed any fueling of Spawner Minecarts (no more flowers etc)
- Every Spawner Minecart within a 9x8x9 area of another now reduces the spawn rate, at least 4 other Spawner Minecarts reduce it to 0 (subject to change)
- Spawner Minecarts accept any spawner, unless it's a smithed block or a mob inside the spawner data has `gm4_spawner_minecarts:{ban_pickup:1}`

### DIFF
--- a/gm4_spawner_minecarts/data/gm4_spawner_minecarts/function/capture_spawner.mcfunction
+++ b/gm4_spawner_minecarts/data/gm4_spawner_minecarts/function/capture_spawner.mcfunction
@@ -11,8 +11,9 @@ execute if data storage gm4_spawner_minecarts:temp spawner.SpawnPotentials[{data
 execute if data storage gm4_spawner_minecarts:temp spawner.SpawnData[{data:{entity:{data:{gm4_spawner_minecrarts:{ban_pickup:1}}}}}] run return run function gm4_spawner_minecarts:capture_spawner/fail_pickup
 
 # pick up spawner
-summon spawner_minecart ~ ~ ~ {Tags:["gm4_spawner_minecart"],MaxNearbyEntities:6,RequiredPlayerRange:16,SpawnCount:4,SpawnData:{entity:{id:"minecraft:pig"}},MaxSpawnDelay:600s,Delay:-1,SpawnRange:4s,MinSpawnDelay:100s,SpawnPotentials:[],Invulnerable:1b}
+summon spawner_minecart ~ ~ ~ {Tags:["gm4_spawner_minecart","gm4_spawner_minecart.new"],MaxNearbyEntities:6,RequiredPlayerRange:16,SpawnCount:4,SpawnData:{entity:{id:"minecraft:pig"}},MaxSpawnDelay:600s,Delay:-1,SpawnRange:4s,MinSpawnDelay:100s,SpawnPotentials:[],Invulnerable:1b}
 setblock ~ ~1 ~ air destroy
-data modify entity @e[tag=gm4_spawner_minecart,limit=1,distance=..1] {} merge from storage gm4_spawner_minecarts:temp spawner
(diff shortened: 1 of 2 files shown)


gm4_spawner_minecarts/data/gm4_spawner_minecarts/function/capture_spawner.mcfunction
```diff
@@ -0,0 +1,18 @@
...
+data modify storage gm4_spawner_minecarts:temp spawner set from block ~ ~1 ~
+
+# check if this spawner is blacklisted
+execute if data storage gm4_spawner_minecarts:temp spawner.SpawnPotentials[{data:{entity:{data:{gm4_spawner_minecrarts:{ban_pickup:1}}}}}] run return run function gm4_spawner_minecarts:capture_spawner/fail_pickup
+execute if data storage gm4_spawner_minecarts:temp spawner.SpawnData[{data:{entity:{data:{gm4_spawner_minecrarts:{ban_pickup:1}}}}}] run return run function gm4_spawner_minecarts:capture_spawner/fail_pickup
+
+# pick up spawner
+summon spawner_minecart ~ ~ ~ {Tags:["gm4_spawner_minecart"],MaxNearbyEntities:6,RequiredPlayerRange:16,SpawnCount:4,SpawnData:{entity:{id:"minecraft:pig"}},MaxSpawnDelay:600s,Delay:-1,SpawnRange:4s,MinSpawnDelay:100s,SpawnPotentials:[],Invulnerable:1b}
+setblock ~ ~1 ~ air destroy
+data modify entity @e[tag=gm4_spawner_minecart,limit=1,distance=..1] {} merge from storage gm4_spawner_minecarts:temp spawner
```
BPR02: This could target existing spawner minecarts on the same block as the minecart this function is run at, leaving an uninitialized pig spawner minecart. Maybe use `execute summon` or a temporary tag.

gm4_spawner_minecarts/data/gm4_spawner_minecarts/function/main.mcfunction
```diff
@@ -1,21 +1,7 @@
...
-
-# slow clock for active regeneration
-scoreboard players add active_clock gm4_spawner_fuel 1
-execute if score active_clock gm4_spawner_fuel matches 2.. run scoreboard players set active_clock gm4_spawner_fuel 0
+schedule function gm4_spawner_minecarts:main 16t
 
-# global timer
-execute if score passive_clock gm4_spawner_fuel matches 0 run scoreboard players add spawner_global_time gm4_spawner_time 1
+# check for minecarts under spawners
+execute as @e[type=minecart,tag=!smithed.entity] at @s if block ~ ~1 ~ spawner align xyz positioned ~0.5 ~1.5 ~0.5 unless entity @e[tag=smithed.block,distance=..0.5] at @s if block ~ ~2 ~ #gm4_spawner_minecarts:piston[facing=down] if block ~ ~3 ~ redstone_block run function gm4_spawner_minecarts:capture_spawner
```
BPR02: Could use a `limit=1` on the smithed block `unless entity` check. Just in case there are a bunch of entities within 0.5 blocks. In practice probably doesn't matter.

BPR02 (changes requested):
Yeah this looks good. I think the nearby spawner minecart check solves the balancing issue in an elegant and simple way. No more hyper spawners from infinitely stacked minecarts in a single block and no more complicated rules to keep the minecart running. Plus with the player required radius, it means there is a finite amount of spawner minecarts a single player can load. Maybe we could make spawner minecarts slower than normal spawners (i.e. increased delay) if we wanna keep a "running cost" instead of allowing spawners to be moved with no downside at all.

PR #1211, opened 2025-11-14
Soul Glass - change to normal furnace

TheThanathor:
Changes Soul Glass recipe to use a normal furnace instead of a blast furnace, which makes much more sense
Also removes the old marker removal, which I think is probably no longer required

### DIFF
--- a/gm4_soul_glass/beet.yaml
+++ b/gm4_soul_glass/beet.yaml
@@ -27,6 +27,8 @@ meta:
     website:
       description: Reverse the effects of beacons with corrupting glass! But be careful, a poison Beacon may sound great at first, but is incredibly painful in pratice.
       recommended: []
+      notes:
+        - On versions <1.21, Blast Furnaces may need to be broken and replaced to be able to craft Soul Glass when the module is first installed.
     modrinth:
       project_id: Hi8fkYoI
     video: null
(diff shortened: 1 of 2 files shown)


gm4_soul_glass/data/gm4_soul_glass/function/main.mcfunction
```diff
@@ -1,7 +1,4 @@
-# Kill old blast furnace markers, remove this later
-kill @e[type=marker,tag=gm4_sg_furnace]
```
runcows, line 2: I don't know how I feel about removing this so soon after removing the markers, which was 11 months ago I believe.

gm4_soul_glass/beet.yaml
```diff
@@ -27,8 +27,6 @@ meta:
     website:
       description: Reverse the effects of beacons with corrupting glass! But be careful, a poison Beacon may sound great at first, but is incredibly painful in pratice.
       recommended: []
-      notes:
-        - On versions <1.21, Blast Furnaces may need to be broken and replaced to be able to craft Soul Glass when the module is first installed.
```
runcows, line 31: This note was added 3 months ago and I don't think it should be removed yet. It is still true for versions < 1.21

gm4_soul_glass/data/gm4_soul_glass/function/main.mcfunction
TheThanathor: I think it can prob be removed now, but no real harm in keeping it a while longer

misode (approved):

PR #1222, opened 2025-12-02
Moneo Shamir - Offhand compat

runcows:
Adds offhand compat for moneo as a separate function chain.

~~Currently the only moneo tool that can be used in the offhand is shears. As such, it is the only thing in the offhand function chain.~~


BPR02:
The point of moneo is to prevent the player from accidentally breaking their tool while using it. Does mining fatigue slow down the right click abilities? If not, does it at least give an obvious indication that they should stop using the tool?

misode:
We'll have to revisit offhand support later

PR #1225, opened 2025-12-03
Make `beet.yaml` video field optional

runcows:
A majority of our modules have no video, so in the `beet.yaml` the field sits as null. This is easily forgotten, as has happened several times in new module PRs. 

This PR makes the field optional as well as cleaning up existing video fields in all modules `beet.yaml` files.
- `video: null` fields have been removed. They would still work, I just figured I'd clean them up while I was here
- XP storages video link has been adjusted to no longer point to a playlist
- ~~Particles Pack and Poses Pack each had their video link adjusted to no longer include a time argument~~

### DIFF
--- a/gm4/plugins/manifest.py
+++ b/gm4/plugins/manifest.py
@@ -61,7 +61,7 @@ class ManifestModuleModel(BaseModel):
 	name: str
 	version: str
 	hash: str
-	video_link: str|None = ""
+	video_link: str = ""
 	wiki_link: str = ""
 	credits: CreditsModel
 	requires: list[str] = []
@@ -109,7 +109,7 @@ def create(ctx: Context):
 					name = config.name,
 					version = config.version,
 					hash = "",
-					video_link = gm4_meta.video if gm4_meta.video else None,
+					video_link = gm4_meta.video or "",
 					wiki_link = gm4_meta.wiki or "",
 					credits = gm4_meta.credits,
 					requires = [e for e in gm4_meta.versioning.required.keys() if not e.startswith("lib")] if gm4_meta.versioning else [],


misode:
I believe that this was an intentional decision. Keeping `video: null` in all the `beet.yaml` files makes it clear where you can place the video link. I don't see any advantage in this tbh.

> This is easily forgotten, as has happened several times in new module PRs.

What is easily forgotten? Adding the field? It shouldn't be if you either copy bat grenades or follow the making-a-module docs.

> Particles Pack and Poses Pack each had their video link adjusted to no longer include a time argument

This is intended. They are part of the better armour stands video.

runcows:
> They are part of the better armour stands video.

Ah I totally missed that, woops. That change has been reverted

gm4/plugins/manifest.py
```diff
@@ -61,7 +61,7 @@ class ManifestModuleModel(BaseModel):
 	name: str
 	version: str
 	hash: str
-	video_link: str = ""
+	video_link: Optional[str]
```
misode: Don't change the manifest format, keep this as an empty string, as the website may rely on it

gm4/plugins/manifest.py
```diff
@@ -50,7 +50,7 @@ class ManifestConfig(PluginOptions, extra=Extra.ignore):
 	smithed: Optional[SmithedConfig]
 	pmc: Optional[PMCConfig]
 	# promo
-	video: str|None
+	video: Optional[str]
```
misode: Use `str|None` just like `wiki` does

misode (changes requested):

gm4/plugins/manifest.py
```diff
@@ -109,7 +109,7 @@ def create(ctx: Context):
 					name = config.name,
 					version = config.version,
 					hash = "",
-					video_link = gm4_meta.video or "",
+					video_link = gm4_meta.video if gm4_meta.video else None,
```
misode: Revert this as well. This will currently output `null` in the json, when it was previously an empty string

gm4/plugins/manifest.py
```diff
@@ -61,7 +61,7 @@ class ManifestModuleModel(BaseModel):
 	name: str
 	version: str
 	hash: str
-	video_link: str = ""
+	video_link: str|None = ""
```
misode: Revert this change in the manifest

misode (approved):
I'm still not convinced that we even need this at all, but it looks good in that this would work.

PR #1231, opened 2025-12-08
Crossbow Cartridges - Fix torch place fail playing placement sound

### LINKED ISSUE #1187: Crossbow Cartridges Torch placement failure shouldn't play the place sound
Currently, `block.wood.place` is played for fired crossbow cartridges torches even when the torches fail to place. Failure to place happens when there's already a torch in the target block or if the target block isn't of `#gm4:replaceable`. This playsound is currently in the `gm4_crossbow_cartridges:projectile/<torch>/check` functions.

Likely what we should do is move this playsound command into the `gm4_crossbow_cartridges:projectile/<torch>/place` functions. Additionally, we could have a different sound play when the torch fails to place, potentially `entity.item.pickup` or another similar item pop sound.

runcows:
Closes #1187

### DIFF
--- a/gm4_crossbow_cartridges/data/gm4_crossbow_cartridges/tags/item/valid_items.json
+++ b/gm4_crossbow_cartridges/data/gm4_crossbow_cartridges/tags/item/valid_items.json
@@ -5,6 +5,7 @@
 		"minecraft:torch",
 		"minecraft:soul_torch",
 		"minecraft:redstone_torch",
+		{"id": "minecraft:copper_torch", "required": false},
 		"minecraft:splash_potion",
 		"minecraft:lingering_potion",
 		"minecraft:fire_charge",


gm4_crossbow_cartridges/data/gm4_crossbow_cartridges/function/projectile/copper_torch/check.mcfunction
```diff
@@ -4,16 +4,17 @@
 # run from projectile/copper_torch/loop
 
 # summon torch item if torch already exists at this block
-execute if block ~ ~ ~ #gm4_crossbow_cartridges:torch run loot spawn ~ ~ ~ loot minecraft:blocks/copper_torch
+execute if block ~ ~ ~ #gm4_crossbow_cartridges:torch store success score $dropped_item gm4_cb_misc run loot spawn ~ ~ ~ loot minecraft:blocks/copper_torch
```
misode: Do we even need this first check? The next "#gm4:replaceable" should also fail if the current block is a torch. The `execute store` can be moved just before calling the `place` function, and then the "summon torch" and "item drop" can check this score with opposite values.

runcows: Oh you're so right actually. Did some refactoring, should make more sense now

runcows:
Also added copper torch to `#gm4_crossbow_cartridges:valid_items` since I guess I forgot to for the 1.21.9 update, meaning that it wasn't working this entire time .,,..,

Anyway these changes are now tested in game and work as intended

gm4_crossbow_cartridges/data/gm4_crossbow_cartridges/function/projectile/copper_torch/check.mcfunction
```diff
@@ -3,17 +3,13 @@
...
 # place torch if possible
 execute if block ~ ~ ~ #gm4:replaceable run function gm4_crossbow_cartridges:projectile/copper_torch/place
 
 # summon torch item if placing torch failed
-execute unless block ~ ~ ~ #gm4_crossbow_cartridges:torch run loot spawn ~ ~ ~ loot minecraft:blocks/copper_torch
-
-# place block sound
-playsound minecraft:block.wood.place block @a[distance=..15]
+execute unless score $torch_placed gm4_cb_misc matches 1 run loot spawn ~ ~ ~ loot minecraft:blocks/copper_torch
+execute unless score $torch_placed gm4_cb_misc matches 1 run playsound minecraft:entity.item.pickup block @a[distance=..15]
```
misode, line 11: I don't think the score checks are correct here. They need to be opposite conditions. If the torch was placed, you play the sound, otherwise you spawn the loot table.

runcows: That sound is for the item dropping so they should both happen at the same time. The sound for placing the torch is run from the place function, it was moved there and this item drop sound was added.

misode (approved):
Tested, works well.

PR #1238, opened 2025-12-10
Make apple trees check a region before being placed

BPR02:


misode:
Since this uses `clone filtered force` it creates block updates nearby, that may not be ideal (see image). An alternative would maybe be a predicate that checks all required block positions, similar to our soul glass beacon beam predicate.

javaw_ALP84knDle

BPR02:
Hmm, yeah. Do we know if many block checks in a predicate significantly contributes to lag (spikes)? The clone method is currently the same method used for Towering Trees, so we might wanna do some performance testing to see if a predicate would be fine (especially for Towering Trees since those would have a lot more block checks). I think for apple trees, the predicate wouldn't significantly affect performance, so I can switch to that.

misode (approved):
Tested, code looks good.

PR #1240, opened 2025-12-14
DRAFT - timeline module

TheThanathor:
Not yet sure if this will even turn into a gm4 module, but it's a place to test out and discuss what's possible with Timelines.

Note that it currently doesn't seem like this will be very compatible with basically anything, since it needs to override worldgen and completely breaks the /time command (pls fix Mahjong)


misode:
> completely breaks the /time command

What about it breaks? Is there a mojira issue?

PR #1242, opened 2025-12-23
Add settings to Desire Lines

YoavTC:
Players can now toggle on and off if flowers & grass blocks break or not using the `/function gm4_desire_lines:config` command
image


YoavTC:
You do you, I just modified it for my own server and thought I could also make a PR for the project.
Great stuff btw!

PR #1244, opened 2025-12-24
Lavish livestock address review comments

runcows:
- removed the CSV file in favor of just having the list in `generate.py`
- adds basic guidebook
- addresses the remaining review comments

### DIFF
--- a/gm4_lavish_livestock/generate.py
+++ b/gm4_lavish_livestock/generate.py
@@ -7,7 +7,7 @@ logger = logging.getLogger(__name__)
 def beet_default(ctx: Context):
     # List of supported animals defined in beet.yaml as well as the guidebook page
     # for each supported entity, render a copy of the "templates" directory with the appropriate entity-id
-    for entity in ctx.meta['animals']:
+    for entity in ctx.meta['livestock']:
         subproject_config = {
             "data_pack": {
                 "load": [
--- a/gm4_lavish_livestock/beet.yaml
+++ b/gm4_lavish_livestock/beet.yaml
@@ -16,7 +16,7 @@ pipeline:
   - gm4.plugins.extend.module
 
 meta:
-  animals:
+  livestock:
     - chicken
     - cow
     - hoglin
--- a/gm4_lavish_livestock/data/gm4_lavish_livestock/tags/entity_type/livestock.json
+++ b/gm4_lavish_livestock/data/gm4_lavish_livestock/tags/entity_type/livestock.json
@@ -1,3 +1,3 @@
 {
-  "values": ctx.meta['animals']
+  "values": ctx.meta['livestock']
 }


runcows:
I actually firmly disagree.

I think the CSV was actually more buried and less friendly since it was in the `raw` folder. If I were a less code savvy person, unfamiliar with our code base, I'd first check data, but see nothing, then see a file called generate and check there. Separating the CSV then adds a 3rd place to check, which you'd only find if you can read the python CSV reading or just check everywhere out of frustration.

runcows:
I actually firmly disagree.

I think the CSV was actually more buried and less friendly since it was in the `raw` folder. If I were a less code savvy person, unfamiliar with our code base, I'd first check data, but see nothing, then see a file called generate and check there. Separating the CSV then adds a 3rd place to check, which you'd only find if you can read the python CSV reading or just check everywhere out of frustration.

runcows:
I can change it back to the CSV if that's desired, but I do disagree. Not enough to die on this hill and stall this PR longer than needed though 😅

gm4_lavish_livestock/data/gm4_lavish_livestock/guidebook/lavish_livestock.json
```diff
@@ -0,0 +1,58 @@
...
+            "translate":"text.gm4.guidebook.lavish_livestock.list_header",
+            "fallback":"Livestock",
+            "underlined":true
+          },
+          {
+            "text":"\n\n"
+          },
+          {
+            "translate":"text.gm4.guidebook.lavish_livestock.list",
+            "fallback": "Chicken\nCow\nHoglin\nMooshroom\nPig\nRabbit\nSheep",
```
BPR02: Maybe this should just use the translate strings of the entities?

gm4_lavish_livestock/data/gm4_lavish_livestock/templates/advancement/breed.json
```diff
@@ -1,6 +1,6 @@
 {
     "criteria": {
-        "breed_pigs": {
+        "breed_{{ entity_id }}s": {
```
BPR02: "breed_sheeps"

gm4_lavish_livestock/translations.csv
```diff
@@ -0,0 +1,5 @@
+key,en_us
+text.gm4.guidebook.module_desc.lavish_livestock,"Selectively breed your livestock for increased yields! Commercialize your farming!"
+text.gm4.guidebook.lavish_livestock.description,"Breeding livestock may result in offspring that is abnormally large.\n\nLarger livestock drops more loot."
+text.gm4.guidebook.lavish_livestock.list_header,"Livestock"
+text.gm4.guidebook.lavish_livestock.list,"Chicken\nCow\nHoglin\nMooshroom\nPig\nRabbit\nSheep"
```
BPR02: this translate string `text.gm4.guidebook.lavish_livestock.list` is no longer necessary

BPR02 (approved):

gm4_lavish_livestock/generate.py
```diff
@@ -1,19 +1,23 @@
...
+    entity_ids = [
+        "chicken",
+        "cow",
+        "hoglin",
+        "mooshroom",
+        "pig",
+        "rabbit",
+        "sheep"
+    ]
+    ctx.meta['animals'] = entity_ids
```
misode: Does it not work to define this list in the `meta` object in `beet.yaml`?

runcows: I didn't realize that was possible

runcows: @Bloo-dev what are your thoughts on that?

Bloo-dev: I didn't realize that was possible either, but that seems like the obvious choice! In that case maybe we should use a somewhat more verbose name for the list, maybe `supported_animals`?

runcows: Moved the list to the beet.yaml, just saw the edited message though. Can make that rename change later. What do you think about unifying all the naming into "Livestock"? it's what I've used for the tag list and the guidebook list

Bloo-dev: Yeah, that seems more clear. That's also consistent with `README.md`

runcows:
Alright renamed the list to `livestock`, that should now be everything for this PR

PR #1250, opened 2026-01-14
Add Hozz to contributors.json

BPR02:
This has no visual effect anywhere since the website will display it without a link, but the contributors file should contain the name of every contributor.

### DIFF
--- a/gm4/contributors.json
+++ b/gm4/contributors.json
@@ -44,12 +44,12 @@
         "links":[]
     },
     {
-        "name":"Hozz",
-        "links":[]
-    },
-    {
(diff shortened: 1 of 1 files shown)


gm4/contributors.json
```diff
@@ -43,6 +43,10 @@
         "name":"foodiebonus",
         "links":[]
     },
+    {
+        "name":"Hozz",
```
misode: H comes after F in the alphabet

BPR02: I didn't realize this was alphabetical tbh

PR #1228, opened 2025-12-07
Towering Trees

TheThanathor:
Adds mega / normal variants to all vanilla saplings
Also updates lib_trees to support this module

_Note that this does not currently add giant bamboo or azalea, as those are not grown from saplings_

PR #1252, opened 2026-01-16
Lib Trees - support mangrove and pale oak

TheThanathor:
Updates lib trees to version 1.4.x
- fixes mangrove propagules to be properly supported
- adds pale oak saplings
- adds the function tag `place_sapling` as alternative to the advancement

### DIFF
--- a/lib_trees/beet.yaml
+++ b/lib_trees/beet.yaml
@@ -1,7 +1,7 @@
 id: gm4_trees
 name: Gamemode 4 Trees
 version: 1.4.X
-description: A mcfunction library that facilites the creation of custom trees.
+description: A mcfunction library that facilitates the creation of custom trees.
 
 data_pack:
   load: 


runcows:
Looks good to me. Wasn't sure on the function tag in the review of Towering Trees, but I was definitely wrong and it is needed for a hook in.
Should be merged before #1228 (Towering Trees)

lib_trees/data/gm4_trees/tags/function/place_sapling.json
```diff
@@ -0,0 +1,5 @@
+{
+    "values": [
+        "gm4_trees:sapling/place_sapling"
```
misode, line 3: I don't see this function tag called from anywhere?

runcows: It's currently being used in towering trees to bypass the advancement for lib trees. The lib trees advancement only triggers if the placed sapling has custom data, which towering trees needs to bypass to work with vanilla saplings. Since functions from libraries are namespaced with the version numbers, it's better to add a function tag as a hook here rather than baking in the version number into the function call in towering trees

misode: I see

lib_trees/data/gm4_trees/function/sapling/advance_stage.mcfunction
```diff
@@ -10,8 +10,9 @@ execute if block ~ ~ ~ minecraft:spruce_sapling run setblock ~ ~ ~ minecraft:spr
 execute if block ~ ~ ~ minecraft:jungle_sapling run setblock ~ ~ ~ minecraft:jungle_sapling[stage=0]
 execute if block ~ ~ ~ minecraft:acacia_sapling run setblock ~ ~ ~ minecraft:acacia_sapling[stage=0]
 execute if block ~ ~ ~ minecraft:dark_oak_sapling run setblock ~ ~ ~ minecraft:dark_oak_sapling[stage=0]
-execute if block ~ ~ ~ minecraft:mangrove_propagule[waterlogged=false] run setblock ~ ~ ~ minecraft:mangrove_propagule[stage=0,waterlogged=false]
-execute if block ~ ~ ~ minecraft:mangrove_propagule[waterlogged=true] run setblock ~ ~ ~ minecraft:mangrove_propagule[stage=0,waterlogged=true]
+execute if block ~ ~ ~ minecraft:pale_oak_sapling run setblock ~ ~ ~ minecraft:pale_oak_sapling[stage=0]
+execute if block ~ ~ ~ minecraft:mangrove_propagule[waterlogged=false,hanging=false,age=4] run setblock ~ ~ ~ minecraft:mangrove_propagule[stage=0,age=4,waterlogged=false]
+execute if block ~ ~ ~ minecraft:mangrove_propagule[waterlogged=true,hanging=false,age=4] run setblock ~ ~ ~ minecraft:mangrove_propagule[stage=0,age=4,waterlogged=true]
```
misode, line 15: Why does this need to check the `age`? According to the wiki *"The age of a hanging propagule. This has no meaning for non-hanging propagules."*

TheThanathor: It doesn't technically need to _check_ the age, just needs to set it. Checking is just another way to make sure it's not replacing some custom sapling that messes with the age

misode: Do vanilla standing propagules always have `age=4`?

Bloo-dev: I just checked in game, 1.21.10. It seems standing propagules always have age=4.

TheThanathor: Vanilla should always be `age=4`, though setblock has `age=0` as default

PR #1255, opened 2026-01-23
Blasting Maces

drodo-mc:
Blasting Maces module submission.

Adds a new Blasting enchantment that allows the player to dig areas of 3x3 (with Blasting I) and 5x5 (with Blasting II). The breaking area is directional, depending on where the player is looking. Durability loss scales with the number of blocks broken.

The module also takes the grindstone into account, adding or removing the tool component when the item is enchanted or when the enchantment is removed via a grindstone.


gm4/skin_cache.json
TheThanathor, line 1: This doesn't seem to be intentional

gm4_blasting_maces/data/gm4_blasting_maces/enchantment/blasting.json
```diff
@@ -0,0 +1,23 @@
...
+        "base": 10,
+        "per_level_above_first": 20
+    },
+    "max_cost": {
+        "base": 25,
+        "per_level_above_first": 20
+    },
+    "anvil_cost": 4,
+    "slots": [
+        "any"
```
TheThanathor:
```suggestion
        "mainhand"
```

gm4_blasting_maces/data/gm4_blasting_maces/function/init.mcfunction
```diff
@@ -0,0 +1,16 @@
+execute unless score blasting_maces gm4_modules matches 1 run data modify storage gm4:log queue append value {type:"install",module:"Blasting Maces"}
+execute unless score blasting_maces gm4_earliest_version < blasting_maces gm4_modules run scoreboard players operation blasting_maces gm4_earliest_version = blasting_maces gm4_modules
+scoreboard players set blasting_maces gm4_modules 1
+
+# scoreboards for module data
+scoreboard objectives add gm4_blast_data dummy
+
+# scoreboard for tracking mace usage
+scoreboard objectives add gm4_use_mace used:mace
```
TheThanathor:
```suggestion
scoreboard objectives add gm4_blast_use_mace used:mace
```
It's better to use your prefix for everything, just in case

gm4_blasting_maces/data/gm4_blasting_maces/function/mace/add_tool_component.mcfunction
```diff
@@ -0,0 +1,8 @@
+# Adds the tool component to the blasting mace so that it can break blocks
+# @s = player holding a mace with blasting enchantment but no tool component
+# at world spawn
+# run from gm4_blasting_maces:tick
+
+# add tool component
+execute if items entity @s weapon.mainhand minecraft:mace run item modify entity @s weapon.mainhand gm4_blasting_maces:add_tool
+execute if items entity @s weapon.offhand minecraft:mace run item modify entity @s weapon.offhand gm4_blasting_maces:add_tool
```
TheThanathor: You should check to make sure the mace you're changing is actually enchanted. Right now it can change an unenchanted mace if you're holding an enchanted one in the other hand

gm4_blasting_maces/data/gm4_blasting_maces/function/mace/remove_tool_component.mcfunction
```diff
@@ -0,0 +1,8 @@
+# Removes the tool component from a mace that no longer has the blasting enchantment
+# @s = player holding a mace with tool component but no blasting enchantment
+# at world spawn
+# run from gm4_blasting_maces:tick
+
+# remove tool component
+execute if items entity @s weapon.mainhand minecraft:mace run item modify entity @s weapon.mainhand gm4_blasting_maces:remove_tool
+execute if items entity @s weapon.offhand minecraft:mace run item modify entity @s weapon.offhand gm4_blasting_maces:remove_tool
```
TheThanathor: Same as the one before, check to make sure the mace doesn't have the enchantment

gm4_blasting_maces/data/gm4_blasting_maces/function/mining/destroy_block.mcfunction
```diff
@@ -0,0 +1,8 @@
+# Destroys a single block at the current position if it is breakable and not unbreakable
+# @s = none
+# at block position
+# run from gm4_blasting_maces:mining/*
+
+# destroy block and increment counter (must be in the whitelist AND not in the blocklist)
+execute if block ~ ~ ~ #gm4_blasting_maces:breakable unless block ~ ~ ~ #gm4_blasting_maces:unbreakable run scoreboard players add $blast_count gm4_blast_data 1
+execute if block ~ ~ ~ #gm4_blasting_maces:breakable unless block ~ ~ ~ #gm4_blasting_maces:unbreakable run setblock ~ ~ ~ air destroy
```
TheThanathor: Technically having a single list is more performant, but this might be worth it for readability.
In any case you can reduce block checks with this:
```suggestion
scoreboard players set $blast_success gm4_blast_data 0
execute if block ~ ~ ~ #gm4_blasting_maces:breakable unless block ~ ~ ~ #gm4_blasting_maces:unbreakable store success score $blast_success gm4_blast_data run setblock ~ ~ ~ air destroy
scoreboard players operation $blast_count gm4_blast_data += $blast_success gm4_blast_data
```

gm4_blasting_maces/data/gm4_blasting_maces/function/mining/detonate.mcfunction
```diff
@@ -0,0 +1,56 @@
...
+scoreboard players set $blast_count gm4_blast_data 0
+
+# pitch determines up/down/horizontal:
+# < -45 = looking up (mine ceiling)
+# > 45 = looking down (mine floor)
+# -45 to 45 = horizontal (mine wall based on yaw)
+
+# level 1 (3x3x1)
+# looking up/down - mine 3x3 horizontal plane (X-Z)
+execute if predicate gm4_blasting_maces:blasting_level_1 unless predicate gm4_blasting_maces:blasting_level_2 if score @s gm4_blast_pitch matches ..-45 run function gm4_blasting_maces:mining/horizontal_plane_3x3
```
TheThanathor: You always use `if predicate gm4_blasting_maces:blasting_level_1 unless predicate gm4_blasting_maces:blasting_level_2`, it would be better to use a predicate that checks specifically for level 1 instead

gm4_blasting_maces/data/gm4_blasting_maces/function/player/analyze_mace.mcfunction
TheThanathor, line 1: Since this is only called from one location you might as well just call the single command directly

gm4_blasting_maces/data/gm4_blasting_maces/function/player/blast.mcfunction
```diff
@@ -0,0 +1,30 @@
...
+scoreboard players set $success gm4_blast_data 0
+execute as @e[type=item,limit=1,distance=..8,sort=nearest,nbt={Age:0s}] at @s store success score $success gm4_blast_data \
+  if items entity @s container.* #gm4_blasting_maces:breakable align xyz run summon marker ~0.5 ~0.5 ~0.5 {Tags:["gm4_blast_source","gm4_blast_marker","smithed.entity","smithed.strict"]}
+
+# if we found a valid block, analyze the mace and start mining
+execute if score $success gm4_blast_data matches 1 run function gm4_blasting_maces:player/analyze_mace
+
+# store player rotation for directional mining
+execute store result score @s gm4_blast_pitch run data get entity @s Rotation[1] 1
+execute store result score @s gm4_blast_yaw run data get entity @s Rotation[0] 1
```
TheThanathor: If you only need these if `$success gm4_blast_data` is 1 add a check here to save on possible player nbt reading. It's probably also better to copy the player Rotation into storage, then setting scores from that

gm4_blasting_maces/data/gm4_blasting_maces/function/player/modify_mace_durability.mcfunction
```diff
@@ -0,0 +1,23 @@
+# Reduces the durability of the blasting mace according to the number of blocks broken
+# @s = player with blasting mace who just broke blocks
+# at @s
+# run from gm4_blasting_maces:player/blast
+
+# get max damage - 1 for mace (durability 500)
+scoreboard players set $max_damage gm4_blast_data 499
```
TheThanathor: Check for existence of a damage component on the item first, in case it has custom durability, also check if it has the unbreakable tag!

gm4_blasting_maces/data/minecraft/tags/enchantment/non_treasure.json
```diff
@@ -0,0 +1,5 @@
+{
```
TheThanathor, line 1:
```suggestion
{
    "replace": false,
```
You need this or it'll override the vanilla tag

gm4_blasting_maces/beet.yaml
```diff
@@ -0,0 +1,24 @@
...
+  gm4:
+    versioning:
+      schedule_loops: [tick]
+    website:
+      description: Dig a hole like a mole!
+      recommended:
+        - gm4_resource_pack
+      notes: []
+    video: null
+    wiki: null
```
TheThanathor: You can add a wiki link here, since they're always named after the module
```suggestion
    wiki: "https://wiki.gm4.co/wiki/Blasting_Maces"
```

gm4_blasting_maces/beet.yaml
```diff
@@ -0,0 +1,24 @@
...
+    credits:
+      Creator:
+        - Drodo
```
TheThanathor, line 23: You're not in our `gm4/contributors.json` yet, you should add an entry for yourself there (can use an empty list if you don't have a link)

TheThanathor (changes requested):
I've left a few comments for small issues, adding some bigger points here:

1. I'm not really a fan of the level 2 of this enchantment, breaking a 5x5 area "feels" off to me, it's kind of breaking with the one-block-at-a-time philosophy (yes I know the 3x3 does as well, but it's kind of okay somehow?) Personally I'd like to more if it's just the single level for the enchantment, but good to get more views on this.
2. The biggest improvement that could be made is to make use of local coordinates (the `^` things) so you don't need different functions for each orientation.
3. In my opinion it would make more sense for the breaking orientation to depend on the face of the block the player targeted instead of their facing direction (this might be hard to figure out though)
4. Also up for debate, should the breaking count as if it was broken by an explosion? (so metallurgy drops), it would make sense lore wise. The Mace is late game enough that I feel it would probably not be too overpowered, but need more input on this.

I've tested the module in-game and everything worked as expected, though I didn't test the ~~dark magic~~ Guidebook

gm4_blasting_maces/data/minecraft/tags/enchantment/non_treasure.json
BPR02: it defaults to `"replace": false`, so it's not actually necessary

gm4/skin_cache.json
drodo-mc: This was changed automatically after running beet. Could be due to the following lines inside guidebook/blasting_maces.json

"icon": {
    "id": "minecraft:mace"
  },

runcows: Oh did you run the build for release code or dev? Every time i run the build for release it updates the skin cache and i just discard the change

drodo-mc: I did "uv run beet dev". Guess I should removed it then

gm4_blasting_maces/data/gm4_blasting_maces/guidebook/blasting_maces.json
```diff
@@ -0,0 +1,94 @@
...
+        ]
+      ],
+      "pages": [
+        [
+          {
+            "insert": "title"
+          },
+          {
+            "translate": "text.gm4.guidebook.blasting_maces.description",
+            "fallback": "Smash through stone and soil alike and blast your way forward in seconds. Clear massive areas with a single swing. Perfect for carving tunnels, flattening mountains or digging holes like a mole. Just keep an eye on that durability!"
```
runcows: I'm unsure of this guidebook entry. The default behaviour of a mace is unchanged with this module. This entry not mentioning the enchantment at all is therefore confusing. If I were doing this I'd probably just have the entry on the blasting enchantment and leave this one out

gm4/skin_cache.json
runcows: Oh hmm, yeah this definitely isn't your fault, its just a weird oddity that's been happening that nobody's gotten around to 😅

gm4_blasting_maces/data/gm4_blasting_maces/guidebook/blasting_maces.json
drodo-mc: I wasn't sure how to do it. 
1. Do we want the player to know the enchantment exists when they obtain a mace
2. Or directly give them the information on how it works after they discover the enchantment?

runcows: For Reeling Rods I did the second option. It felt like the most natural feeling option, like a player gets an enchant they don't know about and go "what's this?" and immediately get a guidebook page

gm4_blasting_maces/data/gm4_blasting_maces/function/player/modify_mace_durability.mcfunction
drodo-mc: Didn't know that was a thing. Arborenda should be updated to use it as well

runcows: I don't know if any of our modules account for custom durability values set by other datapacks or commands...

gm4_blasting_maces/data/gm4_blasting_maces/function/mining/detonate.mcfunction
drodo-mc: Got it. Wasn't using it because I was reusing the level 1 predicate with the min to check for mace being held.

gm4/skin_cache.json
```diff
@@ -1104,15 +1103,6 @@
...
       "gm4_custom_crafters:custom_crafter"
-    ],
-    "gm4_zauber_liquids": [
-      "gm4_potion_liquids:liquids/harming",
-      "gm4_potion_liquids:liquids/healing",
-      "gm4_potion_liquids:liquids/leaping",
-      "gm4_potion_liquids:liquids/poison",
-      "gm4_potion_liquids:liquids/regeneration",
-      "gm4_potion_liquids:liquids/strength",
-      "gm4_potion_liquids:liquids/swiftness"
```
drodo-mc: This was also automatically changed after running beet.

Bloo-dev:
I have re-targeted this at `update-26.1`. Changes for compatibility with 26.1 may still be required.

TheThanathor:
Uh ignore that one, not sure what happened there, just reviewing this now

gm4_blasting_maces/data/gm4_blasting_maces/function/mining/horizontal_plane_3x3.mcfunction
TheThanathor, line 1: This could be combined with the vertical wall by setting the rotation before calling the function, not a necessity though

TheThanathor (approved):
The code looks good, I do have a few points on the design:
- I still think a 5x5 area is too much and would prefer if the enchantment is just 1 level, maybe something to discuss in the discord forum channel
- the explosion sound is a bit loud in my opinion, and it can get a bit grating when it plays a lot of times in a row

drodo-mc:
- Switched the trigger from breaking a block to holding right click to charge and releasing to blast
- Removed the blasting level 2
- Breaking orientation now depends on the face of the block the player targeted (uses raycasting with fallback for diagonals that still uses player rotation)
- Now works in both offhand and mainhand (just like a bow/trident)
- Added a knockback effect using a wind charge
- Added a 16t main loop for optimizing how often the consumable component updates
- Updated the guidebook and advancement strings

SpecialBuilder32 (approved):

Bloo-dev:
Ready to merge once another code review is done.

TheThanathor (approved):
Code looks good! Maybe some small optimization possible but since it only runs when a player uses the Mace it's not a huge deal.
Module has been tested in game and is working with the exception of the wind charge at the floor

PR #1259, opened 2026-01-28
Lib Hooked Entity

runcows:
Separates the logic Reeling Rods uses to select the hooked entity into its own library targetting the `update-26.1` branch.
Updates Reeling Rods to use this library.

Not entirely sure on the naming of this library, since it's a very specific scenario. I considered `lib_hooked_entity_selection` but that is quite verbose.

Also I noticed that the readme files of all the other libraries were using a broken link to the repo, so this also just fixes that while I'm here.

Tested in game with Reeling Rods. Works as expected.


lib_hooked_entity/data/gm4_hooked_entity/function/player/cast_line.mcfunction
```diff
@@ -0,0 +1,7 @@
+# sets rods_cast to 0 and schedules id loop
```
Bloo-dev: Yeah, it does do that... I suggest something a bit less literal.
```suggestion
# Resets the player's rods cast score and initializes the bobber-id-assignment loop
```

PR #1263, opened 2026-02-22
fix guidebook infinite loop

BPR02:
This adds a flag to let modules know when it's okay to start granting guidebook pages (via advancements). In addition to updating guidebook, all modules need to built and updated to have the fix applied, but this technically does not break old modules (those modules just keep the bug until it is updated), and thus is not a breaking change.

### DIFF
--- a/gm4_guidebook/data/gm4_guidebook/function/init.mcfunction
+++ b/gm4_guidebook/data/gm4_guidebook/function/init.mcfunction
@@ -19,7 +19,9 @@ scoreboard players reset $module_count gm4_guide
 #declare storage gm4_guidebook:pages
 schedule function gm4_guidebook:update_storage/setup_markers 1t
 
+scoreboard players set $ready gm4_guide 0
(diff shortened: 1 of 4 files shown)


gm4_guidebook/data/gm4_guidebook/function/init.mcfunction
```diff
@@ -19,7 +19,9 @@ scoreboard players reset $module_count gm4_guide
 #declare storage gm4_guidebook:pages
 schedule function gm4_guidebook:update_storage/setup_markers 1t
 
+scoreboard players set $ready gm4_guide 0
 schedule function #gm4_guidebook:setup_storage 5t
+schedule function gm4_guidebook:update_storage/mark_ready 6t
```
misode: The `5t` and `6t` here is a bit hacky. I think it would be better to schedule a non-tag function in 5 ticks. In that function you call the setup_storage tag and then set the scoreboard.

PR #1264, opened 2026-02-22
26.1 Tag List Updates

runcows:
A bunch of tag list updates.
- Delete unused `#gm4:air` and `#gm4_teleportation_anchors:vehicles`
  - Mixed in here is a change to teleportation anchors `in_vehicle` predicate. This is used for ignoring shulkers or endermen in vehicles. Predicate does not need to check vehicle type, updated.
- Replaced custom tags with identical minecraft tags
  - Replaced `#gm4:undead` with `#minecraft:undead`
  - Replaced `#gm4:boats` with `#minecraft:boat`
  - `#gm4:replaceable` cannot be fully replaced with `#minecraft:replaceable` due to MC-261616 (`#minecraft:replaceable` is missing sculk veins)
- `#gm4_tunnel_bores:not_rail_solid` was not taking into account copper chains, it has been updated to use `#minecraft:chains`. I will make a separate PR to address this for the 1.21.5-1.21.11 branch
- `#gm4_tunnel_bores:rail_replaceable` updated to use `#gm4:replaceable` since it was already targeting the same blocks, save a couple that it should anyway.
- Squash `"required":false` entries used for cross version compatibility


Bloo-dev (approved):
I checked all the changed files; Seems good. What I did not check for is potentially missed tags.

PR #1268, opened 2026-02-22
Generate Soul Glass Predicate

BPR02:
This PR uses bolt to generate the soul glass beam check instead of having it pregenerated. It also makes it check the full height range of -63-319 for the 1.17+ overworld. 

Note this PR also hardcodes the height range check for each dimension, so custom dimensions aren't supported (but can have added support via a function tag) and changes to the vanilla dimension height ranges may cause soul beacons to function improperly.


gm4_soul_glass/data/gm4_soul_glass/predicate/has_beam_overworld.json
```diff
@@ -0,0 +1,71 @@
...
+      {
+        "condition": "minecraft:location_check",
+        "predicate": {
+          "position": {
+            "y": {
+              "min": y
+            }
+          }
+        }
+      },
```
runcows, line 19: What is this location check is doing in the predicate?
I don't fully understand its purpose here or in the current nightmare-fuel predicate, but in this current implementation, this check succeeding would be enough for it to not check the remaining conditions, right?

runcows (changes requested):
I don't like how this PR essentially drops support of custom dimensions without explicit support with a predicate which is cumbersome to create without bolt. I especially worry about the number of help requests we get as a result of this 😓 

My gut tells me there HAS to be a better way of checking if a beacon is active or not, other than checking every block above the beacon, but my mind is coming up short. Would it cause errors for the predicate to check block spaces outside of the build range?

gm4_soul_glass/data/gm4_soul_glass/predicate/has_beam_overworld.json
runcows: Wait ok, i understand this location check now, its to exclude blocks under the beacon. This should be true AND the following block checks. The current version is OR and will succeed regardless of the block checks.

BPR02: If you mean current version as in the one that's in `master` right now, that was made before `all_of` (AND) was added, and thus needed to use DeMorgan's to derive an `all_of` (AND) using `inverted` (OR) and `any_of` (OR). The current version does the same thing as this one.

BPR02:
> I don't like how this PR essentially drops support of custom dimensions without explicit support with a predicate which is cumbersome to create without bolt. I especially worry about the number of help requests we get as a result of this 😓
> 
> My gut tells me there HAS to be a better way of checking if a beacon is active or not, other than checking every block above the beacon, but my mind is coming up short. Would it cause errors for the predicate to check block spaces outside of the build range?

Yes, the predicate will fail if a block check occurs outside of the dimension bounds. We could, at runtime, determine the min and max altitudes of a dimension, then macro a predicate into existence if we want to support custom dimensions (or modified vanilla dimensions).

gm4_soul_glass/data/gm4_soul_glass/predicate/has_beam_overworld.json
runcows: No in that message, current was referring to this PR. For each block in the predicate in this PR, the condition is `any_of` and so the min block position check will cause each block position to succeed if above the execution position.

Edit: I did not understand the predicate fully, it is complicated

runcows:
Misode has informed me of wizardry. If you use a predicate to check the light level in a location check is 0-15, that predicate will only fail if the location is out of the dimension bounds.

In theory we could have just one predicate that covers the max world size possible, but we should benchmark that, since checking 4064 blocks might be a bit much.

BPR02:
> Misode has informed me of wizardry. If you use a predicate to check the light level in a location check is 0-15, that predicate will only fail if the location is out of the dimension bounds.

True, we could check "is this block in bounds, plus the other original checks."

BPR02:
remaking this PR with the light level check predicate for full dimension height range

PR #1269, opened 2026-02-23
Teleportation Anchors: replace display armor stand with item display

runcows:
Replaces display armor stand with item display for Teleportation Anchors and Jammers.
Item displays are located at the center of the blocks, same place as the markers.

Aside from modernizing, this also fixes an issue with the resource pack that cropped up last version, where the small armor stand heads started rendering bigger. This meant that the places we used this had their models inflated a bit and no longer fit the block. This issue affects other modules too, just addressing this per module.

~~Due to the needed upgrade path, the minor version has been bumped. This is of note for https://github.com/Gamemode4Dev/GM4_Datapacks/pull/1256 where we want to bump all minor versions anyway.~~

I have tested the item display in game and it works as expected. I have not yet tested the upgrade path, but after combing over it for a while, I see no reason it shouldn't work. Will test soon though.

### DIFF
--- a/gm4_teleportation_anchors/beet.yaml
+++ b/gm4_teleportation_anchors/beet.yaml
@@ -1,6 +1,6 @@
 id: gm4_teleportation_anchors
 name: Teleportation Anchors
-version: 1.7.X
+version: 1.6.X
 
 data_pack:
   load: .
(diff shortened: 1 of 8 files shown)


misode (approved):
The upgrade path didn't automatically run for me locally, but I think that's because the `earliest_version` scores don't really work in the dev versions. Works otherwise.

runcows:
Oh yep our dev command doesn't inject version scores at all. It's been a bit of a thorn in my side for testing upgrade paths. A few weeks ago I was looking into adding a flag for our dev command to enable it, but I haven't gotten back to it

PR #1270, opened 2026-02-23
Replace team hack for selecting mobs with `type=!#gm4:non_living`

runcows:
This tech no longer works, replacing it with something that does :P


misode (approved):

misode:
Do we want to backport this fix to older versions, and if so, how far back?

PR #1271, opened 2026-02-23
Use custom module icons for output packs where we have them

runcows:
A couple modules had no or incorrect `pack.png` files, causing them to display the default gm4 logo in the built datapack


misode (approved):
Seems fine.

PR #1273, opened 2026-02-23
Disassemblers: Migrate from armor stand to item display

runcows:
Replaces armor stands with item displays.
Resolves a visual bug with the resource pack, where the disassembler texture doesn't fit the block correctly.

Item display is located at the center of the disassembler. Entity version has been incremented.

The caller loot table now runs on the forceloaded armor stand.

Similar to https://github.com/Gamemode4Dev/GM4_Datapacks/pull/1272, this also included is a change to the placement function, moving from 6 functions, one for each cardinal direction, to one which uses macros to control the rotation changes.

There is an included upgrade path, I've tested it in game and it works as expected.

~~Due to the upgrade path, the minor version has been bumped. This is of note for https://github.com/Gamemode4Dev/GM4_Datapacks/pull/1256 where we want to bump all minor versions anyway.~~


gm4_disassemblers/data/gm4_disassemblers/function/machine/place_rotated.mcfunction
```diff
@@ -0,0 +1,21 @@
...
+data merge block ~ ~ ~ {CustomName:{"translate":"gm4.second","fallback":"%1$s","with":[{"translate":"container.gm4.disassembler","fallback":"Disassembler"},[{"text":" ","font":"gm4:half_invert"},{"translate":"container.gm4.disassembler","fallback":"Disassembler","font":"gm4:half_invert"},{"translate":"container.gm4.disassembler","fallback":"Disassembler","font":"gm4:offscreen"},{"translate":"gui.gm4.disassembler","fallback":"","font":"gm4:container_gui","color":"white"},{"text":" ","font":"gm4:half_invert"},{"translate":"container.gm4.disassembler","fallback":"Disassembler","font":"gm4:half_invert"},{"translate":"container.gm4.disassembler","fallback":"Disassembler","font":"gm4:default","color":"#404040"}]]}}
+
+# summon display and marker entity
+$summon item_display ~ ~ ~ {\
+  CustomName:"gm4_disassembler_display",\
+  Tags:["gm4_no_edit","gm4_disassembler_display","gm4_machine_display","smithed.entity","smithed.strict","gm4_new_machine"],\
+  item:{id:"minecraft:tnt",count:1,components:{"minecraft:custom_model_data":{"strings":["$(model)"]}}},\
+  item_display:head,\
+  brightness:{sky:15,block:15},\
+  Rotation:$(Rotation),\
```
misode: I think I would prefer all lowercase snake_case macro variables as a convention, at least when we're creating the keys ourself. This isn't yet in our naming guidelines but I think it should be.

gm4_disassemblers/data/gm4_disassemblers/function/upgrade_paths/2.6.mcfunction
```diff
@@ -0,0 +1,6 @@
+# @s = player
+# at @s
+# run via upgrade paths util
+
+execute as @e[type=armor_stand,tag=gm4_disassembler_stand] at @s positioned ~ ~0.4 ~ \
+  run function gm4_disassemblers:upgrade_paths/2.6/item_display
```
misode, line 6: I don't think we should start adding random line continuations like these. Mcfunctions always have had very long lines so your editor should already be wrapping lines imo.

gm4_disassemblers/data/gm4_disassemblers/function/process.mcfunction
```diff
@@ -4,4 +4,4 @@
 # run from main
 
 data modify storage gm4_disassemblers:temp Items set from block ~ ~ ~ Items
-execute if data storage gm4_disassemblers:temp Items[0] unless data storage gm4_disassemblers:temp Items[1] positioned ~ ~-0.4 ~ as @e[type=armor_stand,tag=gm4_disassembler_stand,limit=1,distance=..0.01] positioned ~ ~0.4 ~ run function gm4_disassemblers:check_item
+execute if data storage gm4_disassemblers:temp Items[0] unless data storage gm4_disassemblers:temp Items[1] as 344d47-4-4-4-f04ce104d run function gm4_disassemblers:check_item
```
misode, line 7: I think it's unfortunate that we need to rely on the forceloaded chunk here (which I generally like to avoid, mostly because I hope in the future we'll be able to eliminate it) - but for this module it's not too bad because lib_machines already needs it.

Because we have the massive loot table to compute the disassebly items I'm not sure what alternatives we have though.

gm4_disassemblers/data/gm4_disassemblers/function/machine/destroy.mcfunction
```diff
@@ -3,8 +3,9 @@
 # located at @s
 # run from gm4_disassemblers:machine/verify_destroy
 
-# kill entities related to machine block
+# kill entities related to machine block (armor stand needs to be killed for legacy reasons)
 execute positioned ~ ~-0.4 ~ run kill @e[type=armor_stand,tag=gm4_disassembler_stand,limit=1,distance=..0.01]
+kill @e[type=item_display,tag=gm4_disassembler_display,limit=1,distance=..0.01]
```
misode, line 9: Add a comment here to remind us that we can remove this in a couple versions.

misode (commented):
Mostly some stylistic comments.

I actually quite like the macro approach here. The macro will always be cached and it's not super complicated with bolt or jinja.

gm4_disassemblers/data/gm4_disassemblers/function/upgrade_paths/2.6.mcfunction
runcows: I have bad news for you. I've done this in many places that are already merged 😅 

It's also not entirely random, I mainly use it to separate out the execute sub commands from the command that gets run at the end

misode (approved):

PR #1274, opened 2026-02-25
Calling Bell: refactor to use advancements instead of scoreboard checks

runcows:
Large refactor of Calling Bell.

No longer has any ticking functions, all logic is now only run on the event of a player ringing a bell with an emerald, via an advancement.
Raycast now also uses block interaction range.

Build will fail due to mecha throwing an error due to `time query day repetition` which is the new way of getting day number in the 26.1 snapshots. We can force the build by prefixing the line with `raw ` and requiring bolt


runcows:
This PR has had bolt added with `raw ` to make the build run.
I believe we decided this was best for now, and we can just remove it later once Mecha updates with a 26.1 command tree

misode (approved):
Tested. Code looks good.

PR #1138, opened 2025-06-08
Horsemanship

TheThanathor:
Level up your horses by using them to improve their health, speed and jump strength

Full module description will be added soon once the design is more finalized

PR #1276, opened 2026-02-28
Improve Guidebook Item Resolver

BPR02:
This PR adds the following changes to the guidebook item tag resolution (getting a single item from an item tag)
- supports non-vanilla tags
- supports {"id": "...", "required": false} entries in item tags (previously assumed strings)

Required for #1138 to build properly.

### DIFF
--- a/gm4_guidebook/generate_guidebooks.py
+++ b/gm4_guidebook/generate_guidebooks.py
@@ -888,10 +888,10 @@ def get_item_from_tag(ctx: Context, item_tag: str, vanilla: Vanilla) -> str:
   # prepare item tag for searching
   if ":" in item_tag:
     prefix, tag_target = item_tag.split(":")
+    prefix = prefix.removeprefix("#")
   else:
     prefix = ""
     tag_target = item_tag.removeprefix("#")
-  prefix = prefix.removeprefix("#")
 
   # open item tag
   if prefix == "minecraft" or prefix == "":
--- a/gm4_guidebook/generate_guidebooks.py
+++ b/gm4_guidebook/generate_guidebooks.py
@@ -882,26 +882,34 @@ def item_to_display(item: str, components: dict[str, Any] | None, ctx: Context) 
 
 
 """
-Recursively reads vanilla item tags to find a single item to use
+Recursively reads item tags to find a single item to use
 """
-def get_item_from_tag(item_tag: str, vanilla: Vanilla) -> str:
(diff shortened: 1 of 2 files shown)


gm4_guidebook/generate_guidebooks.py
```diff
@@ -882,26 +882,37 @@ def item_to_display(item: str, components: dict[str, Any] | None, ctx: Context)
...
   # prepare item tag for searching
-  if "minecraft" in item_tag:
-    if "#" in item_tag:
-      item_tag = item_tag[11:]
-    else:
-      item_tag = item_tag[10:]
-  elif item_tag.split(":")[0] != "minecraft":
-    raise ValueError("Only vanilla item tags are supported")
+  if ":" in item_tag:
+    prefix, tag_target = item_tag.split(":")
```
Bloo-dev: This could throw a ValueError in case the string `"item.tag"` contains multiple `":"`.
```suggestion
    prefix, tag_target = item_tag.split(":", maxsplit=1)
```

gm4_guidebook/generate_guidebooks.py
```diff
@@ -882,26 +882,37 @@ def item_to_display(item: str, components: dict[str, Any] | None, ctx: Context)
...
+    item_tags = vanilla.mount("data/minecraft/tags").data["minecraft"].item_tags
+  else:
+    item_tags = ctx.data[prefix].item_tags
+  items: list[str|dict[str, Any]] = item_tags[tag_target].data["values"]
+
+  # get first item
+  if isinstance(items[0], str):
+    res: str = items[0]
+  else:
+    res: str = items[0]["id"]
```
Bloo-dev, line 907: Technically you don't know whether the `"id"` key exists.

Bloo-dev (changes requested):
I found some potential issues. These _should_ never occur, but it would be nice to catch them and raise custom ValueErrors instead.

gm4_guidebook/generate_guidebooks.py
BPR02: If the id doesn't exist it is not a valid entry in a function tag and thus having a KeyError be raised is a valid result.

BPR02:
Yeah, I agree the Guidebook script could use better error messages, but it might be better to do that in a separate PR that doesn't modify the actual result of a successful run

Bloo-dev (approved):

PR #1279, opened 2026-03-15
26.1 time query day repetition

runcows:
With 26.1, `time query day` no longer gets the day count, but the ticks elapsed each day. The new format is `time query day repetition`, which this PR updates the modules untouched by #1274 to use. (This PR does NOT touch Calling Bell)

Currently, Mecha doesn't support this command structure, so this PR also prefixes the relevant commands with `raw ` and requires bolt. We can remove this later, once Mecha is updated; I've added a TODO comment to remind us.

### DIFF
--- a/gm4_mysterious_midnights/beet.yaml
+++ b/gm4_mysterious_midnights/beet.yaml
@@ -9,6 +9,9 @@ pipeline:
   - gm4.plugins.extend.module
   - gm4.plugins.include.lib_forceload
 
+require:
+  - bolt
+
 meta:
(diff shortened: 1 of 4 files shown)


runcows:
Found another instance of `time query day` was in Zauber, but as that's being addressed #1262, that command has been left out of this PR.

Relevant file on the `update-26.1` branch:
`gm4_zauber_cauldrons\data\gm4_zauber_cauldrons\function\cauldron\extra_items\process_bottled_vex_items.mcfunction`

PR #1283, opened 2026-03-22
26.1 Bump version numbers

runcows:
Bump all minor version numbers, excluding:
- Zauber Liquids, as even with #1262, it will remain in limbo, unless misode wants to update that too
- Lib Hooked Entity, since it's new

Bump specific major version numbers
- Zauber Cauldrons, since #1262 is significant
- ~~Calling Bell, since its basically a full rewrite, even if the module is quite small~~
- Standard Crafting, we're in a post-custom-crafters world 🎉

Also went through all beet.yaml `required:` fields and updated those to match current version numbers

### DIFF
--- a/base/beet.yaml
+++ b/base/beet.yaml
@@ -1,4 +1,4 @@
-version: 1.8.0
+version: 1.9.0
 id: gm4
 
 data_pack:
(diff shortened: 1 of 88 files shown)


misode (commented):
I would revert the calling bell major bump, since it's only an internal refactor, not a feature update.

PR #1285, opened 2026-03-25
Interaction `response:1b` -> `/swing`

runcows:
For the interaction entities in Rope Ladders and Lively Lily Pads (`response` isn't used anywhere else), switches them to use the new `/swing` command.
This fixes a visual bug where using the offhand would still play a mainhand arm swing.

For Lively Lily Pads, this has meant upping the minor version (again💀) for an upgrade path to remove `response:1b` from existing interactions.

##### I'm not super happy with the number of repeated score checks here, but i was feeling lazy and didnt want to make ~10 new functions, each with just 2 commands....

### DIFF
--- a/gm4_lively_lily_pads/beet.yaml
+++ b/gm4_lively_lily_pads/beet.yaml
@@ -8,9 +8,6 @@ data_pack:
 pipeline:
   - gm4.plugins.extend.module
 
-require:
-  - bolt
-
 meta:
   gm4:
     versioning:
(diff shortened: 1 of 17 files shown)


runcows:
This PR has been updated to remove `raw ` and bolt since mecha has been updated

misode (approved):
I can confirm the issue existed and is fixed by this PR.

One minor detail, I would not make a minor version bump for lively lily pads, this is only a patch fix. Or was this only done so you could add an upgrade path?

runcows:
> was this only done so you could add an upgrade path?

Yep

PR #1288, opened 2026-04-03
Generate Soul Glass Predicate

BPR02:
This PR updates soul glass to properly support 1.17+ the overworld dimension height range. It also adds support for any custom dimension height range. This is done by generating the predicate with bolt to search each block (instead of having it pregenerated).


misode:
@BPR02 I've changed the predicate a bit to make it more understandable. I tested this in game, but please verify and confirm the logic makes sense.

misode (approved):

BPR02:
Yes, can confirm the logic makes sense. Since we check if the block is within bounds using the light level check, we no longer need to verify that it is within bounds of the world via hardcoded Y-level checks. So it makes sense to just check that the 4064 blocks above the beacon is both within bounds and a beacon passable block.

PR #1290, opened 2026-05-03
Add plain:true to some text components

misode:


misode:
I'm just going to merge this I guess

Issue #1013, opened 2024-07-25
Implementation Upgrades


SpecialBuilder32:
This issue (mirroring #1010) exists to track needed *implementation updates* to modules from new (or not so new now) datapack features, such as display entities. Please reply to this issue with any missing required updates and this header comment will be updated appropriately. 

* Item Display Migration
  - [ ] Apple Trees
  - [ ] Book Binders
  - [ ] Forming Press
  - [ ] Auto Crafting
  - [ ] Dissassemblers
  - [ ] Enchantment Extractors
  - [ ] Liquid Tanks
  - [ ] Metallurgy*
  - [ ] Orb of Ankou (Soul Forge)
  - [ ] Relocators
  - [ ] Smelteries
  - [ ] Teleportation Anchors

PR #1306, opened 2026-09-02
Update Apple Trees armor stands into display entities

### LINKED ISSUE #1289: Apples on Apple Trees Use Wrong Texture When Mature
**Describe the bug**
This does not directly affect gameplay, it just grants the player access to apples with the "unripe" texture. The Gamemode 4 RP is required.

**To Reproduce**
1. Install Apple Trees on a 26.1 world and get the Gamemode 4 RP
2. Watch an apple grow
3. Observe that the grown apple uses the "unripe" texture (see screenshots below)

**Expected behavior**
Grown apples should use the default apple texture.

**Screenshots**
Image
Image

**Environment**
 - Minecraft version: 26.1
 - Server version: 26.1
 - Installed Datapacks: Apple Trees
 - Installed Mods: None

SpecialBuilder32:
Upgrades apple trees apples from armor stands to item displays and interaction entities (finally). 

Other upgrades:
* Randomly adjust apple ages when spawned so whole tree does not drop at once
* Golden apples and golden roots can progress age even when unloaded
  - Capped by one-growth stage

Checks one more off #1013
Closes #1289 (but not for 26.1)

We may wish to delay this merge until 26.3, but I'll get yalls thought on that.


Bloo-dev:
We did a quick test of this in multiplayer and it seems to work just fine. It is to be noted that we did not wait for apples to grow much.

Bloo-dev:
The pixel scales are not yet set correctly and we are looking for new textures.

PR #1313, opened 2026-09-13
Split publish action into separate github job for better reporting

SpecialBuilder32:
Splits the release build pipeline into separate jobs so we get better logging when one fails (i.e., expired tokens). Each distribution site has it's own job and can fail or be re-run independently of the others. 

Publishing to gm4.co remains part of the `build` job, since it just pushes a commit to the release branch.

image

@misode please let me know if this looks correctly idempotent to you? I think I've implemented it that way, so long as we're requesting a re-run within the 2 day artifact expiry window.

Also also, I don't know if this is best merged before of after 26.3. Maybe at the same time? Let me know if I should redirect this from `master` to `update-26.3`


misode:
I think it would be better if the `build` step uploaded the datapack zips to the github artifact together with the beet cache manifest. Instead of the publish steps pulling the zips from the release branch.

